PR #4, opened 2024-08-07
CLI scaffolding

drewoldag:
This is meant to be a starting point for the CLI work, and is meant to give us a place to start hanging more code. The next step after this is integrated in my mind is to implement the user configuration file support. 

The current CLI will look something like this: 
```
>>> fibad
usage: fibad [-h] [--version] [{train,predict}] ...

Fibad CLI

positional arguments:
  {train,predict}  Verb to execute
  args             Arguments for the verb

options:
  -h, --help       show this help message and exit
  --version        Show version

FIBAD is the Framework for Image-Based Anomaly Detection
```

Getting top level help
```
>>> fibad --help
fibad --help
usage: fibad [-h] [--version] [{train,predict}] ...

Fibad CLI

positional arguments:
  {train,predict}  Verb to execute
  args             Arguments for the verb

options:
  -h, --help       show this help message and exit
  --version        Show version

FIBAD is the Framework for Image-Based Anomaly Detection
```

Running a fibad action:
```
>>> fibad train
Prending to run training...
Runtime config: None
```

Action-specific help
```
>>> fibad train --help
usage: fibad-train [-h] [-c RUNTIME_CONFIG]

Training with Fibad.

options:
  -h, --help            show this help message and exit
  -c RUNTIME_CONFIG, --runtime-config RUNTIME_CONFIG
                        Full path to runtime config file
```

Passing in a config argument
```
>>> fibad predict --runtime-config='/some/dir/over/here.toml'
Prending to run prediction...
Runtime config: /some/dir/over/here.toml
```

This work is heavily inspired by the CLI work done for Sorcha here: https://github.com/dirac-institute/sorcha/blob/main/src/sorcha_cmdline/main.py

### DIFF
--- a/src/fibad_cli/main.py
+++ b/src/fibad_cli/main.py
@@ -1,0 +1,53 @@
+import argparse
+import shutil
+import subprocess
+import sys
+from importlib.metadata import version
+
+
+def main():
+    """Primary entry point for the Fibad CLI. This handles dispatching to the various
+    Fibad actions. The actions are defined in the pyproject.toml project.scripts
+    section.
+    """
+
+    description = "Fibad CLI"
+    epilog = "FIBAD is the Framework for Image-Based Anomaly Detection"
+
+    #! We could potentially make this dynamic
+    fibad_verbs = ["train", "predict"]
+
+    parser = argparse.ArgumentParser(description=description, epilog=epilog)
+    parser.add_argument("--version", dest="version", action="store_true", help="Show version")
+    parser.add_argument("verb", nargs="?", choices=fibad_verbs, help="Verb to execute")
+    parser.add_argument("args", nargs=argparse.REMAINDER, help="Arguments for the verb")
+
+    args = parser.parse_args()
+
+    if args.version:
+        print(version("fibad"))
+        return
+
+    if not args.verb:
+        parser.print_help()
+        sys.exit(1)
+
+    fibad_action = f"fibad-{args.verb}"
+
+    # Ensure the action is available
+    if not shutil.which(fibad_action):
+        print(f"Error: '{fibad_action}' is not available.")
+        print("Is the action defined in the pyproject.toml project.scripts section?")
+        sys.exit(1)
+
+    # Execute the action with the remaining arguments
+    try:
+        result = subprocess.run([fibad_action] + args.args, check=True)
+        sys.exit(result.returncode)
+    except subprocess.CalledProcessError as e:
+        print(f"Error: Command '{fibad_action}' failed with exit code {e.returncode}.")
+        sys.exit(e.returncode)
+
+
+if __name__ == "__main__":
+    main()
--- a/src/fibad_cli/train.py
+++ b/src/fibad_cli/train.py
@@ -1,0 +1,27 @@
+import argparse
+
+
+def main():
+    """Argument parser to process command line arguments when training with Fibad."""
+
+    parser = argparse.ArgumentParser(description="Training with Fibad.")
+
+    parser.add_argument("-c", "--runtime-config", type=str, help="Full path to runtime config file")
+
+    args = parser.parse_args()
+
+    run(args)
+
+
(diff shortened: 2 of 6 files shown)


src/fibad_cli/main.py
```diff
@@ -0,0 +1,53 @@
...
+
+    description = "Fibad CLI"
+    epilog = "FIBAD is the Framework for Image-Based Anomaly Detection"
+
+    #! We could potentially make this dynamic
+    fibad_verbs = ["train", "predict"]
+
+    parser = argparse.ArgumentParser(description=description, epilog=epilog)
+    parser.add_argument("--version", dest="version", action="store_true", help="Show version")
+    parser.add_argument("verb", nargs="?", choices=fibad_verbs, help="Verb to execute")
```
mtauraso, line 22: Do we want to use subparsers here and let the individual verbs define their own flags?

src/fibad_cli/train.py
```diff
@@ -0,0 +1,27 @@
+import argparse
+
+
+def main():
+    """Argument parser to process command line arguments when training with Fibad."""
+
+    parser = argparse.ArgumentParser(description="Training with Fibad.")
+
+    parser.add_argument("-c", "--runtime-config", type=str, help="Full path to runtime config file")
```
mtauraso, line 9: Do we intend for these config files to be substantially different from one another?

mtauraso (approved):
A few questions, but nothing blocking IMO. This should get merged.

src/fibad_cli/train.py
aritraghsh09: The major difference that comes to mind would probably be sub-options for the different kinds of models that are implemented within fibad. But those can still probably be handled from within one "generic type" of config file.

aritraghsh09 (approved):
No major science-facing comments from my end. This looks good to be merged!

src/fibad_cli/main.py
drewoldag: That was one option, but I was leaning toward a simpler approach for now. It may make sense to go down that route. I still want to think a little more about what we were talking about on Tuesday - namely unlocking many or all of the keys defined in the config file as arguments. 

If we go down that route, subparsers may be the right approach. But I'm a little unsure/apprehensive about the order of events needed to load the config file, then use the contents to dynamically generate the parser arguments. In the base case of just looking at `fibad`'s default config file, this isn't too challenging - but as we open this up to plug-in style DataLoaders and Models, each with their own bespoke config parameters - then we have to do some code introspection to identify available classes, grab their default configs, merge that with the base default configs...

Perhaps I'm overcomplicating it 🤷  that's why it would be nice to chat it out 😄

src/fibad_cli/train.py
drewoldag: My initial thought here is that the configs should _not_ be substantially different. At this point I would advocate for all the user defined config values to be in the same file, even if they aren't used for a particular action. One obvious follow up would be, why not move the `--runtime-config` argument up to the `main.py`. Perhaps we could, but it's kind of nice to see `fibad train --help` produce output that shows that it expects a config file. 

I'm happy to hear counter points to that though. My opinion here isn't a super strong one 😃 

Very hand-wavy a config file might look something like the following. Note that it could include the configs for everything, even if it's just being used for training now, and just prediction later.

```
[fibad]  # general configs for running fibad???
max_gpus = 4  # just an example

[train]  # general configs for training
data_loader = base_loader
model_name = drewtonian  # this is type of model, right?
num_epochs = 100  # again, just an example

[train.base_loader]
data_directory = /foo/bar/data
...

[train.drewtonian]  # parameters for this model
num_layers = 10
...

[predict]  # general configs for prediction
data_loader = base_loader

[predict.base_loader]
data_directory = /foo/baz/new_data
```

I would love to talk to you both more about the structure of the config file - what is going to be most intuitive and easiest to use... etc.

PR #6, opened 2024-08-08
Scaffolding for config file handling and model selection

drewoldag:
This PR represents the scaffolding work for two aspects of fibad:

1. Basic support for reading and loading the default and the user's defined runtime configuration file for use within fibad. 
2. Basic support for dynamically loading models, either from a registry of built in models or via a user defined module spec, specified in the user's runtime configuration file. 

To use a built in model from the registry a user specifies in the .toml config file the following:
```
[train]
model_name = "name_of_built_in_model_class"
```

To specify a model outside of fibad, the users .toml config would include something like the following example. Basically what the user would write in a `from ... import ...` statement. 
```
[train]
model_class = "package.submodule.subsubmodule.ModelClass" 
```

Included here is an implementation of PyTorch's "Getting Started" CNN for classifying images in the CIFAR10 dataset - note that this is incomplete, and meant to be a way to test the model registration process, and will be a testbed for future work.


src/fibad/config_utils.py
```diff
@@ -0,0 +1,90 @@
+import os
+
+import toml
+
+DEFAULT_CONFIG_FILEPATH = "default_runtime_config.toml"
```
mtauraso, line 5: Perhaps `fibad_default_config.toml` ?

Also for later we may consider packaging the default config.

mtauraso (approved):
LGTM. Good to add tests, but I'm in favor of this getting merged even as-is.

src/fibad/config_utils.py
drewoldag: yep, that name sounds fine. I agree that we should package a default config. I would like to chat about how to make it consistent, specifically for test purposes. I've seen default config-like files get really out of hand with several version floating around in a package (i.e. one example in docs, another in tests, another here or there). I don't want to accidentally devolve in to that :)

drewoldag: Updated.

PR #7, opened 2024-08-09
Downloader prototype and CLI rework

mtauraso:
This is (like a lot of this project now) super preliminary. Comments encouraged, but may be addressed on followup for the sake of building out the scaffolding of this project quickly 🙂 

Reworking the CLI to have a single entrypoint. The intent here is to centralize config parsing in the hopes that we can keep discipline about reproducibility via the config file mechanism, and keep the CLI small. There is a potential trade-off in CLI ergonomics which we're going to keep an eye on as development continues.

Downloader: This is a basic downloader that can process the `z_0_025.fits` catalog and download snapshots. It is currently hard-coded to only attempt the first 10 sources. It uses the downloadCutouts utility which has been imported from HSC's repository.


src/fibad/__init__.py
```diff
@@ -1,3 +1,3 @@
 from .example_module import greetings, meaning
 
-__all__ = ["greetings", "meaning"]
+__all__ = ["greetings", "meaning", "download"]
```
drewoldag, line 3: Interesting, does this just grab the module named download? I don't think I've seen this syntax before.

src/fibad/downloadCutout/downloadCutout.py
drewoldag, line 1: Am I correct in assuming that this is the file that was pulled over from HSC-world?

src/fibad/download.py
```diff
@@ -0,0 +1,128 @@
...
+    finally:
+        os.chdir(old_cwd)
+
+
+def run(args, config):
+    """
+    Main entrypoint for downloading cutouts from HSC for use with fibad
+    """
+
+    config = config["download"]
```
drewoldag, line 39: We can be pretty confident that "download" will be a key in the config at this point, but I think it would be good practice to try to be as defensive as possible with something like `config.get('download', {})`. 

If we adopt something that enforces more of the config structure like Pydantic or just using dataclasses that might let us be a bit less explicit.

src/fibad/download.py
```diff
@@ -0,0 +1,128 @@
...
+    locations = filterfits(config.get("fits_file"), column_names)
+
+    # Sort by tract, ra, dec to optimize speed that the cutout server can serve us
+    #
+    # TODO: See if this sort is performed by downloadCutouts
+    # It appears downloadCutouts is doing some sorting prior to download, but
+    # unclear if it is the same sort
+    locations.sort(variable_fields)
+
+    # TODOxcxc slice up the locations
```
drewoldag, line 54: Ooo, what does `TODOxcxc` do? Is there some fancy IDE magic here? 😄

src/fibad/download.py
```diff
@@ -0,0 +1,128 @@
...
+    Takes our Download config and loads cutout config
+    common to all cutouts into a prototypical Rect for downloading
+    """
+    return dC.Rect.create(
+        sw=config["sw"],
+        sh=config["sh"],
+        filter=config["filter"],
+        rerun=config["rerun"],
+        type=config["type"],
+    )
```
drewoldag, line 96: I like how explicit this is, **kwargs always feels like it's hiding too much.

src/fibad/download.py
```diff
@@ -0,0 +1,128 @@
...
+    Rects are created with all fields in the default rect pre-filled
+
+    Offset here is to allow multiple downloads on different sections of the source list
+    without file name clobbering during the download phase. The offset is intended to be
+    the index of the start of the locations table within some larger fits file.
+    """
+    rects = []
+    for index, location in enumerate(locations):
+        args = {"lineno": index + offset}
+        args.update((field, location[field]) for field in variable_fields)
```
drewoldag, line 113: Could we use dictionary comprehension here? It's possible that under the hood Python is doing that anyway. `variable_fields` isn't huge, but inside the `locations` for loop it might add up? 

Something like:
```
for index, location in enumerate(locations):
    args = {}
    args = {(field, location[field]) for field in variable_fields}
    args['lineno'] = index + offset
    ....
```

src/fibad/download.py
drewoldag, line 1: While it is still preliminary code, don't forget to add the input params to the docstrings. 

I use "autoDocstring" as a plugin for VSCode that will automatically populate the docstring using the type hints. With the style set to `numpy`, it interfaces really smoothly with all the sphinx automation to generate the API docs. 

Name: autoDocstring - Python Docstring Generator
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring

src/fibad_cli/main.py
```diff
@@ -15,12 +31,16 @@ def main():
...
-    fibad_verbs = ["train", "predict"]
+    #! Somewhat difficult (perhaps impossible) to get this list from importlib/pkglib given
+    #! That a fibad verb is simply an object in a file.
+    fibad_verbs = ["train", "predict", "download"]
 
     parser = argparse.ArgumentParser(description=description, epilog=epilog)
+
     parser.add_argument("--version", dest="version", action="store_true", help="Show version")
+    parser.add_argument("-c", "--runtime-config", type=str, help="Full path to runtime config file")
+
```
drewoldag, line 42: As discussed, just want to confirm that both of the following will work on the command line still:

`fibad download --runtime-config=<foo>`
and
`fibad --runtime-config=<foo> download`

drewoldag (approved):
Looks pretty good to me. I left a few comments on here, but no blockers for sure. After the comments are addressed, feel free to merge.

src/fibad/downloadCutout/downloadCutout.py
mtauraso: Yes, LINCC_README.md documents its existance

src/fibad/download.py
mtauraso: pre-commit hook at job-2 used to catch the existance of "xcxc" to mean don't check in this comment.

Just an oversight.

src/fibad/download.py
mtauraso: oohh, that's cool. I'm going to try it out.

src/fibad/download.py
mtauraso: Given how sloppy the code I'm calling is with rect objects, this seemed necessary.

Below the call into the downloadCutout there is a lot of code that is happy to allow malformed rects to exist.

src/fibad/download.py
mtauraso: That's cleaner than what I wrote, and might be faster. I'll take it.

PR #9, opened 2024-08-12
Adding bandwith measurement to downloadCutout

mtauraso:
- Stats tracking of timing, data size, and bandwidth for cutout transfer
- Ability to specify an arbitrary list of filters to the downloader
- Configurable offset and extent for downloading from a large .fits catalog
- download() helper function codepath now supports kwargs passed from download() -> _download() -> _download_chunk() -> urllib.request.urlopen()
- Timout and chunk size for downloadCutout.py are configurable via this kwargs system.

### DIFF
--- a/src/fibad/download.py
+++ b/src/fibad/download.py
@@ -1,5 +1,7 @@
 import contextlib
+import datetime
 import os
+import urllib.request
 from pathlib import Path
 from typing import Union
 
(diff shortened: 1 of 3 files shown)


src/fibad/download.py
```diff
@@ -38,34 +40,34 @@ def run(args, config):
...
     column_names = ["object_id"] + variable_fields
-    locations = filterfits(config.get("fits_file"), column_names)
+    locations = filterfits(fits_file, column_names)
 
-    # Sort by tract, ra, dec to optimize speed that the cutout server can serve us
-    #
-    # TODO: See if this sort is performed by downloadCutouts
-    # It appears downloadCutouts is doing some sorting prior to download, but
-    # unclear if it is the same sort
-    locations.sort(variable_fields)
```
mtauraso, line 52: This sorting turned out to be unnecessary.

The `Rect` class in `downloadCutout.py` has a very particular sort order imposed on it, and the download codepath calls `Rect.explode()` under the hood, which calls `.sort()` on a list of `Rect`s particular to each request. This results in the server receiving lists sorted by Tract, RA, DEC and some other fields.

src/fibad/download.py
```diff
@@ -118,11 +120,77 @@ def create_rects(locations: Table, offset: int = 0, default: dC.Rect = None) ->
     return rects
 
 
+stats = {
```
mtauraso, line 123: This whole stats subsystem is pretty hacky. I've tried to make it unobtrusive to the CLI, but if we end up using it for anything beyond spot checks, we likely want to go find ourselves a library to do this.

drewoldag (approved):
This seems plenty fine to me!

PR #17, opened 2024-08-13
Retry and resume functionality for downloader

mtauraso:
Implementation of retry and resume within downloadCutout.py.

Retry means that when a request fails we will try again (defaults to 3 attempts). This is intended to address connection drops and the like. We use configurable exponential backoff to avoid a thundering herd if load from our client is causing some backend failure.

Resume describes the situation where the download fails for unrecoverable reasons (HSC infra goes down) or is terminated (e.g. downloads only occur at night). This generates a resume_download.toml file in the download directory, which allows the exact same download to resume from the chunk that was in progress when the interruption occurred. This resume functionality is off by default to preserve the `download()` interface used by `downloadCutout.py`'s CLI which does not support resume.

### DIFF
--- a/src/fibad/downloadCutout/downloadCutout.py
+++ b/src/fibad/downloadCutout/downloadCutout.py
@@ -1114,7 +1114,7 @@ def _download(
 
     datalist: list[tuple[int, dict, bytes]] = []
 
-    failed_rect_index = None
+    failed_rect_index = 0
 
     start_rect_index = 0
     if not onmemory and resume:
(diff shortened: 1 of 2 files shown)


aritraghsh09 (approved):
The changes look good to me, and I have no objections. But I have two comments:- 

1. The files that fail even after the specified number of attempts -- can we keep a running list of those object ids somewhere and then dump them as a .npy array or something similar?

2. Alternatively, I guess a separate check-script can be written that verifies whether an `object_id.fits` exists for each `object_id` in the download table; and then makes an array of all the `object_id`s for which image files were not found.

src/fibad/downloadCutout/downloadCutout.py
```diff
@@ -1078,21 +1114,160 @@ def _download(
...
+    RuntimeError
+        "Resume data in <path> corrupt" when the file is not a toml file containing keys
+        'checksum' and 'start_rect_index'
+    RuntimeError
+        "Resume data failed checksum ..." when the rect list has changed from when the resume data file was
+        written
+    """
+    # Load resume data so we start at the appropriate chunk.
+    if not os.path.exists(resume_data_filename):
+        return 0
```
drewoldag, line 1209: Was this a `RuntimeError` before? Seems like it makes sense for it to be a raised exception, but curious if there's a good reason for it to be `return 0`.

drewoldag (approved):
This looks pretty good. Only one little comment about return 0 vs. raise Exception.

src/fibad/downloadCutout/downloadCutout.py
mtauraso: Yeah, the reason for this is so if we're called with `resume=True` but there is no resume data, we will just download from the beginning.

This avoids a CLI flag (or other mechanism) in `download.py` or higher which has to know whether the user intends to resume or not. 

Maybe `resume` should really be `resume_if_possible` since that is what it really means.

drewoldag: oh, ok, I see what you mean here. I don't think I have a strong opinion here other than perhaps cleaning up the docstring and perhaps leaving a comment like "can't find the file, so starting from index 0" or something.

PR #18, opened 2024-08-13
Scaffolding for first data loader

drewoldag:


mtauraso (approved):
I think this is sufficiently done that its useful and ought to be merged.

PR #21, opened 2024-08-14
Organize pyproject.toml, don't accidentally include `data/` in commit

drewoldag:
Moving torchvision out of core dependencies in pyproject.toml. Now in an optional dependency section, `examples`. 
Users can install from source with `pip install -e .[dev,examples]`.

Added `data/` to .gitignore so that we don't accidentally commit something in there to the repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -148,3 +148,6 @@ _html/
 
 # Project initialization script
 .initialize_new_project.sh
+
+# General purposes data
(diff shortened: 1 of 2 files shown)


aritraghsh09 (approved):
Looks good!

mtauraso (approved):
LGTM 🚀

PR #27, opened 2024-08-15
Skip failed chunks

mtauraso:
- New hook added to downloadCutout._download() to collect failed chunks
- Retries exceeded behavior in downloadCutout._download() altered to default to continuoing the download. If callers of downloadCutout.download() want to stop the download, they can provide a failed_chunk_hook which raises
- Failed chunk collector class keeps failed sky locations and outputs them with astropy.table.Table.write()
- Cleanup documentation of downloadCutout._download and downloadCutout._download_chunk kwargs to also exist at top level in downloadCutout.download() for ease of discoverability.
- Cleanup: convert stats system to a context manager class to avoid proliferation of globals in download.py, and centralize stat pretty printing.

### DIFF
--- a/src/fibad/download.py
+++ b/src/fibad/download.py
@@ -5,6 +5,7 @@ import urllib.request
 from pathlib import Path
 from typing import Union
 
+import numpy as np
 from astropy.table import Table, hstack
 
 import fibad.downloadCutout.downloadCutout as dC
@@ -161,6 +162,9 @@ def create_rects(locations: Table, offset: int = 0, default: dC.Rect = None) -> 
         args["tract"] = str(args["tract"])
         # Sets the file name on the rect to be the object_id, also includes other rect fields
         # which are interpolated at save time, and are native fields of dc.Rect.
+        #
+        # This name is also parsed by FailedChunkCollector.hook to identify the object_id, so don't
+        # change it without updating code there too.
         args["name"] = f"{location['object_id']}_{{type}}_{{ra:.5f}}_{{dec:+.5f}}_{{tract}}_{{filter}}"
         rect = dC.Rect.create(default=default, **args)
         rects.append(rect)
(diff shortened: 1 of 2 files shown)


src/fibad/download.py
```diff
@@ -161,102 +162,177 @@ def create_rects(locations: Table, offset: int = 0, default: dC.Rect = None) ->
...
+            The exception that was thrown on the final attempt to request this chunk
+        attempts : int
+            The number of attempts that were made to request this chunk
+
+        """
+        print("Failed chunk handler got called.")
+
+        for rect in rects:
+            # Relies on the name format set in create_rects to work properly
+            self.object_id.append(int(rect.name.split("_")[0]))
```
mtauraso, line 302: So this actually has a bug.  If you try to request 10 locations across 5 filters, and a location fails, we will save it 5 times, so the resulting fits file isn't really able to be fed back into the program.

I'm going to add a commit to this PR which addresses this issue by de-duplicating sky location rects using object_id as a unique key.

src/fibad/download.py
```diff
@@ -269,15 +345,19 @@ def download_cutout_group(rects: list[dC.Rect], cutout_dir: Union[str, Path], us
...
             onmemory=False,
-            request_hook=request_hook,
+            request_hook=stats.hook,
+            failed_chunk_hook=failed_chunks.hook,
             resume=True,
             chunksize=10,
         )
 
-    print("")  # Print a newline so the stats stay and look pretty.
+        failed_chunks.save("failed_locations.fits", format="fits")
```
mtauraso, line 363: Also a bug here: If you are running a resumed download where failed_locations.fits already exists this will fail to write out the file.

Will be addressed in next commit.

mtauraso:
I've addressed the bugs I found on my own review of this, and tested it with resume functionality.

PR #30, opened 2024-08-16
Scaffolding for Fibad library instance object

mtauraso:
* All current verbs use a unified interface verb.run(config) with no args passed
* New Fibad object holds the fully merged config, hands it out to verbs
* CLI uses Fibad object to invoke verbs.
* fibad_default_config.toml is now part of the fibad package and is loaded as a default config always


src/fibad_cli/main.py
```diff
@@ -56,14 +32,8 @@ def main():
         parser.print_help()
         sys.exit(1)
 
-    fibad_action = f"fibad-{args.verb}"
-
-    # Ensure the action is available
-    if args.verb not in fibad_verbs:
```
mtauraso, line 62: I removed this because in practice you cannot hit this case. If you provide a verb other than what ArgumentParser is expecting, argumentparser will give you an entirely different error and exit.

drewoldag (approved):
Looks great. Merge it!!!! 🚢

PR #31, opened 2024-08-17
Initial version of logging for fibad.

mtauraso:
* Almost all print statements have been changed to logs.
* Logging is initialized in the Fibad object, where the config is parsed
* The logger name is 'fibad' which allows automatic roll-up from other modules within fibad (but not fibad_cli) by simply creating a local logging handle with:

  import logging logger = logging.getLogger(__name__)

* This mechanism also works for the downloadCutout code, which has been converted to logging.
* Added documentation to the default config about the log levels and configurable destinations.

### DIFF
--- a/src/fibad/fibad.py
+++ b/src/fibad/fibad.py
@@ -1,3 +1,5 @@
+import logging
+import sys
 from pathlib import Path
 from typing import Union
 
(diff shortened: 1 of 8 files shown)


src/fibad/fibad.py
```diff
@@ -23,8 +25,61 @@ def __init__(self, *, config_file: Union[Path, str] = None):
...
+        # Choosing "fibad" as our log name also means that modules like fibad.models, or fibad.dataloaders
+        # can get a logger with logging.getLogger(__name__) and will automatically roll up to us.
+        #
+        # A downside is that multiple Fibad objects all log the same place and have the ability to clobber
+        # one another's settings, and combine logs with one another.
+        self.logger = logging.getLogger("fibad")
+
+        # default to warning level
+        level = general_config.get("log_level", "warning")
+        if not isinstance(level, str):
```
drewoldag, line 43: The log setup is fine here for now, but I would advocate for moving the logging configuration to a separate module.

drewoldag (approved):
Looks good. One small comment, but doesn't need to be addressed now.

PR #33, opened 2024-08-22
Make `fibad train` work end-to-end with HSC cutouts

mtauraso:
HSC Data Loader and example Autoencoder to make `fibad train` work end-to-end on HSC data.


pyproject.toml
```diff
@@ -19,6 +19,7 @@ dependencies = [
     "astropy", # Used to load fits files of sources to query HSC cutout server
     "toml", # Used to load configuration files as dictionaries
     "torch", # Used for CNN model and in train.py
+    "torchvision", # Used in data loader and example autoencoder
```
drewoldag, line 22: If we're adding `torchvision` as a required dependency, we should probably remove it from the `examples` optional dependencies.

src/fibad/fibad_default_config.toml
```diff
@@ -40,18 +40,20 @@ weights_filepath = "example_model.pth"
...
+# Name of data loader to use
+name = "HSCDataLoader"
 
 # An example of requesting an external data loader class
 # external_class = "user_package.submodule.ExternalDataLoader"
 
 # Directory path where the data is stored
-path = "./data"
+# path = "./data"
+path = "../hscplay/cutouts/"
```
drewoldag, line 51: Minor point here, but if this is our default location for data, can we call it something slightly more descriptive? I would be ok with `../hsc_data/cutouts/'.

src/fibad/train.py
```diff
@@ -18,13 +18,24 @@ def run(config):
         dict
     """
 
+    data_loader_cls = fetch_data_loader_class(config)
+    fibad_data_loader = data_loader_cls(config.get("data_loader", {}))
+    data_loader = fibad_data_loader.get_data_loader()
+
     model_cls = fetch_model_class(config)
-    model = model_cls(config.get("model", {}))
+    model = model_cls(model_config=config.get("model", {}), shape=fibad_data_loader.shape())
```
drewoldag, line 26: This looks ok for now, two small points: 

1. I'm starting to wonder if there's a way that we can tidy this section up. Granted users won't interact much with this part of the code, but I wonder if instantiating the data loader inside of our model class might be a good thing. Nothing to change now, just something for us to consider.
2. Not a blocker, but I thought we were trying to shy away from names like `model_*`. What do you think of: `model = model_cls(config=config, ...)`?

drewoldag: For point 2, I'm also wondering if it makes sense to just pass the entire config dictionary as well to support point 1.

src/fibad/train.py
```diff
@@ -18,13 +18,24 @@ def run(config):
...
 
-    data_loader_cls = fetch_data_loader_class(config)
-    data_loader = data_loader_cls(config.get("data_loader", {})).get_data_loader()
+    cuda_available = torch.cuda.is_available()
+    mps_available = torch.backends.mps.is_available()
+
+    # We don't expect mps (Apple's Metal backend) and cuda (Nvidia's backend) to ever be
+    # both available on the same system.
+    device_str = "cuda:0" if cuda_available else "cpu"
+    device_str = "mps" if mps_available else "cpu"
```
drewoldag, line 34: This is nice 😄

src/fibad/data_loaders/hsc_data_loader.py
```diff
@@ -0,0 +1,370 @@
...
+
+        self.config.get("path", "./data")
+
+        # TODO: What will be a reasonable set of tranformations?
+        # For now tanh all the values so they end up in [-1,1]
+        # Another option might be sinh, but we'd need to mess with the example autoencoder module
+        # Because it goes from unbounded NN output space -> [-1,1] with tanh in its decode step.
+        transform = Lambda(lambd=np.tanh)
+
+        return HSCDataSet(self.config.get("path", "./data"), transform=transform)
```
drewoldag, line 48: Not to be addressed in this PR, but I think we're going to need to do something about the proliferation of `config.get("thing", <default>)`. We've committed to defining defaults in the config, it feels redundant and a potential source of confusion and mismatch to have a second default defined in the code. 

DataClasses, Pydantic, lazyconfig, etc...

pyproject.toml
drewoldag: I've probably just forgotten, is `torchvision` really a required dependency? I might be missing where it's used in this PR.

src/fibad/data_loaders/example_cifar_data_loader.py
```diff
@@ -12,6 +12,9 @@ class CifarDataLoader:
     def __init__(self, data_loader_config):
         self.config = data_loader_config
 
+    def shape(self):
```
drewoldag, line 15: Perhaps later we can look into a way to do this programmatically, since this particular dataset should have a very well defined shape.

drewoldag (approved):
Several comments left throughout, nothing scary though, mostly just thoughts about future work.

pyproject.toml
mtauraso: Good catch. Not having it as a required dep broke the build prior to this PR, not exactly sure where.

I'll go debug a bit more.

drewoldag: I would guess it would crop up during the testing of the hsc dataset. If you decide that torchvision should not be in the primary deps, then we could update the workflows to `pip install . [dev,examples]`. Leaving it as required dep is easier for now. TBD in the long run.

src/fibad/data_loaders/hsc_data_loader.py
mtauraso: Yeah, i have this worry every time I write `config.get()` Filed as https://github.com/lincc-frameworks/fibad/issues/50

src/fibad/train.py
mtauraso: Yeah each time I'm passing just one part of the config, I think maybe we should always pass the whole thing, and that would more easily allow the data loader to "see" a download path, and the model to "see" data loader params.

So far nothing has forced this, but it seems like passing the whole config everywhere maybe a good thing for us to just change today to prevent our current implementation choices from affecting the ergonomics of the config too much. Then we can be more free-form as we get feedback about the config to simply change its organization to make sense to the user.

Filed as https://github.com/lincc-frameworks/fibad/issues/51

mtauraso: As far as 1, yeah the creation of the data loader does seem a little awkward.

Also, most of the code I wanted to write for the hsc data loader ended up in an hsc data set helper class, so I feel like we probably picked the module boundary not quite right. I'd be happy to try it another way.

src/fibad/fibad_default_config.toml
mtauraso: IMO we should just stick with ./data as the default.  I put this directory in for testing and forgot to take it out. Good catch

pyproject.toml
mtauraso: You're right. Adding tests for hsc data set made this required. I'm going to leave it required and remove from example deps for now.

PR #41, opened 2024-08-22
Adding configs for retries, retrywait, timeout, and chunksize

mtauraso:
Pretty straightforward addition of these configs with documentation.

### DIFF
--- a/src/fibad/fibad_default_config.toml
+++ b/src/fibad/fibad_default_config.toml
@@ -30,6 +30,16 @@ cutout_dir = "../hscplay/cutouts/"
 offset = 0
 num_sources = 10
 
+# These control the downloader's HTTP requests and retries
+# `retry_wait` How long to wait before retrying a failed HTTP request in seconds. Default 30s
+retry_wait = 30
+# `retries` How many times to retry a failed HTTP request before moving on to the next one. Default 3 times
+retries = 3
+# `timepout` How long should we wait to get a full HTTP response from the server. Default 3600s (1hr)
+timeout = 3600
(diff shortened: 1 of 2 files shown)


src/fibad/fibad_default_config.toml
```diff
@@ -30,6 +30,16 @@ cutout_dir = "../hscplay/cutouts/"
...
 
+# These control the downloader's HTTP requests and retries
+# `retry_wait` How long to wait before retrying a failed HTTP request in seconds. Default 30s
+retry_wait = 30
+# `retries` How many times to retry a failed HTTP request before moving on to the next one. Default 3 times
+retries = 3
+# `timepout` How long should we wait to get a full HTTP response from the server. Default 3600s (1hr)
+timeout = 3600
+# `chunksize` How many sky location rectangles should we request in a single request. Default is 990
+chunksize = 990
```
drewoldag, line 41: I must have missed something, why 990? Is this a magic number?

drewoldag (approved):
Looks fine to me. One small question, not a blocker.

src/fibad/fibad_default_config.toml
mtauraso: 990 is the "magic" number that is the default that `downloadCutout.py` hardcodes.

The HSC docs report that this number can be [as high as 100,000](https://hsc-release.mtk.nao.ac.jp/das_cutout/pdr3/manual.html#limitations-number). 

I haven't experimented much with higher numbers here, I suspect there is some reason that 990 got committed on HSC side, but I have no ida.

drewoldag: Hmm, ok, weird. Thank for clarifying.

PR #43, opened 2024-08-23
Getting Started notebook for downloader and Logging fixup

mtauraso:
The issue about python logging being global vs our library having an instance object caused weird notebook behavior, which is resolved in this commit with an extensive comment on the chosen solution.


src/fibad/fibad.py
```diff
@@ -16,32 +16,72 @@ class Fibad:
...
-    def __init__(self, *, config_file: Union[Path, str] = None):
+    def __init__(self, *, config_file: Union[Path, str] = None, setup_logging: bool = True):
         """Initialize fibad. Always applies the default config, and merges it with any provided config file.
 
         Parameters
         ----------
         config_file : Union[Path, str], optional
             filename or pathlib.Path to a config file, by default None
+
+        setup_logging : bool, optional
```
drewoldag, line 27: It's good that there's a substantial explanation in the comments below, but could we percolate some of that back up here to the docstring? In particular, I could imagine that a user looking at the autogenerated documentation would like to know the difference in the output of true vs false here.

drewoldag (approved):
Looks good, the only request is to put a little more detail in the docstring so that autodoc generation picks it up.

src/fibad/fibad.py
mtauraso: Added some better docs, pushing this to main.

PR #46, opened 2024-08-23
Fix divide by zero in stats module bandwidth calculation

mtauraso:
We nopw print 0 MB/s or 0 files/s for bandwidth measurements where the duration of the measurement is zero microseconds. This occurs in certain error cases where no data was actually transferred, and two timing calls hit in the same microsecond.


drewoldag (approved):
Looks good to me.

PR #52, opened 2024-08-27
Remove Michael's password from default config file

mtauraso:
- Also make it easier to run fibad commands by using ./fibad_config.toml as the user-supplied config in situations where no config is specified explicitly

An example config you will probably need after this change merges looks something like:
```
[download]
username = "mtauraso@local"
password = "cCw+nX53lmNLHMy+JbizpH/dl4t7sxljiNm6a7k1"
fits_file = "../hscplay/temp.fits"
cutout_dir = "../hscplay/cutouts/"

[data_loader]
name = "HSCDataLoader"
path = "../hscplay/cutouts"
```

For downloader and data loader.

If this merges I'm planning to reset my HSC password shortly thereafter in order to avoid the current situation where our pubic repo has live credentials to their system.

### DIFF
--- a/src/fibad/config_utils.py
+++ b/src/fibad/config_utils.py
@@ -4,6 +4,7 @@ from typing import Union
 import toml
 
 DEFAULT_CONFIG_FILEPATH = Path(__file__).parent.resolve() / "fibad_default_config.toml"
+DEFAULT_USER_CONFIG_FILEPATH = Path.cwd() / "fibad_config.toml"
 
 
 def get_runtime_config(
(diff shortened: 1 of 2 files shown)


src/fibad/config_utils.py
```diff
@@ -38,6 +39,10 @@ def get_runtime_config(
     with open(default_config_filepath, "r") as f:
         default_runtime_config = toml.load(f)
 
+    # If a named config exists in cwd, and no config specified on cmdline, use cwd.
+    if runtime_config_filepath is None and DEFAULT_USER_CONFIG_FILEPATH.exists():
+        runtime_config_filepath = DEFAULT_USER_CONFIG_FILEPATH
```
drewoldag, line 44: I'm guessing that this is meant to be a convenience so that you don't have to pass in a user config, yeah? If that's the case, I can understand where you're coming from. 

However, it also feels like it might open a door for unintended settings. i.e. A user has a fibad_config.toml and a fibad_config2.toml, and fails to pass the config when they want to use the "2" version. 

Maybe I'm misinterpreting the utility here though. Let me know if I'm missing a broader point.

drewoldag:
Generally this looks fine, but I would like to get your response to the comment I left, just to make sure I understand the intention.

src/fibad/config_utils.py
mtauraso: There's not a broader point to this part beyond the convenience of not passing `--runtime-config` or `config=` every time.

PR #53, opened 2024-08-27
Add a configurable filter list to HSC data set

mtauraso:


src/fibad/data_loaders/hsc_data_loader.py
```diff
@@ -109,20 +120,36 @@ def __init__(
...
+
         files = {}
         # Go scan the path for object ID's so we have a list.
         for filepath in Path(self.path).glob("[0-9]*.fits"):
             filename = filepath.name
-            m = re.match(r"([0-9]{17})_.*\_(HSC-[GRIZY]).fits", filename)
+            m = re.match(full_regex, filename)
+
+            # Skip files that don't match the pattern.
+            if m is None:
```
aritraghsh09, line 150: If it doesn't make the process super slow, can we log the name of the file being skipped?

aritraghsh09 (approved):
Looks goone! One minor comment.

src/fibad/data_loaders/hsc_data_loader.py
mtauraso: I'm more worried about log spam. Adding a debug or info level log here shouldn't slow things down unless the log is being emitted to a console.

I am thinking though that the better solution is to output that manifest fits table, which will have all the skipped files explicitly and not create a potential foot-gun for people changing the logging level to info/debug.

Thoughts?

drewoldag: I would advocate for @mtauraso's approach here. Perhaps a middle ground would be logging some summary metrics at the end along with a message saying to look in the manifest fits table for skipped files?

aritraghsh09: Yes, putting the info in the manifest table sounds good! I also like @drewoldag's idea of some summary metrics at the end if it's easy to implement!

PR #54, opened 2024-08-28
First implementation with pytorch ignite to support cpu->multi-gpu scaling

drewoldag:
First attempt at using pytorch ignite to support the promise of scaling from cpu->gpu->multi-gpu.


src/fibad/fibad_default_config.toml
drewoldag, line 1: Modified this file just to make training a bit more straightforward - i.e. no need to download HSC data, the model is a bit more simple, etc.

mtauraso (approved):

PR #57, opened 2024-09-05
Multithreaded manifest-based downloader

mtauraso:
- Support for multiple connections via 'concurrent_connections' config
- Replaced FailedChunkCollector and old resume system in downloadCutout.py with manifest-based resume system
- Added __hash__ and __eq__ to downloadCutout Rects so they can be dictionary keys
- Pulled downloader functions into a class so class constants could be namespaced

### DIFF
--- a/src/fibad/download.py
+++ b/src/fibad/download.py
@@ -1,11 +1,10 @@
-import contextlib
 import datetime
 import itertools
 import logging
-import os
+import time
 import urllib.request
 from pathlib import Path
-from threading import Thread
+from threading import Lock, Thread
 from typing import Optional, Union
 
 from astropy.table import Table, hstack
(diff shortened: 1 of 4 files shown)


aritraghsh09:
@mtauraso  I was testing this out and ran a few examples. In `manifest.fits`, all entries of the `sh` and `sw` columns are set to `0.0061111111111111` when the specification set in the configuration file is `22asec` for both. Is this expected behavior? Is some unit conversion happening in the backend?

src/fibad/download.py
```diff
@@ -38,143 +35,430 @@ def working_directory(path: Path):
...
+        # Early return if there is nothing to download.
+        if len(rects) == 0:
+            logger.info("Download already completed according to manifest.")
+            return
 
-def rect_from_config(config: dict) -> dC.Rect:
-    """Takes our runtime config and loads cutout config
-    common to all cutouts into a prototypical Rect for downloading
+        # Create thread objects for each of our worker threads
+        num_threads = config.get("concurrent_connections", 2)
```
aritraghsh09, line 103: The default `.toml` has a `max_connections` parameter. Is this different from `concurrent_connections`? Since this is an important parameter, might be good to put in the default`.toml` as well.

src/fibad/download.py
```diff
@@ -38,143 +35,430 @@ def working_directory(path: Path):
...
+    # must be provided by config.
+    VARIABLE_FIELDS = ["tract", "ra", "dec"]
 
-    fits_file = config.get("fits_file", "")
-    logger.info(f"Reading in fits catalog: {fits_file}")
-    # Filter the fits file for the fields we want
-    column_names = ["object_id"] + variable_fields
-    locations = filterfits(fits_file, column_names)
+    # These are the column names we retain when writing a rect out to the manifest.fits file
+    RECT_COLUMN_NAMES = VARIABLE_FIELDS + ["filter", "sw", "sh", "rerun", "type"]
```
aritraghsh09, line 59: Although for the downloads we will be doing, "filter", "sh", "sw", "rerun", "type" will be set to the same value for the whole download; this might not be the case for other situations. 

I think it would be best to merge this code as-is and open a low-priority feature-request issue so that the downloader can support variable  "filter", "sh", "sw", "rerun", "type".

aritraghsh09:
> @mtauraso I was testing this out and ran a few examples. In `manifest.fits`, all entries of the `sh` and `sw` columns are set to `0.0061111111111111` when the specification set in the configuration file is `22asec` for both. Is this expected behavior? Is some unit conversion happening in the backend?

I delved into the code, and it seems like that in the manifest file, `sh` and `sw` are being written in degrees. This behaviour is fine as long as we make this abundantly clear in our docs.

aritraghsh09 (approved):
This looks good to me! Please take a look at/resolve comments prior to merging.

src/fibad/download.py
mtauraso: I Agree. Filed: https://github.com/lincc-frameworks/fibad/issues/58

mtauraso:
> > @mtauraso I was testing this out and ran a few examples. In `manifest.fits`, all entries of the `sh` and `sw` columns are set to `0.0061111111111111` when the specification set in the configuration file is `22asec` for both. Is this expected behavior? Is some unit conversion happening in the backend?
> 
> I delved into the code, and it seems like that in the manifest file, `sh` and `sw` are being written in degrees. This behaviour is fine as long as we make this abundantly clear in our docs.

This is indeed what is happening, I'll add some documentation of the manifest file format so its super clear

src/fibad/download.py
mtauraso: Yeah this is new and different. Really `max_connections` ought not to have been added in the first place.

I'll address this in a followup PR, because I already have several race-condition fixes in my local checkout that are all fixups to this PR.

mtauraso: Actually strike that, I'm jut going to roll all the fixes into this PR.

PR #60, opened 2024-09-10
Proper multithreaded stats

mtauraso:
- Stats have their own thread for printing at a configurable cadence
- There are 1 hr trailing window and cumulative stats
- File rate is now reported as files/24h
- Stats live in a separate file from the downloader
- Stats code is (hopefully) easier to read

### DIFF
--- a/src/fibad/download_stats.py
+++ b/src/fibad/download_stats.py
@@ -1,0 +1,390 @@
+import copy
+import datetime
+import functools
+import logging
+import time
+import urllib.request
+from threading import Lock, Thread
+from typing import Optional, Union
+
+logger = logging.getLogger(__name__)
+
+
+class StatRecord:
+    """Recording object that represents one or many stats measurements reported by download request threads.
+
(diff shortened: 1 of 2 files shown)


src/fibad/download_stats.py
```diff
@@ -0,0 +1,390 @@
...
+        Parameters
+        ----------
+        log_level : int
+            The logging level to emit the log message at.
+        num_threads : int
+            Number of threads actively reporting into the stats object we want to print, by default 1
+        prefix : str
+            String to prefix all log messages with. "" by default
+        """
+        # Some things are calculated in print because we need the number of threads.
```
drewoldag, line 253: Is this comment still needed? Perhaps I'm missing it, but it seems like you've done a good job of encapsulating the calculations.

drewoldag (approved):
This all looks pretty good to me. I like the abstraction that was done here.

aritraghsh09 (approved):
This looks good!

PR #62, opened 2024-09-13
Various Fixups for klone download

mtauraso:
- Added configs so image/variance/mask layer downloads can be configured.
- Put filter lower in the sort order for dC.Rect so that duplicate dowloading in a resume situation is minimized for large downloads.
- Concurrent connection error is now correct to our understanding of HSC's server-side behavior.

### DIFF
--- a/src/fibad/download.py
+++ b/src/fibad/download.py
@@ -71,9 +71,10 @@ class Downloader:
             return
 
         # Create thread objects for each of our worker threads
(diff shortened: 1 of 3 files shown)


aritraghsh09 (approved):
This looks good. 
While reading the code, an enhancement came to mind which won't be needed for our work, but might be important later for other users. 
I have opened the issue here -- https://github.com/lincc-frameworks/fibad/issues/63

PR #67, opened 2024-09-19
Remove `external_class` as config parameter

drewoldag:
Removing `external_class` as a config parameter, and checking if `name` is a key the registry or a libpath.

In the case of an externally defined model and config file, if the user specified the external_class as the model to use, we would instead default to the model defined in the `name` key of the `fibad_default_config.toml` file. Clearly not what the user would expect. 

So for now, we allow `name` tobe either a key into the registry dictionary _or_ a libpath string for an external model class.

This isn't quite where I would like it. There are plenty of exceptions raised, but it would be nice to add some simple logging as well.

### DIFF
--- a/src/fibad/plugin_utils.py
+++ b/src/fibad/plugin_utils.py
@@ -27,22 +27,22 @@ def get_or_load_class(config: dict, registry: dict) -> type:
         a `name` nor `external_cls` key was found in the config.
     """
 
-    # User specifies one of the built in classes by name
+    #! Once we have confidence in the config having default values, we can remove this check
     if "name" in config:
         class_name = config.get("name")
+        returned_class = None
 
-        if class_name not in registry:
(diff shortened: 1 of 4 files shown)


mtauraso (approved):
LGTM

src/fibad/plugin_utils.py
```diff
@@ -27,22 +27,22 @@ def get_or_load_class(config: dict, registry: dict) -> type:
         a `name` nor `external_cls` key was found in the config.
     """
 
-    # User specifies one of the built in classes by name
+    #! Once we have confidence in the config having default values, we can remove this check
     if "name" in config:
         class_name = config.get("name")
```
mtauraso, line 32: Since the new config stuff went in you will probably have to make this `config[name]` to get it working again.

PR #68, opened 2024-09-19
Centralize config defaults, add safety for Config updates

mtauraso:
- All configuration dicts are now ConfigDict. A ConfigDict is a version of dict that raises RuntimeError quite a lot when you try to mutate it or when you access values that weren't set at construction time.
- There is a new validate step for runtime config that ensures every config key that appears at runtime has a default defined in fibad_default_config.toml.
- Usages of .get() on config dictionaries have been removed, so all defaults now live in fibad_default_config.toml. Config access is entirely via __getitem__ / [].
- The sum total of the above means that we expect any new config which is missing a default value to generate a runtime error as early as possible if it is referenced anywhere in code, or in a user-supplied configuration.
- Several config values have changed representation, because "not specified" was their default. toml does not support "None" so all of these values have been set to false in fibad_default_config.toml. Accesses to these configs have been updated to be aware that the default value of 'false' either triggers modal behavior (data_loader/crop_to, data_loader/filters) or triggers a RuntimeError to prompt the user to provide it (download/username, download/password).
- Except where necessary for model/dataloader class loading, passsing of individual config sections has been removed. For the most part all parts of the code now have access to the entire runtime configuration.
- The data_loader/data and download/cutout_dir configs have been combined into general/data_dir, to avoid an invalid configuration state where the downloader and data loader are looking at different directories for data.

### DIFF
--- a/src/fibad/config_utils.py
+++ b/src/fibad/config_utils.py
@@ -108,6 +108,10 @@ def _validate_runtime_config(runtime_config: ConfigDict, default_config: ConfigD
             raise RuntimeError(msg)
 
         if isinstance(runtime_config[key], dict):
+            if not isinstance(default_config[key], dict):
+                msg = f"Runtime config contains a section named {key} which is the name of a value in the "
+                msg += "default config. Please choose another name for this section."
+                raise RuntimeError(msg)
             _validate_runtime_config(runtime_config[key], default_config[key])
 
 
--- a/src/fibad/config_utils.py
+++ b/src/fibad/config_utils.py
@@ -1,3 +1,4 @@
+import logging
 from pathlib import Path
 from typing import Union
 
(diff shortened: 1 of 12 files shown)


src/fibad/config_utils.py
```diff
@@ -6,6 +7,160 @@
...
+        Raised if any config that exists in the runtime config does not have a default defined in
+        default_config
+    """
+    for key in runtime_config:
+        if key not in default_config:
+            msg = f"Runtime config contains key or section {key} which has no default defined."
+            msg += f"All configuration keys and sections must be defined in {DEFAULT_CONFIG_FILEPATH}"
+            raise RuntimeError(msg)
+
+        if isinstance(runtime_config[key], dict):
```
drewoldag, line 110: Do you think it would be worth validating that if runtime_config[key] is a dict, that default_config[key] is as well?

src/fibad/config_utils.py
```diff
@@ -6,6 +7,160 @@
...
+class ConfigDict(dict):
+    """The purpose of this class is to ensure key errors on config dictionaries return something helpful.
+    and to discourage mutation actions on config dictionaries that should not happen at runtime.
+    """
+
+    # TODO: Should there be some sort of "bake" method which occurs after config processing, and
+    # percolates down to nested ConfigDicts and prevents __setitem__ and other mutations of dictionary
+    # values? i.e. a method to make a config dictionary fully immutable (or very difficult/annoying to
+    # mutuate) before we pass control to possibly external module code that is relying on the dictionary
+    # to be static througout the run.
```
drewoldag, line 22: This comment seems like a really good idea. It will be interesting to see, when we have an actual user(s) interacting with Fibad if there is a desire to modify the merged config prior to running train or predict. But I agree, that during runtime, the config should be/becomes immutable.

drewoldag (approved):
This looks good to me.

src/fibad/config_utils.py
mtauraso: Yeah this is reasonable paranoia.

src/fibad/config_utils.py
mtauraso: Agree, I do feel like its complicated enough that we should hold off for now.

PR #70, opened 2024-09-20
Predict and Output scaffolding

mtauraso:
- Predict basically works
- Most Pytorch ignite stuff has been factored out
- The results directory also basically works
- Predict is not yet outputting its results

### DIFF
--- a/src/fibad/config_utils.py
+++ b/src/fibad/config_utils.py
@@ -257,7 +257,7 @@ def create_results_dir(config: ConfigDict, prefix: Union[Path, str]) -> Path:
         The path created by this function
     """
     results_root = Path(config["general"]["results_dir"]).resolve()
-    timestamp = datetime.datetime.now().strftime("%Y%m%d-%H%m%S")
+    timestamp = datetime.datetime.now().strftime("%Y%m%d-%H%M%S")
     directory = results_root / f"{prefix}-{timestamp}"
     directory.mkdir(parents=True, exist_ok=False)
     return directory
--- a/src/fibad/config_utils.py
+++ b/src/fibad/config_utils.py
@@ -1,3 +1,4 @@
+import datetime
 import logging
 from pathlib import Path
 from typing import Union
@@ -203,9 +204,6 @@ def get_runtime_config(
     else:
         final_runtime_config = default_runtime_config
 
-    # ~ Uncomment when we have a better place to stash results.
-    # log_runtime_config(final_runtime_config)
-
     return final_runtime_config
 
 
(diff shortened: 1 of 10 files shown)


src/fibad/config_utils.py
```diff
@@ -236,16 +234,46 @@ def merge_configs(default_config: dict, user_config: dict) -> dict:
...
+        The verb name of the run.
+
+    Returns
+    -------
+    Path
+        The path created by this function
+    """
+    results_root = Path(config["general"]["results_dir"]).resolve()
+    timestamp = datetime.datetime.now().strftime("%Y%m%d-%H%m%S")
+    directory = results_root / f"{prefix}-{timestamp}"
```
drewoldag, line 261: Perhaps a trivial question here, but I wanted to get your thoughts. When we were first talking about this, I envisioned a directory scheme like this: 
```
/results
   /train
   /predict
   /<verb>
```

I like what you have here more. But I'm also curious about the need to scan the directory listing and see the collated run history if that makes sense. 

By default we'll get alphabetized runs - would it be helpful to name the directories `{timestamp}-{verb}` to maintain the chronology? Maybe not worth it, since the user can just `ls -haltr` and get the directory contents sorted by creation.

mtauraso: I did actually run into this during testing, where I wanted timestamp-first ordering because I needed not the current run but the previous run, and it was a little annoying with the verb being first.

I'm going to put the timestamp first, and keep things flat for now.

src/fibad/data_loaders/hsc_data_loader.py
```diff
@@ -356,7 +356,7 @@ def _get_file(self, index: int) -> Path:
         filter = filter_names[index % self.num_filters]
         return self._file_to_path(filters[filter])
 
-    def _all_object_ids(self):
+    def ids(self):
         """Private read-only iterator over all object_ids that enforces a strict total order across
```
mtauraso, line 360: NIT: Docstring should say Public not private

src/fibad/pytorch_ignite.py
```diff
@@ -0,0 +1,194 @@
...
+from torch.utils.data import Dataset
+
+from fibad.config_utils import ConfigDict
+from fibad.data_loaders.data_loader_registry import fetch_data_loader_class
+from fibad.models.model_registry import fetch_model_class
+
+logger = logging.getLogger(__name__)
+
+
+def setup_model_and_dataset(config: ConfigDict) -> tuple:
```
mtauraso, line 18: The astute reader will notice that the `data_loader` constructed in this function never leaves!

Both `train` and `predict` construct their real dataloader with a separate call to`dist_data_loader`

I think the real thing we want researchers to swap out is the data set, and I'd suggest:
* data_loader registry becomes data_set registry
* We add a `[data_set]` config section for `name=HSCDataSet` or the like
* we keep the `[data_loader]` config section to define data loader things (batch size, shuffling, num_workers)

src/fibad/train.py
drewoldag, line 1: I love how simple this file has become :)

src/fibad/predict.py
```diff
@@ -1,19 +1,83 @@
...
+                RuntimeError("The path to save results for object {object_id} already exists.")
+            np.save(savepath, tensor.numpy(), allow_pickle=False)
+            write_index += 1
+
+    evaluator = create_evaluator(model, _save_batch)
+    evaluator.run(data_loader)
+    logger.info("finished evaluating...")
+
+
+def load_model_weights(config: ConfigDict, model):
```
drewoldag, line 53: This looks good. I might look at lifting it a level so that it can be used by `train` as well when we're restarting a training run.

drewoldag (approved):
This looks great! Thanks for putting this together

src/fibad/predict.py
mtauraso: oohh, that is a good idea for it to end up elsewhere.

PR #71, opened 2024-09-23
Initial commit to consolidate name and external_cls.

drewoldag:
After a botched attempt to rebase from main, I've reimplemented the work from PR #67. This does the same work. FWIW, required changes will be ported over to the only external package, kbmod-ml.


mtauraso (approved):
LGTM

PR #76, opened 2024-09-27
Switching to Dataset based modularity

mtauraso:
- data_loader_registry is now data_set_registry
- Everything custom data loader is now data set
- [data_loader] config now read in only one place (when we make the pytorch ignite data loader)
- [data_set] config used to set which data set you are using.
- train and predict verbs work.


drewoldag (approved):
Cool, this looks like it's moving things in the right direction. Thanks!

PR #79, opened 2024-10-01
Updating to latest PPT version

drewoldag:


.copier-answers.yml
```diff
@@ -1,10 +1,23 @@
...
+create_example_module: false
+custom_install: true
+enforce_style:
+- ruff_lint
+- ruff_format
+failure_notification: []
+include_benchmarks: true
+include_docs: true
+include_notebooks: true
+mypy_type_checking: none
```
mtauraso, line 15: We have a lot of type annotations. Do we want to turn type checking on as a way to keep that habit?

(I ask not knowing if this does what I think it does or not 😄 )

.github/workflows/build-documentation.yml
```diff
@@ -1,3 +1,4 @@
+
```
mtauraso, line 1: I'm sad because I think I know why this line has been added, and I think it is a bug in my recently-merged PPT change.

mtauraso (approved):

.copier-answers.yml
drewoldag: Hmm, it could be a beneficial thing to do so. It will slow down committing code, but the value add might be appreciated by the end users. I'll create an issue for it.

PR #84, opened 2024-10-07
Fixing CNN `shape` issue

drewoldag:
Mostly this fixes the "missing `shape` parameter" error that was popping up with the example CNN model. Also adds a little demo notebook for training a model.


src/fibad/models/example_cnn_classifier.py
```diff
@@ -16,7 +16,7 @@
 
 @fibad_model
 class ExampleCNN(nn.Module):
-    def __init__(self, config, _):
+    def __init__(self, config, shape):
```
mtauraso, line 19: Was this the change that fixed the bug?

mtauraso (approved):
Looks good to me, but then again so did the last version which was apparently broken 🤷‍♂️

src/fibad/models/example_cnn_classifier.py
drewoldag: I don't think so, this is just left over because I was also getting errors when passing shape.

drewoldag:
For the time being, I don't think that this PR should be merged. Going to leave it lingering for a bit, just to see if this problem returns.

src/fibad/models/example_cnn_classifier.py
drewoldag: FWIW, though, I do think that this needs to be either `shape` or `**kwargs` I guess... Otherwise you get :

```
Traceback (most recent call last):
  File "/home/drew/miniconda3/envs/fibad/bin/fibad", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/drew/code/fibad/src/fibad_cli/main.py", line 34, in main
    getattr(fibad_instance, args.verb)()
  File "/home/drew/code/fibad/src/fibad/fibad.py", line 166, in train
    return run(config=self.config, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/drew/code/fibad/src/fibad/train.py", line 22, in run
    model, data_set = setup_model_and_dataset(config)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/drew/code/fibad/src/fibad/pytorch_ignite.py", line 40, in setup_model_and_dataset
    model = model_cls(config=config, shape=data_set.shape())
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: ExampleCNN.__init__() got an unexpected keyword argument 'shape'
```

mtauraso: Ah, I understand now.

PR #90, opened 2024-10-09
Implement checkpointing during training

drewoldag:
This is the beginning of implementing checkpointing for training models. 

Need to pass in the config dictionary so that we can specify were to place the checkpoints, and read in a given checkpoint if the user chooses to do so. 

Also need to implement the ability to create a checkpoint at the "last best <metric>" value.


aritraghsh09 (approved):
This looks good to me! I especially like the best checkpoint logging feature!

Issue #98, opened 2024-10-22
Download Resume does not function when image or variance is set to true in fibad config


mtauraso:
If you have a config that downloads mask or variance, resume will fail to work.

The cause is that the image, variance, and mask rect parameters aren't saved to the manifest file, so on resume `_prune_downloaded_rects()` cannot match the current manifest with the rects generated from the catalog. In this case the rects generated from the catalog have config-provided values for image, mask and variance. The rects generated from reading the manifest have the Rect default values of image=True, mask=False and variance=False.  

To the extent that the config does not match these rects, none of the existing downloaded images will match as being part of the current catalog, and the resuming download will RuntimError.

The solution is to change `Downloader.RECT_COLUMN_NAMES` to include these fields (and all the other rect immutable_fields) so that errors of this type cannot happen.

PR #99, opened 2024-10-22
Include all immutable dc.Rect fields in manifest so Resume works in all configs.

mtauraso:
This is a breaking change for the manifest file format, in that downloads started before this change will not be able to resume if resumed using a version of fibad after this change.

Prior to this change If you have a config that downloads mask or variance, resume would simply fail (issue #98)

The cause was that the image, variance, and mask rect parameters weren't saved to the manifest file, so on resume _prune_downloaded_rects() could not match the  manifest with the rects generated from the catalog.

In this case the rects generated from the catalog had config-provided values for image, mask and variance. The rects generated from reading the manifest had the dc.Rect default values of image=True, mask=False and variance=False.

This change ensures that all the dc.Rect immutable fields, which are used in equality and hash checking are written out to the manifest.

### DIFF
--- a/src/fibad/download.py
+++ b/src/fibad/download.py
@@ -22,7 +22,11 @@ class Downloader:
     VARIABLE_FIELDS = ["tract", "ra", "dec"]
 
     # These are the column names we retain when writing a rect out to the manifest.fits file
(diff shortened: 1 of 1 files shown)


drewoldag (approved):
Looks ok to me

PR #100, opened 2024-10-22
Handle externally defined default configuration files 

drewoldag:
Adding a ConfigManager class to help wrangle the various defaults that could be imported.

The approach we're implementing is the following.
* Load the fibad_default_config
* Load the user_specific_config (USC), if defined
* Scan the USC for any external libraries, they look like lib specs under the `name` keys. i.e. `"module.submod.class"`
* Collect the file paths to default config files that are defined in the external libraries. For now, we assume the files are called `default_config.toml`.
* Merge all the external default configs to create a merged-external-default-config
* Merge the fibad default and the merged-external default configs to create a total-default-config
* Merge the USC into the total-default-config
* Validate the USC against the total-default-config

There's lots of clean up still to do.

### DIFF
--- a/src/fibad/config_utils.py
+++ b/src/fibad/config_utils.py
@@ -1,4 +1,5 @@
 import datetime
+import importlib
 import logging
 from pathlib import Path
 from typing import Union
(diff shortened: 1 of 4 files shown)


src/fibad/fibad.py
```diff
@@ -43,12 +43,14 @@ def __init__(self, *, config_file: Union[Path, str] = None, setup_logging: bool
             - You have a single Fibad object in use at any time. This is true in most notebook like
               environments.
         """
-        self.config = get_runtime_config(runtime_config_filepath=config_file)
+        self.config_manager = ConfigManager(runtime_config_filepath=config_file)
+        self.config = self.config_manager.config
+        # self.config = get_runtime_config(runtime_config_filepath=config_file)
```
mtauraso, line 48: I think we can remove `get_runtime_config`

src/fibad/fibad.py
```diff
@@ -43,12 +43,14 @@ def __init__(self, *, config_file: Union[Path, str] = None, setup_logging: bool
...
 
         # TODO: For now this is part of Fibad.__init__() In future when external modules can define their own
         # configuration keys and associated default values, this will need to occur after those external
         # modules have been resolved and loaded. This is why it is separate from get_runtime_config()
-        validate_runtime_config(self.config)
+        # validate_runtime_config(self.config)
```
mtauraso, line 53: We can probably also remove `validate_runtime_config` and `merge_configs` It does not look like either is called anymore.

src/fibad/config_utils.py
```diff
@@ -62,6 +63,141 @@ def clear(self):
...
+
+        self.overall_default_config = self.fibad_default_config.copy()
+
+        for path in self.external_library_config_paths:
+            external_library_config = self._read_runtime_config(path)
+            self.overall_default_config = self._merge_configs(
+                self.overall_default_config, external_library_config
+            )
+
+    def _merge_configs(self, default_config: dict, overriding_config: dict) -> dict:
```
mtauraso, line 140: This could be `@staticmethod` AFAICT it doesn't read or modify any data member.

src/fibad/config_utils.py
```diff
@@ -62,6 +63,141 @@ def clear(self):
...
+        final_config = default_config.copy()
+        for k, v in overriding_config.items():
+            if k in final_config and isinstance(final_config[k], dict) and isinstance(v, dict):
+                final_config[k] = merge_configs(default_config[k], v)
+            else:
+                final_config[k] = v
+
+        return final_config
+
+    def _validate_runtime_config(self, runtime_config: ConfigDict, default_config: ConfigDict):
```
mtauraso, line 166: Also could be `@staticmethod`

src/fibad/config_utils.py
```diff
@@ -62,6 +63,141 @@ def clear(self):
...
+        for key, value in runtime_config.items():
+            if isinstance(value, dict):
+                default_configs |= self._find_external_library_default_config_paths(value)
+            else:
+                if key == "name" and "." in key:
+                    external_library = value.split(".")[0]
+                    if importlib.util.find_spec(external_library) is not None:
+                        try:
+                            lib = importlib.import_module(external_library)
+                            lib_default_config_path = Path(lib.__file__).parent / "default_config.toml"
```
mtauraso, line 117: A thought, which I don't think should block any of this.

How do external library writers discover this magic path?
Do we want to provide any flexibility here for library writers to tell us where their file is? 

Short term I think the answers are "docs" and "no", but long term the second one might be "yes"

src/fibad/config_utils.py
```diff
@@ -62,6 +63,141 @@ def clear(self):
...
+            self.user_specific_config
+        )
+
+        self.overall_default_config = {}
+        self._merge_defaults()
+
+        self.config = self._merge_configs(self.overall_default_config, self.user_specific_config)
+        self.is_valid_config = self._validate_runtime_config(self.config, self.overall_default_config)
+
+    def _read_runtime_config(self, config_filepath: Union[Path, str] = DEFAULT_CONFIG_FILEPATH) -> ConfigDict:
```
mtauraso, line 87: Another potential `@staticmethod`

src/fibad/config_utils.py
```diff
@@ -62,6 +63,141 @@ def clear(self):
...
+        with open(config_filepath, "r") as f:
+            parsed_dict = toml.load(f)
+            return ConfigDict(parsed_dict)
+
+    def _find_external_library_default_config_paths(self, runtime_config: dict) -> set:
```
mtauraso, line 92: Another potential `@staticmethod`

src/fibad/config_utils.py
drewoldag: The pie-in-the-sky version of this, for me, would include a PPT-like template, or template repository that an external library write would start with, that would provide the stub for this. 

But, you're right, that this is a magic path right now. And yes, in the future, allowing flexibility here would be nice.

mtauraso (approved):
✅

PR #103, opened 2024-10-23
Set default models to `training` mode in the `train_step`. 

drewoldag:
It turns out that we have to explicitly set the model to `training` mode to actually train it. Oddly this isn't mentioned in the pytorch example code that was used when adding the ExampleCNN model. 

For now, I've added `model.train()` to the `create_trainer` and `model.eval()` to the `create_evaluator` functions. 

## Confusion matrix before adding `model.train()`
Screenshot 2024-10-23 at 2 55 11 PM

## Confusion matrix after adding `model.train()` 10 epochs
Screenshot 2024-10-23 at 2 42 03 PM

## Confusion matrix after adding `model.train()` 50 epochs
Screenshot 2024-10-23 at 3 25 17 PM
Overall accuracy for this is about 64%, which is a little better than the pytorch example code that got to about 54%.

### DIFF
--- a/src/fibad/models/example_autoencoder.py
+++ b/src/fibad/models/example_autoencoder.py
@@ -119,9 +119,6 @@ class ExampleAutoencoder(nn.Module):
         Current loss value
             The loss value for the current batch.
         """
-        # Set the model to train mode
-        self.train()
-
         # When we run on a supervised dataset like CIFAR10, drop the labels given by the data loader
         x = batch[0] if isinstance(batch, tuple) else batch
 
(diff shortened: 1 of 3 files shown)


mtauraso (approved):

drewoldag:
Yeah, I'm not sure what the best way to test this would be. It seems like there is an attribute `model.training` that is a boolean that we could use as a signal. I'm surprised by this, it feels like a gotcha that isn't obviously documented in the pytorch getting started guides (and I would have assumed it would be)

drewoldag:
new issue here: https://github.com/lincc-frameworks/fibad/issues/104

PR #105, opened 2024-10-23
Splits for HSCDataSet

mtauraso:
- HSCDataSet has been significantly modified to implement splits
- Constructors for fibad_data_sets now take a split argument
- Tests pass, but new functionality not yet tested
- New configs added in new "prepare" section to define splits
- Some configs in "model" reorganized to a "train" section that is similar to the "prepare" "predict" and "download" sections in that it configures the action more than anything else.
- Added "split" config to train and predict sections to select the split that will be used.

### DIFF
--- a/src/fibad/data_sets/example_cifar_data_set.py
+++ b/src/fibad/data_sets/example_cifar_data_set.py
@@ -11,11 +11,17 @@ class CifarDataSet(CIFAR10):
     FIBAD config with a transformation that works well for example code.
     """
 
-    def __init__(self, config):
+    def __init__(self, config, split: str):
         transform = transforms.Compose(
             [transforms.ToTensor(), transforms.Normalize((0.5, 0.5, 0.5), (0.5, 0.5, 0.5))]
         )
-        super().__init__(root=config["general"]["data_dir"], train=True, download=True, transform=transform)
+
+        if split not in ["train", "test"]:
+            RuntimeError("CIFAR10 dataset only supports 'train' and 'test' splits.")
+
+        train = split == "train"
+
+        super().__init__(root=config["general"]["data_dir"], train=train, download=True, transform=transform)
 
     def shape(self):
         return (3, 32, 32)
--- a/src/fibad/data_sets/hsc_data_set.py
+++ b/src/fibad/data_sets/hsc_data_set.py
@@ -2,7 +2,7 @@
 
 import logging
 import re
-from copy import copy
+from copy import copy, deepcopy
 from pathlib import Path
 from typing import Literal, Optional, Union
 
@@ -26,7 +26,7 @@ class HSCDataSet(Dataset):
 
     """
 
-    def __init__(self, config, split: Union[str, None] = None):
+    def __init__(self, config, split: Union[str, None]):
         # initialize the filesystem references
         self.container = HSCDataSetContainer(config)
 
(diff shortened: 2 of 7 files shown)


src/fibad/data_sets/hsc_data_set.py
```diff
@@ -18,6 +19,235 @@
...
 @fibad_data_set
 class HSCDataSet(Dataset):
+    """Interface object to allow simple access to splits on a corpus of HSC data files
+
+    f/s operations and management are handled in HSCDatSetContainer
+    splits on the dataset and their generation are handled by HSCDataSetSplit
+
+    """
+
+    def __init__(self, config, split: Union[str, None] = None):
```
mtauraso, line 29: The default value of `split=None` here is to get some tests to pass short-term.  The intent is to have the function signature for any `@fibad_data_set` `__init__()` require a split to be provided, or you explicitly pass `None` because it was in a config.

src/fibad/data_sets/hsc_data_set.py
```diff
@@ -18,6 +19,235 @@
...
+        return self.container.shape()
+
+    def __getitem__(self, idx: int) -> torch.Tensor:
+        return self.current_split[idx]
+
+    def __len__(self) -> int:
+        return len(self.current_split)
+
+
+class HSCDataSetSplit(Dataset):
```
mtauraso, line 97: This class had more functionality in the first draft of this, but it now looks a lot like a glorified numpy masked array.

I ended up writing it this way to ensure there's only ever 1 `HSCDatasetContainer` because I expect it to have references to ~10M files at runtime, and I never want a copy made of that object.

If anyone has suggestions to preserve object lifetimes while rewriting this to be shorter with `numpy.ma` I'm interested.

src/fibad/data_sets/example_cifar_data_set.py
drewoldag, line 1: This section looks fine for now. I just searched a bit for CIFAR10 validation sets, and found an inspirational gist here: https://gist.github.com/kevinzakka/d33bf8d6c7f06a9d8c76d97a7879f5cb#file-data_loader-py-L90-L109

Also, I was unaware of the `sampler` parameter that can be passed to DataLoader. I do wonder if it would be helpful, of just get in the way of using HuggingFace datasets.

drewoldag (approved):
Overall I think this looks good for now. The amount of extra code in the hsc_data_set module makes me a little nervous about future users needing to implement similar code for splits, but I also think that is something we can address down the line. 

And I can work with Max to figure out how to adjust the kbmod-ml dataset module to work within this new paradigm.

src/fibad/fibad_default_config.toml
```diff
@@ -92,6 +93,41 @@ batch_size = 32
...
+# It is an error for these values to add to more than 1.0 as ratios or the size of the dataset if expressed
+# as integers.
+#
+# test_size: Size of the test split
+# If `float`, should be between `0.0` and `1.0` and represent the proportion of the dataset to include in the 
+# test split.
+# If `int`, represents the absolute number of test samples.
+# If `false`, the value is set to the complement of the train size.
+# If `train_size` is also `false`, it will be set to `0.25`.
+test_size = 0.6
```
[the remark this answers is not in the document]
mtauraso: I'm going to fix this,  because like you say it is trivial to fix.

I'm also going to adopt your values, because I'm not sure where I got those values from 😄

mtauraso:
> The amount of extra code in the hsc_data_set module makes me a little nervous about future users needing to implement similar code for splits, but I also think that is something we can address down the line.

This also worried me while writing it. I'm going to think about what an external dataset writer can do as we implement huggingface. Hopefully there is an easy way to essentially explain one's data to huggingface and then ask it to perform the split.  I tried to fashion this after their (and scikit.learn's) split code in the hopes that when we get there we're interface compatible. That choice did make the implementation of this PR somewhat more complex than just "add splits"

PR #107, opened 2024-10-29
Adding `dev_mode = false` to `fibad_default_config.toml`

drewoldag:
The primary purpose of this PR is to add a `dev_mode` flag that will allow developers (specifically of external libraries) to move more quickly, and avoid having to define default config values. Instead they will be able to modify and use their user-specific config file during development, with the understanding that, prior to deploying/releasing their code, they will need to introduce a default config file.

Additionally, this PR removes the unused "use_gpu" default config key.


mtauraso (approved):
LGTM.

PR #113, opened 2024-11-01
Initial version of filter_catalog feature.

### LINKED ISSUE #111: Add ability to filter a large HSCDataset down by use of a catalog file
The intent here is to make `HSCDataSetContainer` faster to initialize when large datasets are present, by allowing filtering of the dataset down to a select set of objects.
Feature:
The fibad config would optionally include the name of a catalog file which would contain a list of object IDs that we want to use as a limiter for what files `HSCDataSetContainer` will read in when called by pytorch.

Optionally this catalog file will include filters, filenames, and dimensions. 

If filters and files are provided for each object ID, then `HSCDataSetContainer` will assume all files listed exist and will not list the directory to scan through files during initialization. It will open all files referenced to determine dimensions. If dimensions are also provided, it will not open any files or list any directories during initialization.
 
At present we cannot necessarily rely on a download manifest file with all columns populated due to downloads happening concurrently with fibad development; however, a current-day fibad manifest file would suffice for the performance purpose above (if not the filtering purpose).

mtauraso:
- We can take a fits file as a config
- We filter objects_ids out of a big dataset based on it
- We also skip filesystem checks if there is enough info in the filter catalog.
- Lacks any unit testing
- Added the prepare verb, but right now it just gives you the dataset object when run from a notebook.

Closes #111

### DIFF
--- a/src/fibad/data_sets/hsc_data_set.py
+++ b/src/fibad/data_sets/hsc_data_set.py
@@ -9,6 +9,7 @@ from typing import Literal, Optional, Union
 import numpy as np
 import torch
 from astropy.io import fits
+from astropy.table import Table
 from torch.utils.data import Dataset
 from torchvision.transforms.v2 import CenterCrop, Compose, Lambda
 
@@ -240,6 +241,10 @@ class HSCDataSetSplit(Dataset):
         return self.data[self.indexes[idx]]
 
 
+dim_dict = dict[str, tuple[int, int]]
+files_dict = dict[str, dict[str, str]]
+
+
 class HSCDataSetContainer(Dataset):
     def __init__(self, config):
         # TODO: What will be a reasonable set of tranformations?
@@ -250,12 +255,14 @@ class HSCDataSetContainer(Dataset):
 
         crop_to = config["data_set"]["crop_to"]
         filters = config["data_set"]["filters"]
+        filter_catalog = config["data_set"]["filter_catalog"]
 
         self._init_from_path(
             config["general"]["data_dir"],
             transform=transform,
             cutout_shape=crop_to if crop_to else None,
             filters=filters if filters else None,
+            filter_catalog=Path(filter_catalog) if filter_catalog else None,
         )
 
     def _init_from_path(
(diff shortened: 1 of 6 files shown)


src/fibad/downloadCutout/downloadCutout.py
```diff
@@ -762,6 +764,9 @@ def parse_bool(s: Union[str, bool]) -> bool:
     if isinstance(s, bool):
         return s
 
+    if isinstance(s, np.bool):
```
mtauraso, line 767: This is a bug I ran into which I'm pretty sure numpy fixed in a later version by making `np.bool` an alias of `bool` I ran across it while testing this, but it could impact download/resume (but hasn't so far).

src/fibad/fibad_default_config.toml
```diff
@@ -90,6 +90,10 @@ crop_to = false
 #filters = ["HSC-G", "HSC-R", "HSC-I", "HSC-Z", "HSC-Y"]
 filters = false
 
+# A fits file which specifies object IDs to filter a large dataset in [general].data_dir down
+# Implementation is dataset class dependent. Default is false meaning now filtering.
+filter_catalog = false
```
mtauraso, line 95: I hate the name of this config but couldn't think of a better one.

src/fibad/data_sets/hsc_data_set.py
```diff
@@ -335,11 +362,17 @@ def _scan_file_names(self, filters: Optional[list[str]] = None) -> dict[str, dic
 
         files = {}
         # Go scan the path for object ID's so we have a list.
-        for filepath in Path(self.path).glob("[0-9]*.fits"):
+        for filepath in Path(self.path).iterdir():
```
mtauraso, line 365: The switch to iterdir is intentional. I believe it is doing less work than `glob` and that we might have some perf gain when the directory has 10M files in it.

Yes this is the premature optimization you were warned about 😈

src/fibad/data_sets/hsc_data_set.py
```diff
@@ -359,7 +392,57 @@ def _scan_file_names(self, filters: Optional[list[str]] = None) -> dict[str, dic
 
         return files
 
-    def _scan_file_dimensions(self) -> dict[str, tuple[int, int]]:
+    def _read_filter_catalog(
+        self, filter_catalog_path: Optional[Path]
+    ) -> Optional[Union[list[str], files_dict, tuple[files_dict, dim_dict]]]:
```
mtauraso, line 397: This could do with a docstring.

aritraghsh09 (approved):
I tested this with a small HSC catalog, and it works as expected. The dataset being returned by prepare also works well. 

While using and testing this out, I came across some changes I want to suggest (with respect to splitting and other things); but I will put them into another branch/PR. 

Let's get this merged first.

PR #115, opened 2024-11-08
Rebuild a fits manifest from an HSC data directory.

mtauraso:
- Added a new verb rebuild_manifest
- When run with the HSC dataset class this verb will: 0) Scan the data directory and ingest HSC cutout files 1) Read in the original catalog file configured for download for metadata 2) Write out rebuilt_manifest.fits in the data directory
- Fixed up config resolution so that fibad_config.toml in the cwd works again for CLI invocations.
- File scanning has been parallelized with multiprocessing to reduce time from ~30d -> 6hr on hyak w 24M files
- Number of processes to launch dynamically chosen based on system limits, with intent to achieve max io bandwidth.

### DIFF
--- a/src/fibad/data_sets/hsc_data_set.py
+++ b/src/fibad/data_sets/hsc_data_set.py
@@ -644,7 +644,7 @@ class HSCDataSetContainer(Dataset):
         column_names = Downloader.MANIFEST_COLUMN_NAMES
         columns = {column_name: [] for column_name in column_names}
 
-        # These we vary every object and must be implemented below
+        # These will vary every object and must be implemented below
         dynamic_column_names = ["object_id", "filter", "dim", "tract", "ra", "dec", "filename"]
         # These are pulled from config ("sw", "sh", "rerun", "type", "image", "mask", and "variance")
         static_column_names = [name for name in column_names if name not in dynamic_column_names]
--- a/src/fibad/data_sets/hsc_data_set.py
+++ b/src/fibad/data_sets/hsc_data_set.py
@@ -33,32 +33,6 @@ from .data_set_registry import fibad_data_set
 logger = logging.getLogger(__name__)
 
 
-def num_cpus() -> int:
-    """
-    Checks both os.sched_getaffinity and multiprocessing.cpu_count()
-    to find the number of CPUs we have runtime access to.
-
-    Note borrowed from: https://github.com/pylint-dev/pylint/pull/6098/files
-    But we don't check for cgroup-based cpu time limits.
-
-    Returns
-    -------
-    int
-        Number of CPUs
-    """
-    cpu_count = None
-    sched_getaffinity = getattr(os, "sched_getaffinity", None)
-
-    if sched_getaffinity:
-        cpu_count = len(sched_getaffinity(0))
-    elif multiprocessing:
-        cpu_count = multiprocessing.cpu_count()
-    else:
-        cpu_count = 1
-
-    return cpu_count
-
-
 @fibad_data_set
 class HSCDataSet(Dataset):
     _called_from_test = False
--- a/src/fibad/data_sets/hsc_data_set.py
+++ b/src/fibad/data_sets/hsc_data_set.py
@@ -13,6 +13,17 @@ from astropy.table import Table
 from torch.utils.data import Dataset
 from torchvision.transforms.v2 import CenterCrop, Compose, Lambda
 
+from fibad.download import Downloader
+from fibad.downloadCutout.downloadCutout import (
+    parse_bool,
+    parse_degree,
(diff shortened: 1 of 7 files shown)


src/fibad/rebuild_manifest.py
```diff
@@ -0,0 +1,22 @@
...
+    config : dict
+        The parsed config file as a nested
+        dict
+    """
+
+    _, data_set = setup_model_and_dataset(config, split=config["train"]["split"])
+
+    data_set.rebuild_manifest(config)
+
+    logger.info("Finished Prepare")
```
drewoldag, line 22:
```suggestion
    logger.info("Finished rebuilding manifest")
```

src/fibad/data_sets/hsc_data_set.py
```diff
@@ -553,6 +567,88 @@ def _check_file_dimensions(self) -> tuple[int, int]:
...
+
+        del location_table
+
+        logger.info("Assembling Manifest...")
+
+        # These are the column names expected in a manifest file by the downloader
+        column_names = Downloader.MANIFEST_COLUMN_NAMES
+        columns = {column_name: [] for column_name in column_names}
+
+        # These we vary every object and must be implemented below
```
drewoldag, line 596:
```suggestion
        # These will vary every object and must be implemented below
```

src/fibad/data_sets/hsc_data_set.py
```diff
@@ -553,6 +567,88 @@ def _check_file_dimensions(self) -> tuple[int, int]:
...
+            "rerun": parse_rerun(config["download"]["rerun"]),
+            "type": parse_type(config["download"]["type"]),
+            "image": parse_bool(config["download"]["image"]),
+            "mask": parse_bool(config["download"]["mask"]),
+            "variance": parse_bool(config["download"]["variance"]),
+        }
+
+        for object_id, filter, filename, dim in self._all_files_full():
+            for static_col in static_column_names:
+                columns[static_col].append(static_values[static_col])
```
drewoldag, line 618: At this point in the manifest building process, do we know the number of files that will be yielded by `self._all_files_full()`? If so, can we build out the static_col for loop in one go outside of the outer for loop?

```
for static_col in static_column_names:
    columns[static_col] = [static_values[static_col]] * num_files
```

src/fibad/data_sets/hsc_data_set.py
```diff
@@ -553,6 +567,88 @@ def _check_file_dimensions(self) -> tuple[int, int]:
...
+
+            for dynamic_col in dynamic_column_names:
+                if dynamic_col == "object_id":
+                    columns[dynamic_col].append(int(object_id))
```
drewoldag, line 622: Roughly the same question as above - if we know before hand the total number of files, we could preallocate the arrays. I understand that python is suppose to be efficient with `.append`, but the over allocation of memory will hit in clearly defined steps for each of these arrays and more than double the memory footprint each time. 

That might be just fine, but if we could do the same thing here but just updating a specific index instead of `.append()`ing, we might save ourselves a headache in the future. 

Of course, if we don't know the number of files before hand, then just ignore this :). Perhaps prep for loop that calls the generator just to increment a counter would be worth while?

drewoldag (commented):
I left a few comments about here, I don't believe any of them are blocking, but it would be good to know the answer since they are efficiency related.

I'm guessing that rebuild_manifest is not a process that will happen often, so maybe the efficiency isn't as much of a concern, but faster is always better right?

src/fibad/data_sets/hsc_data_set.py
mtauraso: Yeah we could reverse the loops here, and that might be faster.

I'm mostly worried about memory footprint here rather than speed. I wrote it in this order in case I could integrate fitsio and stream individual objects into the fits file, rather than astropy's model.

src/fibad/data_sets/hsc_data_set.py
mtauraso: Yeah there's probably a speedup here, and we do know the size of these objects at the start of the loop.

src/fibad/data_sets/hsc_data_set.py
mtauraso: Testing on Hyak the entire flow takes ~6hr, with this particular loop taking ~3 minutes on a 8M object/24M file dataset.

The critical section for `rebuild_manifest` is the i/o now in `_fits_file_dims` and is where nearly all of the time is spent. This is also the critical section for dataset loading in general.

I'm going to pass on this optimization for now, and make it so *by default* we skip the f/s scan which is going to save time in the typical case where we have a manifest.

PR #132, opened 2024-12-07
Allow definition of Loss and Optimizer in config file.

drewoldag:
This is an initial implementation that allows defining PyTorch Loss (Criterion) and Optimizers from within the config file. 

Two new tables are introduced, `[criterion]` and `[optimizer]` and we currently use the relatively familiar libpath approach to avoid having to create a registry of pytorch criterion and optimizer functions which would need to be loaded at run time (kind of heavy and slow) 

The keys in the tables need to share the same names as the parameters that would be passed to the loss or optimizer functions. (we just use  `**config` to pass the params in). 

Within the `__init__` function of either the built in models or in an external class the user will still need to call: 
```
```
(I would like to fix that in a followup PR - Issue #134)

Additionally, if the user has a special `def _optimizer()` or `def _criterion()` function that is defined in the model class, we will not over write that.

One final followup would be to allow a Learning Rate scheduler ([described here](https://pytorch.org/docs/stable/optim.html#how-to-adjust-learning-rate)) to be defined in the config as well. That won't happen in this PR though.

### DIFF
--- a/src/fibad/fibad_default_config.toml
+++ b/src/fibad/fibad_default_config.toml
@@ -61,6 +61,18 @@ name = "ExampleAutoencoder"
 base_channel_size = 32
 latent_dim = 64
 
+[criterion]
+# The name of the built-in criterion to use or the libpath to an external criterion
+name = "torch.nn.CrossEntropyLoss"
+
+[optimizer]
+# The name of the built-in optimizer to use or the libpath to an external optimizer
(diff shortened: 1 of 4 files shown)


aritraghsh09 (approved):
This implementation looks perfectly in line with what I had requested! 

No comments. Looks ready to merge to me!

PR #136, opened 2024-12-08
Adding a simple HSC Autoencoder as an example 

aritraghsh09:
## Change Description

- [X] My PR includes a link to the issue that I am addressing

This PR does the following:-
- Adds a very simple autoencoder that works with the HSC data and has a training model so generic that users should be able to use it widely for a large number of DL models 
- Adds instructions to the training demo on how to run `tensorboard` when running it remotely on a server

This is currently marked as a draft because of a an issue in the change I implemented in `fibad_default_config.toml` (see below)

### DIFF
--- a/docs/notebooks/train_model.ipynb
+++ b/docs/notebooks/train_model.ipynb
@@ -87,8 +87,20 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%tensorboard --logdir ./results"
+    "%tensorboard --logdir ./results\n",
+    "\n",
+    "# if running on a remote server, and tunnelling a connection,\n",
+    "# pass the --bind-all flag\n",
+    "# %tensorboard --logdir ./results --bind_all\n",
+    "# and then forward the selected port to your local machine"
    ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
   }
  ],
  "metadata": {
--- a/src/fibad/fibad_default_config.toml
+++ b/src/fibad/fibad_default_config.toml
@@ -61,6 +61,18 @@ name = "ExampleAutoencoder"
 base_channel_size = 32
 latent_dim = 64
 
+[criterion]
+# The name of the built-in criterion to use or the libpath to an external criterion
+name = "torch.nn.CrossEntropyLoss"
+
+[optimizer]
+# The name of the built-in optimizer to use or the libpath to an external optimizer
+name = "torch.optim.SGD"
+
+# Default PyTorch optimizer parameters. The keys match the names of the parameters
+lr = 0.01
+momentum = 0.9
+
 [train]
 weights_filepath = "example_model.pth"
 epochs = 10
--- a/src/fibad/fibad_default_config.toml
+++ b/src/fibad/fibad_default_config.toml
@@ -71,7 +71,7 @@ name = "torch.optim.SGD"
 
 # Default PyTorch optimizer parameters. The keys match the names of the parameters
 lr = 0.01
-momentum = 0.9
+#momentum = 0.9
 
 [train]
 weights_filepath = "example_model.pth"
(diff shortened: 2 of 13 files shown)


src/fibad/fibad_default_config.toml
```diff
@@ -61,6 +61,18 @@ name = "ExampleAutoencoder"
...
+# The name of the built-in criterion to use or the libpath to an external criterion
+name = "torch.nn.CrossEntropyLoss"
+
+[optimizer]
+# The name of the built-in optimizer to use or the libpath to an external optimizer
+name = "torch.optim.SGD"
+
+# Default PyTorch optimizer parameters. The keys match the names of the parameters
+lr = 0.01
+#momentum = 0.9
```
aritraghsh09, line 74: @drewoldag  I am using `optim.Adam` as my optimizer, which doesn't take `momentum` as an argument. So, the only way to make it work in the current implementation was to set `dev_mode = true` in my local config file and to comment out momentum (above) in the default configuration file. 

I tried setting `momentum=false` in my local config, but that didn't take care of the above. What do you think is a better solution here? As all optimizers might/mightn't take different arguments. 

This also raises the question of the default behavior of our configuration system to assume defaults via the default config file and then come up with a combination using the default and user-specified configs? For me, personally, this has been confusing sometimes. Can't we ask users to ask users to copy the default config file and modify it as needed? And then only that file is used for reading the configs?

drewoldag: These are both good questions, and it can drive a good conversation during the sync meeting today. 

For the first question, regarding different arguments for different optimizers and loss functions, one thing we could potentially do is introspect the function that is about to be called (in this case `optim.Adam`) and get the list of parameters that it can accept, then only pass the config values that match one of those. That might be the most transparent to the user in general. 

Another approach would be to create a subtable in the config file for the optimizer of choice, but this feels unnecessarily burdensome. Something like: 
```
[optimizer]
name = 'torch.optim.Adam'
...

[optimizer.Adam]
lr = 0.01
< other Adam params >

[optimzer.Other]
< params that are only for "Other" >
```

Regarding the second question about just making a copy of the config and using that. In the short term we could implement that, but we've built the config system to be able to support external model/dataset libraries as well as those that are built in. 

For a user working with an external library, we would want the defaults of that library to override the fibad defaults, and then the user specific values to override anything else. I _think_ that for a new user, that the first step of merging fibad and external defaults becomes a bit challenging (or at least annoying)

mtauraso: For the suggested workaround of momentum=False you can make the following change:

https://github.com/lincc-frameworks/fibad/commit/47c42213470be3e14c9b7e569d41bde94b9337a8

If you `git fetch` you should be able to `git cherry-pick 47c42213470be3e14c9b7e569d41bde94b9337a8` in your PR branch to pull in the change.

For this PR I'd advocate cherry-picking this change, and putting the rest of the config refactor discussed above and in the meeting into an issue to be completed separately from this PR.

docs/notebooks/train_model.ipynb
```diff
@@ -87,8 +87,20 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%tensorboard --logdir ./results"
+    "%tensorboard --logdir ./results\n",
+    "\n",
+    "# if running on a remote server, and tunnelling a connection,\n",
```
mtauraso, line 92: It seems like there is a tensorboard integration feature buried in this comment :)

We should probably flesh it out a bit.

src/fibad/models/model_registry.py
```diff
@@ -23,6 +23,20 @@ def _torch_load(self: nn.Module, load_path: Path):
...
+    return criterion_function_cls(**arguments)
+
+
+def _torch_optimizer(self: nn.Module):
+    optimizer_cls = get_or_load_class(self.config["optimizer"])
+    arguments = dict(self.config["optimizer"])
+    del arguments["name"]
+    return optimizer_cls(self.parameters(), **arguments)
+
+
```
mtauraso, line 39: You may want to rebase this on main before merging, this looks like some appearance of one of Drew's earlier commits from PR#132 GIt should sort this out in the merge and give you a clean diff.

mtauraso (approved):
Some minor comments but nothing big. This should get merged even if there is disagreement on my commentary.

PR #139, opened 2024-12-13
Breaking `setup_model_and_dataset` into two functions.

drewoldag:
Breaking these apart avoids unnecessary creation of model instances in prepare and rebuild_manifest. And this opens the door for a potentially more intuitive way to create datasets for training that include (train, [validate, test]) subsets. 

This PR only implements the separation of the `setup_model_and_dataset` function and percolates that change down to the various fibad verb implementations.


mtauraso (approved):
LGTM.

PR #141, opened 2024-12-13
Renaming verb `predict` to `infer`

drewoldag:
Just going through and renaming all "predict"s to "infer"s. 

The only major update is the rename of the predict.py file.


mtauraso (approved):
Populist hero, achievement unlocked

PR #143, opened 2024-12-17
WIP - Vector DB investigation

drewoldag:
Lots of throw away code in this particular PR to investigate how to integrate and use a vector db for image similarity searching. 

I wanted to get a checkpoint before I headed home for the day. 

If you want to play with the same small HSC dataset that I'm using, I've stashed a zip file on Zenodo. Let me know if you would to get access to it.


drewoldag:
This PR counts as experimentation at this point, closing this PR.

PR #148, opened 2025-01-09
Retaining comments in persisted config files

drewoldag:
This PR introduces tomlkit to the codebase and uses that as a way to retain comments from the configuration .toml files that are read in, merged, and subsequently written back out as part of training or inference. 

While it was a bit of a convoluted journey, the ultimate discovery was that tomlkit could mostly be used out of the box without too much trouble and provide the desired behavior. I was able to verify that config files are merged as expected with one caveat - in a user specific config, if comments are added above a key that is defined in the default config, those comments will not be copied over. However, inline comments will.

An example user specific config .toml file
```
[general]

# this comment will not be persisted in the final config
log_level = 'warn'  # This comment WILL be persisted

[special_table]
# this comment WILL be persisted
new_key = 1 # This comment WILL be persisted
```

My take is that this is good enough, I would hope that changing the default values of basic fibad config values would only require a small comment to explain.


src/fibad/config_utils.py
```diff
@@ -65,6 +66,14 @@ def clear(self):
         raise RuntimeError("Removing keys or sections from a ConfigDict using clear() is not supported")
 
 
+TOMLDocument.__missing__ = ConfigDict.__missing__  # type: ignore[attr-defined]
```
mtauraso, line 69: Would be good to have a short comment here explaining the intent of these lines. In six months we will have forgotten how this works probably.

mtauraso:
Its good that we're cleanly separating here the pieces of functionality that require us to extend tomlkit vs those that don't. I want to make sure we remember what other functionality we had in mind, even if that functionality is not committed for the MTR.

I don't think this design is going to generalize to the full notebook config help use case, where we envisioned some sort of help interface on a fibad config returning a docstring pulled from comments on default/user configs. If I'm understanding correctly, there are two hurdles to fully accomplish that use case:
1) We need to have the help interface traverse the comments in TomlDocument (much like we whiteboarded) in order to display the appropriate comments.
2) In order to do the user-config-comments as help (or external library defaults as help) ... I think we have to revisit how comments are stored and how we iterate over these dictionaries entirely? IIUC comment information from non-default dictionaries is lost whenever we merge because merge only operates on k/v pairs, not comments.

On balance I think this should merge now because of the mid-term review and the visibility of missing comments in result directories; however, I'd want the additional work for the notebook use case to be captured in an open issue.

mtauraso (approved):

src/fibad/config_utils.py
drewoldag: done

drewoldag:
Created issue https://github.com/lincc-frameworks/fibad/issues/165 to track the last comment

PR #153, opened 2025-01-10
Type checking

mtauraso:
* Added mypy type checking to the codebase
* Fixed all type checking errors (clean `mypy .`)
* Fixed a couple of dataset split issues while testing download and train paths on new code and small datasets.
* Updated python project template to latest version, adding CI and pre-commit type checking

### DIFF
--- a/src/fibad/config_utils.py
+++ b/src/fibad/config_utils.py
@@ -2,7 +2,7 @@ import datetime
 import importlib
 import logging
 from pathlib import Path
-from typing import Union
+from typing import Optional, Union
 
 import toml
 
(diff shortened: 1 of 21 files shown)


src/fibad/config_utils.py
```diff
@@ -197,7 +199,7 @@ def merge_configs(default_config: dict, overriding_config: dict) -> dict:
         return final_config
 
     @staticmethod
-    def _validate_runtime_config(runtime_config: ConfigDict, default_config: ConfigDict):
+    def _validate_runtime_config(runtime_config: dict, default_config: dict):
```
drewoldag, line 202: Oh, interesting, I'm surprised that mypy was upset about these types.

PR #155, opened 2025-01-10
User Controlled Caching

aritraghsh09:
I added a config variable that controls whether images are cached or not during loading. 

This is meant to be a (temporary) fix till we invest time in implementing more elegant caching routines / distributed data loading etc. 

Feel free to push code directly if you spot something inefficient.

### DIFF
--- a/src/fibad/data_sets/hsc_data_set.py
+++ b/src/fibad/data_sets/hsc_data_set.py
@@ -331,6 +331,8 @@ class HSCDataSetContainer(Dataset):
         crop_to = config["data_set"]["crop_to"]
         filters = config["data_set"]["filters"]
 
+        self.use_cache = config["data_set"]["use_cache"]
+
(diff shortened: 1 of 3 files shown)


src/fibad/data_sets/hsc_data_set.py
```diff
@@ -1021,9 +1023,10 @@ def _object_id_to_tensor(self, object_id: str) -> torch.Tensor:
         torch.Tensor
             A tensor with dimension (self.num_filters, self.cutout_shape[0], self.cutout_shape[1])
         """
-        data_torch = self.tensors.get(object_id, None)
-        if data_torch is not None:
-            return data_torch
+        if self.use_cache is True:
```
mtauraso, line 1026: Nit: `if self.use_cache` is canonical. No need to change.

mtauraso (approved):

Issue #138, opened 2024-12-12
Remove the need to define a split when running `predict`


drewoldag:
The current code expects that a split will be defined in the `[predict]` table in the config file. The user can then set the specified split to be 1.0 (100%) of the data in the data directory.

But this feels like an odd requirement for the user to update the `[data_set]` split definitions in order to run inference. 

I would advocate for an approach that doesn't require the user to specify a split in the `[predict]` table, and just runs all the data found through the trained model.

aritraghsh09:
This will be the expected default user behavior in unsupervised learning scenarios. 

In supervised scenarios, users might want to run predict only on the "test" set (i.e., data - train - validation)

I don't think there is a "right" answer here.

drewoldag:
Proposing a solution - what is the default value for the split value in `[predict]` is `false`, which would indicate "no split, just use all the data in the data directory". And if the value is not "false" then we would expect it to be one of ["train", "validate", "test"]. 

I think that would suppose both use cases relatively easily:
1) I just want to use an existing trained model with a bunch of data - no mods required to the config file.
2) I want to test my trained model on a subset of my input data - the user would expect to have to define the subset in the config.

mtauraso:
For drew's solution, we might call the config `use_split`?

I'm going to work up a PR on this and see what folks think of it.

PR #156, opened 2025-01-10
Default to using the whole dataset for inference unless overriden

mtauraso:
See discussion on Issue #138 This implements @drewoldag 's suggestion.


drewoldag (approved):
SHIP IT

PR #157, opened 2025-01-10
Adding credentials file support

mtauraso:
Simple support for a credentials.ini file with documentation


drewoldag (approved):
This looks like the right path forward to me. Thanks for putting it together.

PR #159, opened 2025-01-14
Prep for similarity search: adding chromadb to infer

mtauraso:
* Infer builds a chromadb in the results directory
* Moved example_hsc_dataset.py -> hsc_dataset.py for clarity that it is not an example
* Bugfix to HSCDataset handling of [data_set]filters= config. When given a data dir with more filters than specified in [data_set]filters, it now cuts down the required filters of the dataset to only those specified in config. Previously in this same situation it would prune all objects in the dataset
* Infer verb defaults to choosing the model from the latest training run if no model weight path is specified in config.
* Results directories now append a random string of characters to the end to avoid folder name collisions happening when two fibad processes starting in the same second share a filesystem
* Updated names of configs in tests to match those used currently.
* Fixed parsing of split parameter in HSCDataSet to accept a [infer]split=false
* All chromadb functionality can be disabled with [infer] chromadb=false
* Numpy files from inference are now saved per batch as numpy rec arrays with "id" and "tensor" keys corresponding to the dataset id numbers and the output tensors from the model.


mtauraso:
To give an idea of the impact of this change on a small results directory, this is what you get when running on @drewoldag 's  hsc_8asec_1000 dataset with a batch size of 100:

```
(fibad) mtauraso@Michaels-MacBook-Pro 20250116-204233-infer-J2la % ls -lh
total 1920
drwxr-xr-x  6 mtauraso  staff   192B Jan 16 20:42 6cd9a1ef-3a43-4f1c-b9eb-fc8a669b4ead
-rw-r--r--  1 mtauraso  staff    26K Jan 16 20:42 batch_0.npy
-rw-r--r--  1 mtauraso  staff    26K Jan 16 20:42 batch_1.npy
-rw-r--r--  1 mtauraso  staff    26K Jan 16 20:42 batch_2.npy
-rw-r--r--  1 mtauraso  staff    26K Jan 16 20:42 batch_3.npy
-rw-r--r--  1 mtauraso  staff    26K Jan 16 20:42 batch_4.npy
-rw-r--r--  1 mtauraso  staff    26K Jan 16 20:42 batch_5.npy
-rw-r--r--  1 mtauraso  staff    26K Jan 16 20:42 batch_6.npy
-rw-r--r--  1 mtauraso  staff    26K Jan 16 20:42 batch_7.npy
-rw-r--r--  1 mtauraso  staff    26K Jan 16 20:42 batch_8.npy
-rw-r--r--  1 mtauraso  staff    24K Jan 16 20:42 batch_9.npy
-rw-r--r--  1 mtauraso  staff   648K Jan 16 20:42 chroma.sqlite3
-rw-r--r--  1 mtauraso  staff   1.1K Jan 16 20:42 runtime_config.toml
(fibad) mtauraso@Michaels-MacBook-Pro 20250116-204233-infer-J2la % du -sh 6cd9a1ef-3a43-4f1c-b9eb-fc8a669b4ead
396K	6cd9a1ef-3a43-4f1c-b9eb-fc8a669b4ead
```

Adding up the batch numpy files is around 1/4 of the chromadb sqlite database and index:
```
(fibad) mtauraso@Michaels-MacBook-Pro 20250116-204233-infer-J2la % du -ch *.npy | tail -n 1
280K	total
(fibad) mtauraso@Michaels-MacBook-Pro 20250116-204233-infer-J2la % python3
Python 3.10.14 | packaged by conda-forge | (main, Mar 20 2024, 12:51:49) [Clang 16.0.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> (648+396)/240
4.35
```

For small datasets we're making the results directory ~5x larger with chromadb. I'd be interested in what the similar multiple is for larger datasets @aritraghsh09

aritraghsh09:
This looks good to me; but @drewoldag should do the official review. I will run the testing for chroma-db after this is merged into main.

src/fibad/infer.py
```diff
@@ -34,26 +41,65 @@ def run(config: ConfigDict):
...
+    # Create a chromadb in results dir
+    if config["infer"]["chromadb"]:
+        chromadb_client = chromadb.PersistentClient(path=str(results_dir))
+        collection = chromadb_client.create_collection(
+            name="fibad",
+            metadata={
+                # These are all chromdb defaults. We may want to configure them
+                "hsnw:space": "l2",
+                "hsnw:construction_ef": 100,
+                "hsnw:search_ef": 100,
```
drewoldag, line 53: Not as part of this PR, but it would be good to run some benchmarks on these parameters. It's sensible that we could open these up to uses via the config file, and in doing so we could provide some guidance about the trade off of performance and accuracy for different database sizes.

src/fibad/infer.py
```diff
@@ -69,22 +115,28 @@ def load_model_weights(config: ConfigDict, model):
         The model class to load weights into
 
     """
-    weights_file = config["infer"]["model_weights_file"]
+    weights_file: Optional[Union[str, Path]] = (
+        config["infer"]["model_weights_file"] if config["infer"]["model_weights_file"] else None
+    )
+
+    if weights_file is None:
+        recent_results_path = find_most_recent_results_dir(config, "train")
```
drewoldag, line 123: Maybe logger.info here to say that we're looking in the most recent `train` directory for a model weights file, to try to make the action a little more transparent.

drewoldag (approved):
Overall this looks fine. Only minor comment is a request to add some logging to let someone know that we're looking for a model weights file automatically, just to make it seem less mysterious.

src/fibad/infer.py
mtauraso: Totally agree.

PR #162, opened 2025-01-15
Cleaning up default config formatting, values and comments

drewoldag:
Doing my best to apply some uniform formatting to the config file to make it a little less overwhelming. 

The linting rules that I tried to adhere to:
1. No table-level comments
2. No comments outside of a table
3. Two lines between tables
4. One line between a key and the following comment
5. 2 lines of comments "maximum" per key - I failed on this for the test/validation/train splits.
6. No free floating blocks of comments - Again I failed when it came to the test/validation/train splits, there's just a lot there.
7. Reasonable maximum line length for comments
8. Include optional values when possible (i.e. allowed log levels)
9. Tried to have one comment per key, but used best judgement there. (i.e. username, password only got one comment)

### DIFF
--- a/src/fibad/fibad_default_config.toml
+++ b/src/fibad/fibad_default_config.toml
@@ -1,78 +1,97 @@
 [general]
-# Whether to run in development mode. When `true`, this will slightly modify the
-# behavior of Fibad to make development faster and easier. For example, it will
-# not check for the existance of default config values in external libraries.
+# Set to `true` during development to skip checking for default config values
+# in external libraries. Use `false` otherwise.
 dev_mode = false
 
-# Destination of log messages
-# 'stderr' and 'stdout' specify the console.
+# Destination of log messages. Options: 'stderr', 'stdout' specify the console,
+# "path/to/fibad.log" specifies a file.
 log_destination = "stderr"
-# A path name specifies a file e.g.
-# log_destination = "fibad_log.txt"
 
-# Lowest log level to emit.
-# As you go down the list, fibad will become more verbose in the log.
-#
-# log_level = "critical" # Only emit the most severe of errors
-# log_level = "error"    # Emit all errors
-# log_level = "warning"  # Emit warnings and all errors
-log_level = "info"     # Emit informational messages, warnings and all errors
-# log_level = "debug"    # Very verbose, emit all log messages.
+# Lowest log level to emit. Options: "critical", "error", "warning", "info", "debug".
+log_level = "info"
 
+# Directory where data is stored.
 data_dir = "./data"
-results_dir = "./results" # Results get named <verb>-<timestamp> under this directory
+
+#Top level directory for writing results.
+results_dir = "./results"
+
 
 [download]
+# Cut out width in arcseconds.
 sw = "22asec"
+
+# Cut out height in arcseconds.
 sh = "22asec"
+
+# The filters to download.
 filter = ["HSC-G", "HSC-R", "HSC-I", "HSC-Z", "HSC-Y"]
+
+# The type of data to download.
 type = "coadd"
+
+# The data release to download from.
 rerun = "pdr3_wide"
 
-# Credentials for the downloader. Either provide username/password in a fibad config file
-# or with a credentials.ini file. The format for credentials.ini is as follows:
-#
+# Path to credentials.ini file for the downloader. File contents should be:
 # username = "<your username>"
 # password = "<your password>"
-#
-# It is preferred to use a credentials.ini file to avoid sensitive user credentials being
-# committed to source control systems like git.
-username = false
-password = false
-# Location of the credentials.ini file. Defaults to credentials.ini in the current directory
-# but can be configured to a common location for batch processing.
 credentials_file = "./credentials.ini"
 
-num_sources = -1 # Values below 1 here indicate all sources in the catalog will be downloaded
+# Alternative method to pass credentials to the downloader. Prefer to use a
+#credentials.ini file to avoid exposing credentials with source control.
+username = false
+password = false
+
+# The number of sources to download from the catalog. Default is -1, which
+# downloads all sources in the catalog.
+num_sources = -1
+
+# The number of sources to skip in the catalog. Default is 0.
 offset = 0
+
+# The number of concurrent connections to use when downloading data.
 concurrent_connections = 4
+
+# The number of seconds between printing download statistics.
 stats_print_interval = 60
+
+# The path to the catalog file that defines which cutouts to download.
 fits_file = "./catalog.fits"
 
-# These control the downloader's HTTP requests and retries
-# `retry_wait` How long to wait before retrying a failed HTTP request in seconds. Default 30s
+# The number of seconds to wait before retrying a failed HTTP request in seconds.
 retry_wait = 30
-# `retries` How many times to retry a failed HTTP request before moving on to the next one. Default 3 times
+
+# How many times to retry a failed HTTP request before moving on to the next one.
 retries = 3
-# `timepout` How long should we wait to get a full HTTP response from the server. Default 3600s (1hr)
+
+# Number of seconds to wait for a full HTTP response from the server.
 timeout = 3600
-# `chunksize` How many sky location rectangles should we request in a single request. Default is 990
+
+# The number of sky location rectangles should we request in a single request.
 chunk_size = 990
 
-# Whether to retrieve the image layer
+# Request the image layer from the cutout service
 image = true
-# Whether to retrieve the variance layer
+
+# Request the variance layer from the cutout service
 variance = false
-# Whether to retrieve the mask layer
+
+# Request the mask layer from the cutout service
 mask = false
 
+
 [model]
-# The name of the built-in model to use or the libpath to an external model
-# e.g. "user_package.submodule.ExternalModel" or "ExampleAutoencoder"
+# The name of the model to use. Option are a builtin model class name or import path
+# to an external model. e.g. "ExampleAutoencoder", "user_pkg.model.ExternalModel"
 name = "ExampleAutoencoder"
 
+# The number of output channels from the first layer.
 base_channel_size = 32
+
+# The length of the latent space vector.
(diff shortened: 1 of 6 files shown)


src/fibad/fibad_default_config.toml
```diff
@@ -1,172 +1,189 @@
...
-# but can be configured to a common location for batch processing.
-credentials_file = "./credentials.ini"
 
-num_sources = -1 # Values below 1 here indicate all sources in the catalog will be downloaded
+# The number of sources to download from the catalog. Default is -1, which
+# downloads all sources in the catalog.
+num_sources = -1
+
+# The number of sources to skip in the catalog. Default is 0.
 offset = 0
```
mtauraso, line 51: This shouldn't be a config anymore

src/fibad/fibad_default_config.toml
```diff
@@ -1,172 +1,189 @@
...
 [optimizer]
-# The name of the built-in optimizer to use or the libpath to an external optimizer
+# The name of the built-in optimizer to use or the import path to an external optimizer
 name = "torch.optim.SGD"
 
-# Default PyTorch optimizer parameters. The keys match the names of the parameters
+# learning rate parameter for default optimizer. The keys match the names of the parameters
 lr = 0.01
+
+# momentum parameter for default optimizer. The keys match the names of the parameters
```
mtauraso, line 107: "The keys match the names of the parameters" made sense before. It doesn't now.

I think we want to include the prior meaning.

src/fibad/fibad_default_config.toml
```diff
@@ -1,172 +1,189 @@
...
 crop_to = false
 
-# Limit data loader to only particular filters when there are more in the data set.
-#
-# When not provided by the user, the number of filters will be automatically gleaned from the data set.
-# Defaults behavior is produced by the false value.
-#
-#filters = ["HSC-G", "HSC-R", "HSC-I", "HSC-Z", "HSC-Y"]
+# Specific to `hsc_data_set`. Limit to only particular filters from the data set.
+# When `false`, uses all filters found in the data set.
```
mtauraso, line 136: I think the commented out example was helpful here. I'd advocate for its return.

src/fibad/fibad_default_config.toml
```diff
@@ -1,172 +1,189 @@
...
+# [general].data_dir. Implementation is data_set class dependent. Use `false` for no filtering.
 filter_catalog = false
 
-# How to split the data between training and eval sets.
-# The semantics are borrowed from scikit-learn's train-test-split, and HF Dataset's train-test-split function
-# It is an error for these values to add to more than 1.0 as ratios or the size of the dataset if expressed
-# as integers.
+# For train_size, validation_size, and test_size, the following options are available:
+# A `float` between `0.0` and `1.0` is the proportion of the dataset to include in the split.
+# If `int`, represents the absolute number of samples in the particular split.
```
mtauraso, line 145: This should be "An `int`..." to preserve parallel construction with the line above about 'float' typed values.

src/fibad/fibad_default_config.toml
```diff
@@ -1,172 +1,189 @@
...
+# It is an error for these values to add to more than 1.0 as ratios or the size
+# of the dataset if expressed as integers.
```
mtauraso, line 147: My preference would be to rewrite this preamble as follows:

```
# train_size, validation_size, and test_size use these conventions:
# * A `float` between `0.0` and `1.0` is the proportion of the dataset to include in the split.
# * An `int` representing the absolute number of samples in the particular split.
# * It is an error for these values to add to more than 1.0 as ratios or the size
#   of the dataset if expressed as integers.
```

src/fibad/fibad_default_config.toml
```diff
@@ -1,172 +1,189 @@
...
-# Number to seed with for generating a random split. False means the data will be seeded from
-# a system source at runtime.
+# Number to seed with for generating a random split. Use `false` to seed from a
+# system source at runtime.
 seed = false
 
-#Controls whether images are cached during data loading. For training, this reduces runtimes
-#after the first epoch.
+# If `true`, cache data in memory during training. For training, this reduces runtimes
+#after the first epoch. Set to `false` to disable caching or when running inference.
```
mtauraso, line 169: This should probably have a note that most users want this off, because it only speeds up inference if you're doing something like drop-out.

src/fibad/fibad_default_config.toml
drewoldag: I have an other PR implementing the subtables for criterion and optimizer that should address this. Going to leave it as is for now.

src/fibad/fibad_default_config.toml
```diff
@@ -1,172 +1,189 @@
...
 use_cache = true
 
+
 [data_loader]
-# Default PyTorch DataLoader parameters
-batch_size =32
+# The number of data points to load at once.
+batch_size = 32
```
mtauraso, line 175: I'm of the opinion that this should be (128-1024) to avoid the trap of low gpu performance out of the box.

src/fibad/fibad_default_config.toml
```diff
@@ -1,172 +1,189 @@
...
 validate_size = 0.2
 
-# test_size: Size of the test split
-# If `float`, should be between `0.0` and `1.0` and represent the proportion of the dataset to include in the 
-# test split.
-# If `int`, represents the absolute number of test samples.
-# If `false`, the value is set to the complement of the train size.
-# If `train_size` is also `false`, it will be set to `0.25`.
+# Size of the test split.
+# If `false`, the value is set to the complement of train_size.
```
mtauraso, line 160: I realize this error existed in the original text, but this is not exactly what we do anymore. I think the current behavior is more like:

```
# if `false`, the value is set to the complement of train_size plus validate_size.
# If `false` and `train_size = false`, test_size is set to `0.25`.
```

src/fibad/fibad_default_config.toml
```diff
@@ -1,172 +1,189 @@
...
+# It is an error for these values to add to more than 1.0 as ratios or the size
+# of the dataset if expressed as integers.
 
-# train_size: Size of the train split
-# If `float`, should be between `0.0` and `1.0` and represent the proportion of the dataset to include in the 
-# train split.
-# If `int`, represents the absolute number of train samples.
-# If `false`, the value is automatically set to the complement of the test size.
+# Size of the train split.
+# If `false`, the value is automatically set to the complement of test_size.
```
mtauraso, line 150: This error existed in the original text, but this is not exactly what we do anymore. The current behavior is more like:
```
# If `false`, the value is automatically set to the complement of test_size plus the validate_size (if any).
```

mtauraso:
Left several comments. I think we need to make sure these are fully accurate before merging.

This is also an opportunity to remove configs that serve no purpose and clutter the file. I believe `offset` needs removal, and I suspect `shuffle` does as well.

src/fibad/fibad_default_config.toml
```diff
@@ -1,172 +1,189 @@
...
 [data_loader]
-# Default PyTorch DataLoader parameters
-batch_size =32
+# The number of data points to load at once.
+batch_size = 32
 
-# We could remove this potentially - pytorch-ignite will default to shuffle=True
-# If the user wanted to explicitly require no shuffling, they could set this to false.
+# Whether the input data is shuffled before being loaded.
 shuffle = false
```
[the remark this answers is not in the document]
drewoldag: I'm interested in keeping it for reproducibility purposes. We could probably set the value to `true` to give the expected behavior out of the box. And then it will provide the place for someone to specify `false` if that's what they need.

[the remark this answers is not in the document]
drewoldag: Actually, never mind. Using samplers in the data sets prevents us from using shuffle, so I'll remove this.

src/fibad/fibad_default_config.toml
drewoldag: Bumped to 512

src/fibad/fibad_default_config.toml
drewoldag: Updated

src/fibad/fibad_default_config.toml
drewoldag: Updated

src/fibad/fibad_default_config.toml
drewoldag: This looks fine enough for now. We can go into more detail in RTD documentation.

src/fibad/fibad_default_config.toml
drewoldag: Done

src/fibad/fibad_default_config.toml
```diff
@@ -1,172 +1,189 @@
...
-# Implementation is dataset class dependent. Default is false meaning now filtering.
+# Path to a fits file that specifies object IDs to use from the data stored in
+# [general].data_dir. Implementation is data_set class dependent. Use `false` for no filtering.
 filter_catalog = false
 
-# How to split the data between training and eval sets.
-# The semantics are borrowed from scikit-learn's train-test-split, and HF Dataset's train-test-split function
-# It is an error for these values to add to more than 1.0 as ratios or the size of the dataset if expressed
-# as integers.
+# For train_size, validation_size, and test_size, the following options are available:
```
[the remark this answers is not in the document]
drewoldag: done

src/fibad/fibad_default_config.toml
drewoldag: Added the example back

src/fibad/fibad_default_config.toml
drewoldag: Removed

src/fibad/fibad_default_config.toml
drewoldag: Updated.

drewoldag:
I believe that all the comments have been resolved or addressed at this point. If there aren't other specific comments, I would say that we merge this and fix the one-offs as we find them.

PR #163, opened 2025-01-15
Optimizer and criterion subtables

drewoldag:
This PR allows defining subtables in the config that will sandbox the parameters used for a particular criterion or optimizer. 

This means that parameters for multiple functions can be included in the config without conflicting with each other - the biggest risk is that different functions might have parameters with the same name, but substantially different meanings.


mtauraso (approved):
Looks great :ship: it

PR #164, opened 2025-01-15
Adding the Capability for any numpy function to be passed as a transform variable

aritraghsh09:
## Change Description

Previously, the HSC DataSet class did a tanh transformation before feeding data into models.
This PR:-
- adds the option for the user to specify transformations or skip transformations altogether
- adds the capability for any `numpy` function to be used as a transform.

- [X ] My PR includes a link to the issue that I am addressing
- This PR addresses https://github.com/lincc-frameworks/fibad/issues/160

## Solution Description

- I introduced a new function, `_get_np_function` in `HSCDataSetContainer` that maps passed strings to `numpy` functions
- I also added a few tests to check different scenarios

### DIFF
--- a/src/fibad/data_sets/hsc_data_set.py
+++ b/src/fibad/data_sets/hsc_data_set.py
@@ -7,7 +7,7 @@ import re
 import resource
 from copy import copy, deepcopy
 from pathlib import Path
-from typing import Any, Literal, Optional, Union
+from typing import Any, Callable, Literal, Optional, Union
 
 import numpy as np
 import torch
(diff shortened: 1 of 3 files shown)


drewoldag (approved):
This seems ok to me if it's unblocking your work. I am curious if there will be a case where multiple transforms will need to be used and specified in the config as an array. If that's not on your immediate radar I wouldn't worry too much about it.

aritraghsh09:
> This seems ok to me if it's unblocking your work. I am curious if there will be a case where multiple transforms will need to be used and specified in the config as an array. If that's not on your immediate radar I wouldn't worry too much about it.

I don't think folks will want to do multiple mathematical transforms. But a change we have to implement at some point is the ability to do non-mathematical transforms [ Flips, Rotations, etc.]. We will probably have to implement that using a slightly different config variable that allows any array of transforms supported by PyTorch.

mtauraso (approved):
LGTM :ship: it

PR #170, opened 2025-01-18
WIP Initial commit to get some things logged to mlflow.

drewoldag:
If testing locally be sure to `pip install mlflow`, and optionally `pynvml` if you want to collect gpu metrics.

You should be able to log model parameters and metrics immediately without any initial setup. To view the results you'll need to start the mlflow server using the following at the command line: 
`mlflow ui --backend-store-uri file://<.../results/mlflow>` 

Pass in the full path to the directory `/<where your results are>/results/mlflow` with a leading `/`. 

Then go to `http://127.0.0.1:5000 and you should be presented with the MLFlow UI. 

The default experiment is `notebook`. You can change the experiment name in your config using:
```
[train]
experiment_name = "fav_experiment_name"
```

In this example, I used port 8080 by starting mlflow server like so: `mlflow ui --backend-store-uri ... --port 8080`
mlflow_example

Example showing the parameter logging - by default, it will log all the keys under `[model]` and the criterion and optimizer specific parameters in the configuration used for the run. Other parameters can be added. The screenshot below shows only the model parameter.
mlflow_example2

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,7 @@ dependencies = [
     "numpy", # Used in Holoviews visualization prototype
     "scipy", # Used in Holoviews visualization prototype
     "cython", # Used in Holoviews visualization prototype
+    "mlflow", # Used to log training metrics and compare models
 ]
 
 [project.scripts]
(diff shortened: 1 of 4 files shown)


mtauraso (approved):
LGTM

PR #176, opened 2025-01-22
Remove config.get, use in line if/else for "rebuild_manifest".

drewoldag:
Remove config.get(...) call, replace this with a slightly less cool, if/else.


drewoldag:
Not sure why that linting error is popping up, but I'll go ahead and fix that too while I'm here.

mtauraso (approved):
LGTM

PR #178, opened 2025-01-23
Pass along the HSCDataset id for populating vectordb

drewoldag:
Copied the structure of `shape(self)`.

### DIFF
--- a/src/fibad/infer.py
+++ b/src/fibad/infer.py
@@ -56,7 +56,7 @@ def run(config: ConfigDict):
 
     write_index = 0
     batch_index = 0
-    object_ids: list[int] = list(int(data_set.ids()) if hasattr(data_set, "ids") else range(len(data_set)))  # type: ignore[arg-type]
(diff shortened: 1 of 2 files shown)


src/fibad/infer.py
```diff
@@ -56,7 +56,7 @@ def run(config: ConfigDict):
 
     write_index = 0
     batch_index = 0
-    object_ids: list[int] = list(int(data_set.ids()) if hasattr(data_set, "ids") else range(len(data_set)))  # type: ignore[arg-type]
```
mtauraso, line 59: NIT: I think its still clearer to maintain the invariant that this is a list of ints, but I don't think it breaks anything.

mtauraso (approved):
Looks good!

src/fibad/infer.py
drewoldag: I can make the change here so that we're returning int's. Later it will be worth while thinking through whether we'll need to support str's more. I can imagine a compound id that's filename_index_into_file for datasets that are composed of batched files.

Issue #108, opened 2024-10-29
AttributeError: module 'importlib' has no attribute 'util'


drewoldag:
**Bug report**
@maxwest-uw is working with fibad and kbmod-ml, and ran into this error when trying to push changes. It's clearly lines in Fibad that are causing this:

> Traceback (most recent call last):
>   File "/astro/users/maxwest/fibad_script.py", line 7, in <module>
>     fibad_instance = fibad.Fibad(config_file="kbmod_fibad_config.toml")
>   File "/astro/users/maxwest/fibad/src/fibad/fibad.py", line 46, in __init__
>     self.config_manager = ConfigManager(runtime_config_filepath=config_file)
>   File "/astro/users/maxwest/fibad/src/fibad/config_utils.py", line 85, in __init__
>     self.external_library_config_paths = self._find_external_library_default_config_paths(
>   File "/astro/users/maxwest/fibad/src/fibad/config_utils.py", line 136, in _find_external_library_default_config_paths
>     default_configs |= ConfigManager._find_external_library_default_config_paths(value)
>   File "/astro/users/maxwest/fibad/src/fibad/config_utils.py", line 140, in _find_external_library_default_config_paths
>     if importlib.util.find_spec(external_library) is not None:
> AttributeError: module 'importlib' has no attribute 'util'

drewoldag:
I spent some time trying to reproduce this locally on the command line and wasn't able to do so. Running the following on my local machine with Python 3.11 was fine:

```
import importlib
importlib.util.find_spec("asdf")
```

In some examples of people using find_spec, I see that they typically import the following:
```
import importlib
import importlib.util
...
```
So perhaps that is necessary on some systems?

I also took a look through the python code base to see if perhaps importing just importlib would not include `util`, but it does seem to include it. 

So I'll put together a little PR that adds the second import importlib.util statement and we can see if that fixes things.

drewoldag:
Closed the PR that was associated with this issue since it's not clear that it's actually addressing the root cause. Will reopen the PR (or close this issue) after I have a chance to do a little more investigation with @maxwest-uw in the coming weeks.

drewoldag:
I haven't had a chance to review this with @maxwest-uw, will try to do that before Jan 10

drewoldag:
I've been working with kbmod_ml for several days now and haven't been able to reproduce this error. Going to close as won't do.

drewoldag:
[...]

PR #181, opened 2025-01-23
Being more explicit with the import of importlib.util.

drewoldag:
This seems to address the issue that was raised in issue #108. Still not immediately clear why this is sometimes an issue, but it seems to be.

### DIFF
--- a/src/fibad/config_utils.py
+++ b/src/fibad/config_utils.py
@@ -4,6 +4,7 @@ import importlib
 import logging
 import random
 import re
+from importlib import util as importlib_util
 from pathlib import Path
 from typing import Optional, Union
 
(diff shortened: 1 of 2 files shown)


src/fibad/config_utils.py
```diff
@@ -155,7 +156,7 @@ def _find_external_library_default_config_paths(runtime_config: dict) -> set:
             else:
                 if key == "name" and "." in value:
                     external_library = value.split(".")[0]
-                    if importlib.util.find_spec(external_library) is not None:
```
mtauraso, line 158: OK that is really weird.

mtauraso (approved):

src/fibad/config_utils.py
mtauraso: I wonder if some package deep in fibad or KBMOD's dep tree is conditionally replacing functions in importlib.util?

drewoldag: When we start python on the command line and run the following on baldur in an environment that shares a kbmod installation
```
>>> import importlib
>>> importlib.ut <attempt tab complete, nothing happens>
>>> importlib.util
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'importlib' has no attribute 'util'
```

It does seem unpleasant, but it seems to work.

PR #194, opened 2025-02-04
Configuration help in notebook

drewoldag:
Implementing simple function to print specific config tables. The goal is to allow a user to call `fibad_instance.config.help(...)` and get see the configuration comments printed. 

It's unintuitive how to parse through the data structure that tomlkit maintains in order to print out just the comments surrounding the key of interest. So as a half set, we can just print the entire enclosing table. 

So for instance:
```
import fibad
f = fibad.Fibad()
f.config.help('general', 'dev_mode')
```

Prints the full `[general]` table:
```
[general]
# Set to `true` during development to skip checking for default config values
# in external libraries. Use `false` otherwise.
dev_mode = false

# Destination of log messages. Options: 'stderr', 'stdout' specify the console,
# "path/to/fibad.log" specifies a file.
log_destination = "stderr"

# Lowest log level to emit. Options: "critical", "error", "warning", "info", "debug".
log_level = "info"

# Directory where data is stored.
data_dir = "./data"

#Top level directory for writing results.
results_dir = "./results"
```

Usage for the help looks like this:
```
f = fibad.Fibad()

f.config.help()  # Pretty prints the entire config

f.config.help('table_name')  # Pretty prints the config table

f.config.help('table_name', 'key_name')  # Pretty prints the config table

f.config.help('key_name')  # Searches for the key in all tables, prints each table that contains it

f.config.help('non-existant key or table')  # Prints a nice error message

f.config.help('table', 'key', 'third value')  # Prints a nice error with example usage.
```

### DIFF
--- a/src/fibad/config_utils.py
+++ b/src/fibad/config_utils.py
@@ -368,3 +368,93 @@ def log_runtime_config(runtime_config: ConfigDict, output_path: Path, file_name:
     """
     with open(output_path / file_name, "w") as f:
         f.write(tomlkit.dumps(runtime_config))
+
+
+def config_help(config: TOMLDocument, *args):
+    """
+    Early version of a config help function. It's a bit difficult to parse through
+    the Tomlkit Table to print just one item such that it would include the comments
+    preceding it.
+
+    For now, we support the following cases, and generally print out the entire
+    table for the given key.
+
+    Cases:
+    - if no args, prints the whole config.
+    - if args[0] is a table name, print the whole table
+    - if args[0] is not a table, assume it's a key and search
+    -- print each one of the tables that it is found in.
+
+    Parameters
+    ----------
+    config : TOMLDocument
+        A configuration dictionary that will be used to search for specified tables
+        and keys.
+
+    args : str
+        A variable number of string arguments that specify the table name or key
+        to search for in the configuration dictionary.
+    """
+
+    config_dict = config.value
+
+    # No tables provided, print the whole config
+    if not args:
+        print(config.as_string())
+
+    # Table name provided as args[0], print that config table
+    if len(args) == 1 and args[0] in config_dict:
+        print(f"[{args[0]}]")
+        print(config[args[0]].as_string())
+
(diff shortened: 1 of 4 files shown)


drewoldag:
Requesting an early stage review on this PR, mostly because I want to know if the current state is acceptable, or if it's so annoying that people would be turned off by it.

src/fibad/fibad.py
```diff
@@ -135,6 +135,22 @@ def _initialize_log_handlers(self):
...
+        ```
+        f = Fibad()
+        f.config_help()  # To print the whole configuration
+        f.config_help("<section>")  # To print a particular section of the config.
+        f.config_help("<section>", "<key>")  # To print a section containing a key.
+        f.config_help("<key>")  # Print all sections that contain the given key.
+        """
+
+        config_help(self.config, *args)
+
```
mtauraso, line 153: I like this. We might want to look into how to bind this to the fibad (or fibad.config) object such that it is discoverable.

src/fibad/config_utils.py
```diff
@@ -368,3 +368,93 @@ def log_runtime_config(runtime_config: ConfigDict, output_path: Path, file_name:
     """
     with open(output_path / file_name, "w") as f:
         f.write(tomlkit.dumps(runtime_config))
+
+
+def config_help(config: TOMLDocument, *args):
```
mtauraso, line 373: This approach of "just print the relevant section with comments" is a good first order solution.

src/fibad/fibad.py
drewoldag: Yeah, pulled this off the Fibad class and monkey patched a `help` method onto `TOMLDocument`. If it's needed, we can add a proxy back to the Fibad class so to proxy to the TOMLDocument.help. But for now, we'll just do `fibad_instance.config.help('...')`

mtauraso (approved):
Looks good. Similar to what I said on the WIP PR: I think this is a good interface for us to start with and extend/experiment with.

PR #196, opened 2025-02-04
Update simple training example notebook

drewoldag:
Cleaning up the `train_model.ipynb` file so that it can be used for documentation.

This notebook was failing the documentation build because it was trying to load data from a directory that didn't exist. (It was assuming that a HSC example data directory was present)


aritraghsh09 (approved):
This looks good to me! Ready to be 🚢-ed

Issue #161, opened 2025-01-15
Extract split generation code from data_sets


drewoldag:
Currently both hsc_data_set and example_cifar_data_set implement code that generates splits based on the config values `[data_set].train_size|validation_size|test_size`. 

It represents a pretty good chunk of code, and we have for the time being, agreed on the way that the *_size values are suppose to work. i.e. how to fill in the values if some of the data is missing.

If we can abstract this code out that will make it much easier for external libraries to implement data_sets without having to worry about implementing our agreement in their code. 

Long story short: we should do our best to pull the `sampler` code out of hsc_data_set and example_cifar_data_set and move it to data_set_registry so that we can add it to internal and external data_sets when they are imported. 

example_cifar_data_set: https://github.com/lincc-frameworks/fibad/blob/main/src/fibad/data_sets/example_cifar_data_set.py#L29-L49
hsc_data_set: https://github.com/lincc-frameworks/fibad/blob/main/src/fibad/data_sets/hsc_data_set.py#L58-L146

mtauraso:
So far, looking at this I'm taking the following decisions (major disagreements with these are most welcome 😅 ):
- Alignment of our code with HF dataset splits isn't very important. Why? I clicked through ~10 popular image datasets on HF, and it seems like dataset authors on HF generally use one split ("train") or they name their own ("HSC 1000", "HSC 10k") to pack up multiple datasets into one HF repository. Bottom line: HF ecosystem usage doesn't align with test/train/validate as we mean them in FIBAD.
- The approach to splits in HSCDataSet puts the random choices in our code, but it is cumbersome given that torch.data.SubsetRandomSampler and tf.random.uniform_candidate_sampler exist, and appear to give us enough control to create reproducibility.

Given those decisions the plan looks like:
- Change the dataset creation path used by "infer", "train", and "prepare" to create SubsetRandomSamplers with the appropriate index ranges, removing this code from Cifardataset
- Remove split from all dataset constructors.
- Condense HSCDataSet to functionally remove the HSCDataSet/HSCDataSetContainer/HSCDataSetSplit class distinctions.

This plan has some impacts on (future) external dataset authors:
- Dataset authors write any code that deals with splits as they are meant in FIBAD config. A dataset is an ordered list of data objects, and its a user decision how it is split between "test"/"train"/"validate"
- Dataset authors must (until we write more code!) make their datasets zero-indexable, with a set size at runtime rather than simply implementing `__iter__` All our dataset classes already work this way.

PR #199, opened 2025-02-05
Protect against 0-length validation splits in cifar dataset

drewoldag:
This PR amounted to duplicating a lot of code that was in hsc dataset. 

The protection and approach is broadly applicable, so it probably makes sense to abstract this up a level and inject the code when a dataset class uses the @fibad_dataset decorator. 

The only reason I didn't take that step here was because I want to take a minute to think clearly about how this will work as we expand beyond just Pytorch data sets. A little research into the HuggingFace dataset API would be good.

### DIFF
--- a/src/fibad/data_sets/example_cifar_data_set.py
+++ b/src/fibad/data_sets/example_cifar_data_set.py
@@ -1,10 +1,14 @@
 # ruff: noqa: D101, D102
+import logging
+
 import numpy as np
 import torchvision.transforms as transforms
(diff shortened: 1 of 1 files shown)


mtauraso (approved):
Agree there's probabably a way to give this functionality to datasets via @fibad_dataset

drewoldag:
Issue #161 captures the sentiment of wanting to pull the splitting code out of the dataset and provide it via @fibad_dataset. Seems like a good task for post-MPR but pre-hack week.

PR #200, opened 2025-02-05
UMAP Enchancements 

aritraghsh09:
This PR:-

- Allows UMAPs to be created irrespective of the dimensionality of the arrays produced by `infer`. The arrays are flattened accordingly, and this is logged as a message. 
- Logs the output UMAP directory
- Allows users to pass various arguments to UMAP. 

@mtauraso I saw your comment about whether the `umap` kwargs should be supplied via the config file or via arguments to umap. I went and implemented the second one given the long list of variables that umap accepts (and quite a few of those are important). No strong feelings about this though. It can be changed to moving these to the config file if there is a strong preference for that.


src/fibad/verbs/umap.py
```diff
@@ -33,26 +34,43 @@ def setup_parser(parser: ArgumentParser):
...
     def run_cli(self, args: Optional[Namespace] = None):
         """Stub CLI implementation"""
         logger.info("Search run from cli")
         if args is None:
             raise RuntimeError("Run CLI called with no arguments.")
+
+        # Get valid UMAP parameters and extract those args to pass to umap
+        valid_umap_params = set(inspect.signature(umap.UMAP).parameters.keys())
+        umap_params = {k: v for k, v in vars(args).items() if k in valid_umap_params and v is not None}
+
```
aritraghsh09, line 61: @mtauraso Please check whether this seems like a good way to achieve what I want to do. 

Given my limited understanding of how exactly the CLI variable backend is working, I was worried whether there is any risk in the above code that some other variables might get mismatched to the UMAP variables?

src/fibad/verbs/umap.py
```diff
@@ -81,6 +107,8 @@ def run(self, input_dir: Optional[Union[Path, str]], **kwargs):
         all_ids = np.array([int(i) for i in inference_results.ids()])
         for batch_indexes in np.array_split(all_indexes, num_batches):
             batch = inference_results[batch_indexes]
+            if reshape:
```
aritraghsh09, line 110: Doing this primarily to avoid having to make `.shape` call in every batch

src/fibad/verbs/umap.py
drewoldag: We've been generally trying pretty hard to avoid introducing new command line arguments. Are you opposed to using the config file to pass the umap parameters? 

An existing approach that would probably work well here is similar to what is done for criterion and optimizer functions. 

In the config they look like: 
```
[optimizer]
name = "torch.optim.SGD"

["torch.optim.SGD"]
lr = 0.01
... other named parameters that the function expects ...
```

You could do something similar here: 
```
[umap]
# Number of data points used to fit the umap transform.
fit_sample_size = 1024
name = "umap.UMAP"

["umap.UMAP"]
param1 = val1
param2 = val2
```

Then where reducer is defined:
```
reducer = umap.UMAP(**self.config["umap.UMAP"])
```

src/fibad/verbs/umap.py
```diff
@@ -81,6 +107,8 @@ def run(self, input_dir: Optional[Union[Path, str]], **kwargs):
         all_ids = np.array([int(i) for i in inference_results.ids()])
         for batch_indexes in np.array_split(all_indexes, num_batches):
             batch = inference_results[batch_indexes]
```
mtauraso, line 109: why not: `batch = inference_results[batch_indexes].flatten()` here?

I think this could remove the need for the `reshape` bool, and make the code easier to read.

I'm not sure we need the log message about reshaping either, because it doesn't make sense to run umap on a latent space tensor that has shape in the first place. Maybe a docstring stating that umap is run on the flattened inference output?

mtauraso (changes requested):
Left some comments, I think we want to be careful and thoughtful before adding this many CLI args, and I think there's maybe a simpler way to implement the flattening and still be clear about what's happening under the hood in `umap`

src/fibad/fibad_default_config.toml
```diff
@@ -200,3 +200,9 @@ inference_dir = false
 [umap]
 # Number of data points used to fit the umap transform.
 fit_sample_size = 1024
+name = "umap.UMAP"
+
+["umap.UMAP"]
+# Specify any parameter accepted by https://umap-learn.readthedocs.io/en/latest/api.html#umap
+n_components = 2
+n_neighbors = 15
```
drewoldag, line 208: Formatting suggestion for the default config, since we're trying to use this as a form of documentation as well...

```suggestion

# Name of the umap implementation to use
name = "umap.UMAP"


["umap.UMAP"]
# Specify any parameter accepted by https://umap-learn.readthedocs.io/en/latest/api.html#umap
# Dimension of the embedded space
n_components = 2

# Controls how UMAP balances local versus global structure in the data.
# See official documentation for details.
n_neighbors = 15
```

src/fibad/verbs/umap.py
```diff
@@ -64,7 +82,9 @@ def run(self, input_dir: Optional[Union[Path, str]] = None, **kwargs):
         sample_size = np.min([config_sample_size if config_sample_size else np.inf, total_length])
         rng = np.random.default_rng()
         index_choices = rng.choice(np.arange(total_length), size=sample_size, replace=False)
-        data_sample = inference_results[index_choices].numpy()
+
+        # If the input to umap is not of the shape [samples,input_dims] we reshape the input accordingly
+        data_sample = inference_results[index_choices].numpy().reshape(sample_size, -1)
```
drewoldag, line 87: This took me a little while to fully understand what was going on here. The `-1` syntax was unfamiliar, perhaps an update to this comment could be:
`# Reshape the input data to be an array of 1d arrays`

Perhaps this is very well know numpy syntax, and if so, feel free to ignore, but would the following suggestion work? It seems a bit more intuitive to me to specify the shape as a tuple. Similar story for line 107. 
```suggestion
        data_sample = inference_results[index_choices].numpy().reshape((sample_size, -1))
```

src/fibad/verbs/umap.py
```diff
@@ -79,8 +99,12 @@ def run(self, input_dir: Optional[Union[Path, str]] = None, **kwargs):
...
         all_indexes = np.arange(0, total_length)
         all_ids = np.array([int(i) for i in inference_results.ids()])
-        for batch_indexes in np.array_split(all_indexes, num_batches):
-            batch = inference_results[batch_indexes]
+        for batch_indexes in tqdm(
+            np.array_split(all_indexes, num_batches),
+            desc="Creating Lower Dimensional Representation using UMAP",
+            total=num_batches,
+        ):
+            batch = inference_results[batch_indexes].reshape(len(batch_indexes), -1)
```
drewoldag, line 107: A comment for this line would be good too I think.

drewoldag (approved):
I left a couple of comments that aren't huge, but might make it a bit easier for to read going forward.

src/fibad/verbs/umap.py
aritraghsh09: Done!

PR #201, opened 2025-02-06
Fix github pre-commit mypy and import checks

drewoldag:
Attempting to introduce enough typing in module_registry that mypy in pre-commit github actions will pass.


mtauraso (approved):
LGTM

PR #202, opened 2025-02-06
Adding a UMAP section to the notebook demo.

drewoldag:
Just adding a UMAP section and plot. Also added a space in one of the info logs.


aritraghsh09 (approved):
This looks like a fantastic addition!

PR #208, opened 2025-02-08
Added interactive visualization to the notebook.

drewoldag:
Updating the MPR demo notebook to include a cell showing interactive visualization.


drewoldag:
FYI, I'm merging this demo notebook change in without a review.

PR #211, opened 2025-02-11
Allow user to specify MLFlow run name via the config.

drewoldag:
This PR introduces a new config value, `['train']['run_name']`. The default value is `false`. 

When this value is false and the user runs `fibad.train()`, the resulting MLFlow run will be the same as the auto-generated results directory, i.e. `<timestamp>-train-<uid>`. 

When the user provides a value with a string representation, i.e. "super_turtle", that value will be used as the MLFlow run name.

In the screenshot below the first in the list is user defined, the second in the list is the default timestamped value. The third item in the list is a previously autogenerated run name.  

Screenshot 2025-02-10 at 4 06 18 PM

### DIFF
--- a/src/fibad/fibad_default_config.toml
+++ b/src/fibad/fibad_default_config.toml
@@ -125,6 +125,10 @@ split = "train"
 # The name of the experiment when logging training results to mlflow
 experiment_name = "notebook"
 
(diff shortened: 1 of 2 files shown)


aritraghsh09 (approved):
Looks great! Thanks!

PR #221, opened 2025-02-18
Parallelize umap with process pools

mtauraso:
- We use a process pool to do the transform
- We use another process pool to do writing of inference data out (also impacts infer verb)
- Some of the loudest warnings from the umap package have been supressed


aritraghsh09 (approved):
Looks good to me! No comments!

drewoldag (approved):
This looks good. I haven't had a chance to run it, but what I see makes sense.

PR #222, opened 2025-02-18
Removing create_index codepath

mtauraso:
We should only merge this once we're sure we don't need this codepath anymore.

i.e. all inference results we care about ought have a batch index file.


aritraghsh09 (approved):
Confirming that we don't need this codepath anymore.

PR #226, opened 2025-02-21
Initial investigation of exporting to ONNX after training.

drewoldag:
This PR is an first step toward exporting multiple forms of models. Obviously at this point we're only focusing on PyTorch, but I've tried to write this in a way that will support future different ML frameworks. I probably haven't done this in an ideal way, but it's a start. Two aspects of this are worth noting, and I would appreciate feedback on them.

### A context object
I've created a simple dictionary that carries "context" from the train.run function into the export_to_onnx function. This context object sounds a lot like a python `dataclass`. Additionally, it feels like a reasonable way to percolate information back up to the fibad object after `train` has finished. Extra-additionally, perhaps this is the right time to transform `train` into a proper verb class.

### Using `@singledispatch`
In the `model_exporters` module, the function `export_to_onnx` will accept the context object and then dispatch the request to the appropriate private function that implements the conversion to ONNX. As it stands, it's only the pytorch->ONNX function, but we're trying to plan ahead. This is an ideal use case for `@singledispatch` where we use the model type as the discriminator, i.e. `torch.nn.Module`. However, to do so, _I believe_ would require that we import torch at the top level. And while torch is pretty quick to import, I'm not sure about the other libraries, so I wanted to defer the imports (in the same way we do that with verbs) to avoid a really long "start up" time at the end of `.train()`. At this point, I'm using a string in the context object (`context['ml_framework'] == 'pytorch'`) to determine which specific implementation to use. And while that's fine, it doesn't feel as cool as using `@singledispatch`...

## Remaining work

- [ ] Implement something on the `.infer` side so that we can make use of the .onnx model
- [ ] Take some benchmarks to see if .onnx is actually faster or not
- [ ] A little more research into versions. ONNX (opset_version and IRVersion), PyTorch (support for different opset and IR versions), PyTorch-ignite (newer versions of PyTorch)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -157,3 +157,6 @@ example_model.pth
 # Common location to stash personal or notebook example config
 fibad_config.toml
 credentials.ini
+
+# MLFlow data directory
+mlruns/
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,6 +39,8 @@ dependencies = [
     "pynvml", # Used to gather GPU usage information
     "umap-learn", # Used to map latent spaces down to 2d
     "pooch", # Used to download data files
+    "onnx", # Used to export models to ONNX format
+    "onnxruntime", # Used to run ONNX models
 ]
 
 [project.scripts]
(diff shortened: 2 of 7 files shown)


mtauraso (approved):
I'm in favor of merging this more-or-less as-is, as a first step toward a more full onnx integration.

mtauraso:
Looks good to me still 👍

PR #227, opened 2025-02-22
Extracting dataset split code from data_sets

mtauraso:
- Data Sets no longer have any awareness of splits
- All splitting behavior is negotiated between the verb and pytorch_ignite while creating distributed data loader(s).
- HSCDataSetSplit and HSCDataSetContainer are removed entirely, with HSCDataSet now having the code that was in HSCDataSetContainer.
- CifarDataSet is much shorter and only uses the training CIFAR dataset
- Tests of splits are in their own file and preserve behavior with prior HSCDataSetSplit tests.

### DIFF
--- a/docs/pre_executed/visualize.ipynb
+++ b/docs/pre_executed/visualize.ipynb
@@ -30,34 +30,35 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 16,
+   "execution_count": 2,
    "metadata": {},
    "outputs": [
     {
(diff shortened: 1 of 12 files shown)


mtauraso:
Before this merges I need to do some notebook/cli testing; however, I think this is ready for review.

drewoldag (approved):
This looks pretty good to me, a good refactor and definitely the direction we need to go. I'm glad that all of the split logic is consolidated in one place.

PR #230, opened 2025-02-26
Prevent recursive `_build` directories in docs

drewoldag:
Adding `_build` to the list of excluded directories when pre-commit builds documentation.


drewoldag:
This seems to have done what it was suppose to do, so I'm going to go ahead and merge. We can always revert if needed.

PR #233, opened 2025-03-04
Initial vectordb interface and chromadb impl with sharding

drewoldag:
Implementing sharding for the chromadb vector database because we have seen that the insertion times become overly burdensome as total size of the "collection" (database table) becomes large, by some definition of large. 

Sharding helps keep the insertion costs down with the downstream cost of less efficient nearest neighbor look ups. Now we must scan all the shards for nearest neighbors instead of just scanning a single collection.

### DIFF
--- a/src/fibad/fibad_default_config.toml
+++ b/src/fibad/fibad_default_config.toml
@@ -198,9 +198,6 @@ model_weights_file = false
 # The data_set split to use for inference. Use `false` for entire dataset.
 split = false
 
-# Whether to generate a chromadb vector database of inference results
-chromadb = true
-
 
 [results]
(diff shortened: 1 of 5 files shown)


src/fibad/fibad_default_config.toml
```diff
@@ -223,3 +220,8 @@ n_components = 2
 # Controls how UMAP balances local versus global structure in the data.
 # See official documentation for details.
 n_neighbors = 15
+
+
+[vector_db]
+# The type of vector db to use. Use "false" to disable vector database.
```
mtauraso, line 226: Nice.

mtauraso (approved):
Looks good to me

PR #235, opened 2025-03-08
Metadata interface v1 for Datasets

mtauraso:
- HSCDataSet, InferenceDataSet, and CifarDataSet all have:
  - original_config() to fetch the config required to (re)create the dataset with the metadata
  - metadata_fields() which returns a list of metadata fields
  - metadata() which takes a set of indexes and fields and fills out a table of metadata, passing it back.
- InferenceDatasetWriter saves out the original config for a given dataset
- InferenceDataSet reads in the original config from its f/s rather than using the runtime config as the other datasets do
- infer, umap, and visualize verbs are all modified to use the new interfaces
- visualize verb currently hardcodes ra/dec as metadata fields and displays ra/dec in a table as a POC.


drewoldag (approved):
I don't want to hold things up any longer. This work seems reasonable enough.

PR #244, opened 2025-03-21
Instrumenting HSCDataSet using TensorBoardX

mtauraso:
- New interface on setup_dataset to provide a tensorboardx logger to datasets right after construction.
- Datasets need not define the attribute setup_dataset sets, so logging metrics is entirely optional for a dataset writer
- File caching and reading in HSCDataSet are extensively measured
- Measurements use time.monotoinic_ns(), are logged as floating point numbers of seconds
- The time series given to tensorboard uses integer microseconds as the y axis.
- Tensorboard metrics are logged for datasets during infer runs as well.

### DIFF
--- a/src/fibad/data_sets/hsc_data_set.py
+++ b/src/fibad/data_sets/hsc_data_set.py
@@ -5,6 +5,7 @@ import multiprocessing
 import os
 import re
 import resource
+import time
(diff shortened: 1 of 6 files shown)


aritraghsh09 (approved):
Looks good to me!

PR #245, opened 2025-03-22
First implementation of cache warming for HSCDataSet

mtauraso:
This spawns a pre-cache thread during HSCDataSet initialization. The pre-cache thread starts many worker threads that read in all the dataset files to the tensor cache. The idea is to confront a latency-limited network storage backend with a large number of requests all at once, such that we will receive the files in memory at the greatest overall bandwidth.

This change also logs a `HSCDataSet/preload_1k_obj_s` metric to tensorboard which should give an idea of object bandwidth. It is the elapsed time to load 1000 objects measured every 1000 objects.

The change is protected under a config `config["data_set"]["preload_cache"]` The config is on by default (similar to `use_cache`) because it should be harmless for small datasets and laptop usage.

### DIFF
--- a/docs/pre_executed/mpr_demo.ipynb
+++ b/docs/pre_executed/mpr_demo.ipynb
@@ -98,12 +98,12 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 7,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
     "# Specify the location of the data to use for training\n",
-    "f.config[\"general\"][\"data_dir\"] = \"../../data/hsc_8asec_1000_bu\"\n",
+    "f.config[\"general\"][\"data_dir\"] = \"../../data/hsc_8asec_1000\"\n",
     "\n",
     "# Specify the dataset class that represents the data\n",
     "f.config[\"data_set\"][\"name\"] = \"HSCDataSet\"\n",
--- a/src/fibad/data_sets/hsc_data_set.py
+++ b/src/fibad/data_sets/hsc_data_set.py
@@ -7,8 +7,10 @@ import os
 import re
 import resource
 import time
-from collections.abc import Generator
+from collections.abc import Generator, Iterable, Iterator
+from concurrent.futures import Executor
 from pathlib import Path
+from threading import Thread
 from typing import Any, Callable, Optional, Union
 
 import numpy as np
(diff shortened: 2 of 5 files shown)


docs/pre_executed/mpr_demo.ipynb
```diff
@@ -98,12 +98,12 @@
...
   {
    "cell_type": "code",
-   "execution_count": 7,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
     "# Specify the location of the data to use for training\n",
-    "f.config[\"general\"][\"data_dir\"] = \"../../data/hsc_8asec_1000_bu\"\n",
+    "f.config[\"general\"][\"data_dir\"] = \"../../data/hsc_8asec_1000\"\n",
```
mtauraso, line 106: Minor typo that was making this notebook not work for testing.

src/fibad/data_sets/hsc_data_set.py
```diff
@@ -779,6 +792,84 @@ def _file_to_path(self, filename: str) -> Path:
...
+
+        try:
+            while True:
+                for _ in range(max_futures - len(in_progress)):
+                    id = next(ids_iter)
+                    future = executor.submit(self._read_object_id.__func__, self, id)  # type: ignore[attr-defined]
+                    queue.append(future)
+                    in_progress.add(future)
+
+                _, in_progress = wait(in_progress, return_when=FIRST_COMPLETED)
```
drewoldag, line 858: This seems pretty neat. I don't immediately understand why it is preferable to use `FIRST_COMPLETED` instead of `ALL_COMPLETED` here though. It seems like ultimately we wait for all to complete, in the next while loop, right? 

Is it that this approach allows us to start returning things sooner off the queue?

drewoldag (approved):
This looks pretty cool to me. I did leave a question, but I'm pretty sure that it doesn't affect the work here.

src/fibad/train.py
drewoldag, line 1: Looks great, thanks for explaining the reasoning here.

src/fibad/data_sets/hsc_data_set.py
```diff
@@ -812,6 +825,94 @@ def _file_to_path(self, filename: str) -> Path:
...
 
+    @staticmethod
+    def _determine_numprocs_preload():
+        # This is hardcoded to a reasonable value for hyak
+        # TODO: Unify this function and _determine_numprocs(). Ideally we would have
+        # either a multiprocessing.Pool or concurrent.futures.Executor interface taking
+        # an i/o bound callable which returns a number of bytes read. This reusable
+        # component would titrate the number of worker threads/processes to achieve a
+        # maximal throughput as measured by returns from the callable vs wall-clock time.
+        return 50
```
drewoldag, line 836: :shipit:

drewoldag (approved):
Still looks good to me

PR #251, opened 2025-03-26
Initial dataset class interface

mtauraso:
This is a skeleton of a superclass for Datasets. The intent is to provide a scaffold for folks writing dataset classes such that they have:
1) Clear instructions on what functions to implement
2) "Batteries included" for the more complex features of Dataset classes that HSC relies on (metadata, original_config, etc)

The version below is a noop change to fibad's functionality, and the individual abstract methods are commented with how they would be implemented in the base class to make future dataset classes easier to write.

I'm looking for feedback on the overall interface at this point, since this code doesn't do anything at the moment 🙂

### DIFF
--- a/src/hyrax/data_sets/__init__.py
+++ b/src/hyrax/data_sets/__init__.py
@@ -1,12 +1,6 @@
-from .data_set_registry import DATA_SET_REGISTRY, hyrax_data_set
+from .data_set_registry import DATA_SET_REGISTRY, Dataset
 from .hsc_data_set import HSCDataSet
 from .hyrax_cifar_data_set import HyraxCifarDataSet
 from .inference_dataset import InferenceDataSet
 
-__all__ = [
-    "hyrax_data_set",
-    "DATA_SET_REGISTRY",
-    "HyraxCifarDataSet",
-    "HSCDataSet",
-    "InferenceDataSet",
-]
+__all__ = ["DATA_SET_REGISTRY", "HyraxCifarDataSet", "HSCDataSet", "InferenceDataSet", "Dataset"]
--- a/src/hyrax/data_sets/data_set_registry.py
+++ b/src/hyrax/data_sets/data_set_registry.py
@@ -1,29 +1,118 @@
+# ruff: noqa: D102, B027
 import logging
+from abc import ABC, abstractmethod
+from collections.abc import Generator
 
-from torch.utils.data import Dataset
+import numpy.typing as npt
 
+from hyrax.config_utils import ConfigDict
 from hyrax.plugin_utils import get_or_load_class, update_registry
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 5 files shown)


aritraghsh09:
I am not 100% certain about what the final intended form of `src/fibad/data_sets/data_set_registry.py` is; but here are my comments:

- I think for each *major* data type that we intend to support (i.e., images, time-series) there should be a base class already implemented and ready-to-use with *some basic* "batteries included" (i.e., optional pre-caching, etc.). For example, there could be a fully working FITS DataSet class ready to go. 
     - It's okay to impose restrictions on what the user can and cannot do in such a class (e.g., they might be required to name the initial `.csv` or table `.fits` in a certain way; or they might be required to name the filename column in a very specific way) 
     - I am sure for each major data-type, we can identify minimal features needed for a fully working DataSet class.
     - Then, we can have a guide with comments as is currently done in `src/fibad/data_sets/data_set_registry.py` and/or in the documentation on what exact functions the user needs to edit to make certain important changes e.g., changing the column names) or how they could include more advanced features (e.g., advanced pruning).
                
- In addition to the 2-3 ready-to-use data-type specific classes, we can also have a fully flexible Dataset class  with comments (as is currently done in `src/fibad/data_sets/data_set_registry.py`) to help users who want to take a stab at writing a class from complete scratch.

aritraghsh09 (commented):
I am not 100% certain about what the final intended form of `src/fibad/data_sets/data_set_registry.py` is; but here are my comments:

- I think for each *major* data type that we intend to support (i.e., images, time-series) there should be a base class already implemented and ready-to-use with *some basic* "batteries included" (i.e., optional pre-caching, etc.). For example, there could be a fully working FITS DataSet class ready to go. 
     - It's okay to impose restrictions on what the user can and cannot do in such a class (e.g., they might be required to name the initial `.csv` or table `.fits` in a certain way; or they might be required to name the filename column in a very specific way) 
     - I am sure for each major data-type, we can identify minimal features needed for a fully working DataSet class.
     - Then, we can have a guide with comments as is currently done in `src/fibad/data_sets/data_set_registry.py` and/or in the documentation on what exact functions the user needs to edit to make certain important changes e.g., changing the column names) or how they could include more advanced features (e.g., advanced pruning).
                
- In addition to the 2-3 ready-to-use data-type specific classes, we can also have a fully flexible Dataset class  with comments (as is currently done in `src/fibad/data_sets/data_set_registry.py`) to help users who want to take a stab at writing a class from complete scratch.

mtauraso:
Generally agree with @aritraghsh09 's comment. With this change I'm setting up the groundwork so it is easier for those classes to be implemented. 

I do think the vision articulated is closer to the size of a roadmap item, and the immediate focus of this change (and the next few based on this interface) is to get us to a place where anyone wanting to put a dataset in at the LINCC Up is empowered to write a small number of functions (2-3) rather than the current number (5-6).

PR #253, opened 2025-03-26
Rename FIBAD to Hyrax

drewoldag:
# The big renaming

- Ran `copier update` to remove mypy, update supported python versions.
- Ran find/replace for `fibad`, `Fibad`, and `FIBAD`.
- Updated notebooks as best I could.
- Documentation has been given a light once-over, since we'll be doing more polish later, I felt it was ok to be quick with this.
- Renamed the two builtin models and the built in cifar dataset to be Hyrax-ified.
- Typed "hyrax" enough that it has lost all meaning.

Still need to:

- [x] Update fibad references in the various notebooks.
- [x] Remove assorted junk files that have accumulated over time (out of date examples, random files)
- [x] Rename the built in models and datasets. i.e. `HyraxCifarDataSet` or `HyraxAutoencoder`
- [x] Clean up the README doc and make some basic edits to the documentation so that it doesn't say things like "Hyrax - The Framework for Image-Based Anomaly Detection..."


mtauraso (approved):
This looks good to me. I think we should merge asap and then start working follow-ups.

aritraghsh09 (approved):
Merge it!

PR #259, opened 2025-03-27
Move Dataset code into base classes pt1

mtauraso:
- Configuration subsystem now handles path resolution for a list of config keys. Hardcoded for now, but we can add as external library config interface takes shape.
- original_config removed from base class interface. Logic can live exclusively in InferenceDataset/InferenceDatasetWriter with the config subsystem handling path resolution
- Default implementation of ids() and shape() in base class
- s/Dataset/HyraxDataset to avoid clashing with torch


drewoldag (approved):
This all looks pretty reasonable.

PR #261, opened 2025-03-28
Move dataset code into base class pt 2: Metadata edition

mtauraso:
- Metadata accessors `metadata()` and `metadata_fields()` moved to HyraxDataset base class
- HSCDataset and HyraxCifarDataset modified to use the new interface
- HyraxDataset constructor optionally takes a metadata_table for subclasses that want to opt in to metadata

Sidebar:
- Added a precommit hook which looks for 'xcxc'/'XCXC' across all source files. This is a string you can use in a note-to-self or work-in-progress comment which will flag to you so you don't forget to handle it before checking in.

### DIFF
--- a/src/hyrax/data_sets/inference_dataset.py
+++ b/src/hyrax/data_sets/inference_dataset.py
@@ -75,7 +75,7 @@ class InferenceDataSet(HyraxDataset, Dataset):
     def ids(self) -> Generator[str]:
         """IDs of this dataset
 
-        Note: Not using HyraxDataset.ids() here because we need to be return the ids of whatever
+        Note: Not using HyraxDataset.ids() here because we need to return the ids of whatever
         dataset was used for inference, not the sequential index HyraxDataset.ids() gives.
 
         Returns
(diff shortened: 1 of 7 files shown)


src/hyrax/data_sets/inference_dataset.py
```diff
@@ -71,6 +75,9 @@ def shape(self):
     def ids(self) -> Generator[str]:
         """IDs of this dataset
 
+        Note: Not using HyraxDataset.ids() here because we need to be return the ids of whatever
```
drewoldag, line 78:
```suggestion
        Note: Not using HyraxDataset.ids() here because we need to return the ids of whatever
```

drewoldag (approved):
This all looks pretty reasonable. The only question I have is regarding the optional `metadata_table` for HyraxDataset. Is there a particular reason that you chose to make it an Astropy Table? I don't know enough about astropy table to know if there are clear benefits over numpy recarrays for instance.

mtauraso:
Main reason is that it looked easier to construct an astropy table than a rec array.

What I would ideally want here is something that would take a rec array, astropy table, pandas dataframe, etc... return appropriate errors, and extract column names and the appropriate number of values.

I feel like that is a bug and we should probably do that at some point ^

PR #266, opened 2025-04-01
Add logic so that CiFAR dataset works out of the box

drewoldag:
Adding some logic to check for tuples returned from dataset __getitem__ or __iter__.

Not sure if this is the best long term solution or not, but this fix keeps the existing example notebooks running for the time being.

### DIFF
--- a/src/hyrax/data_sets/data_set_registry.py
+++ b/src/hyrax/data_sets/data_set_registry.py
@@ -146,9 +146,11 @@ class HyraxDataset:
             Shape tuple of the tensor that will be returned from the dataset.
         """
         if hasattr(self, "__getitem__"):
-            return self[0].shape
+            data_sample = self[0]
+            return data_sample[0].shape if isinstance(data_sample, tuple) else data_sample.shape()
         elif hasattr(self, "__iter__"):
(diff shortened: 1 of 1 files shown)


mtauraso (approved):
Agree this is the right way to fix.

I think the model-specifies-its-own-data graphql style interface will allow us to pull this code eventually.

src/hyrax/data_sets/data_set_registry.py
```diff
@@ -146,9 +146,11 @@ def shape(self) -> tuple:
             Shape tuple of the tensor that will be returned from the dataset.
         """
         if hasattr(self, "__getitem__"):
-            return self[0].shape
+            data_sample = self[0]
+            return data_sample[0].shape if isinstance(data_sample, tuple) else data_sample.shape()
         elif hasattr(self, "__iter__"):
-            return next(iter(self)).shape
+            data_sample = next(iter(self))
+            return data_sample[0].shape if isinstance(data_sample, tuple) else data_sample.shape()
```
mtauraso, line 153: I didn't catch this on review, but there should be no `()` at the end of this line and the similar one above.

PR #268, opened 2025-04-04
Fixing InferenceDatasetWriter id type for strings of variable length

mtauraso:
Fix to the issue @drewoldag and @mi-dai ran into with string ids of variable length during inference.


drewoldag (approved):
Approved

PR #273, opened 2025-04-08
Iterable dataset support

mtauraso:
- Adds an example CIFAR iterable dataset
- Changes to pytorch_ignite.py to work around an ignite bug in iterable datasets https://github.com/pytorch/ignite/issues/3372
- New is_iterable() and is_map() interface on dataset base class to unify discernment logic
- Support for abstract base classes that derive from HyraxDataset not being themselves checked for required methods.

### DIFF
--- a/docs/external_libraries.rst
+++ b/docs/external_libraries.rst
@@ -75,12 +75,18 @@ items in the batch. This loss is logged to MLflow and tensorboard.
 Defining a dataset class
 ------------------------
 
-Dataset classes are written as subclasses of both ``hyrax.data_sets.HyraxDataset`` and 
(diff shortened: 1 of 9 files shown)


hombit (commented):
Looks good! It would be nice to provide some docs or tests for example usage. I'm leting @drewoldag to approve it, because I'm not competent here

drewoldag (approved):
This seems pretty reasonable to me. And indeed we'll need to get docs and tests, but this is looking fine.

mtauraso:
I've added some docs to the branch. I'm going to address the testing as a follow-up PR because I think the most natural way to test this is with an end-to-end testing setup, which is already in #283 to validate a major refactor.
After #283 is merged I'll add the new iterable datasets to end-to-end testing.

Issue #286, opened 2025-04-16
Infer slowness on 10 million inference run with HSC DataSets


aritraghsh09:
Currently, the following line 

`metadata = filter_catalog_table_dedup.loc["object_id", sorted_object_ids]`

in https://github.com/lincc-frameworks/hyrax/blob/main/src/hyrax/infer.py takes about 10 hours for infer to run .

I am currently running some tests on benchmarking possible fixes and will report back later in the day. But essentially a lot of the slowness is coming from astropy table v/s numpy rec arrays v/s pandas indexing/sort times. 

I will report back soon with benchmarks on possible fixes.

PR #287, opened 2025-04-16
Slow Metadata Creation Fix

aritraghsh09:
Closes https://github.com/lincc-frameworks/hyrax/issues/286

Background:
`metadata = filter_catalog_table_dedup.loc["object_id", sorted_object_ids]`
in https://github.com/lincc-frameworks/hyrax/blob/main/src/hyrax/infer.py used to take about 10 hours for 10 million objects.

This PR replaces the `.loc` based indexing with a quicker dictionary-based lookup approach. This change preserves the ordering of object_ids while reducing run-time by 50X. 

Also adds a couple of debug-level logs in order to catch future issues in the metadata preparation step.

### DIFF
--- a/src/hyrax/data_sets/hsc_data_set.py
+++ b/src/hyrax/data_sets/hsc_data_set.py
@@ -79,6 +79,7 @@ class HSCDataSet(HyraxDataset, Dataset):
 
         # Relies on self.filters_ref and self.filter_catalog_table which are both determined
         # inside _init_from_path()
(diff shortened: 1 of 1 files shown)


aritraghsh09:
@mtauraso Please check that the new logic preserves what you were trying to have in metadata

mtauraso (approved):
Yes, this appears to extract the same rows in the same order as the old code.

PR #288, opened 2025-04-17
First version of LSSTDataset

mtauraso:
This is so you can -- with a source install of hyrax on an RSP -- commit code to LSSTDataset and test it.


aritraghsh09 (approved):
This looks good and thanks for the readme

PR #289, opened 2025-04-17
Infer returns an InferenceDataSet object

mtauraso:
This is a mega-PR with two major pieces. Recommend reviewing via individual commits:
1) A wholesale move of the infer verb to the new class-based verb infrastructure
2) A fix to an ordering bug in infer, affecting any runs that set `config["infer"]["split"]`


drewoldag (approved):
Looking good to me

aritraghsh09:
Closes https://github.com/lincc-frameworks/hyrax/issues/292

PR #295, opened 2025-04-18
Moving vectordb insertion to new verb.

drewoldag:
## Moving vectordb insertion out of `infer` verb and into it's own special verb, `index`!

The change here creates a new verb, `index` that can be called from the CLI or in a notebook. It also introduces a few new default config parameters. 

When running from the cli, there are two possible flags that can be passed `--input-dir` and `--output-dir`. If those are not provided, index will default to using the latest `infer` directory, and will create a new `index` output directory. 

CLI usage looks like this:
``` 
>> hyrax index

>> hyrax index -i <path_to_infer_results>

>> hyrax index -o <new_or_existing_database_directory>

>> hyrax index -i <path_to_infer_results> -o <new_or_existing_database_directory>
```

The notebook interface looks familiar, and an example notebook is included.


src/hyrax/verbs/vdb_index.py
```diff
@@ -0,0 +1,127 @@
...
+            index_mask = inference_data_set.batch_index["batch_num"] == batch
+
+            # Get the ids of the data in this batch file
+            ids = inference_data_set.batch_index["id"][index_mask]
+
+            # Retrieve the vectors from the batch file using the ids. We use the
+            # ids here so that we only have to open one file to get the vectors.
+            vectors = inference_data_set._load_from_batch_file(batch, ids)
+
+            vector_db.insert(ids=list(ids), vectors=vectors["tensor"])
```
mtauraso, line 127: I'd rather this access pattern use `__getitem__` but I can see that isn't posssible the way `InferenceDataSet` and `InferenceDataSetWriter` are written today.

I think we should change `InferenceDataSet` so this sort of code can just use `__getitem__`. I also think that such a change would be more convenient if `InferenceDataSet` is returning dictionaries with the `object_id` from the original dataset, so you aren't having to track id's separately here.

src/hyrax/vector_dbs/chromadb_impl.py
```diff
@@ -214,6 +226,8 @@ def search_by_vector(self, vectors: list[np.ndarray], k: int = 1) -> dict[int, l
         if k < 1:
             raise ValueError("k must be greater than 0")
 
+        #! Should to check that the input `vectors` is list[np.arrays]
```
mtauraso, line 229: Should we check?

tests/hyrax/test_chromadb_impl.py
```diff
@@ -0,0 +1,279 @@
...
+        collections = chromadb_instance.chromadb_client.list_collections()
+
+        assert collections is not None
+        assert len(collections) == 1
+        assert collections[0].name == "shard_0"
+
+
+def test_insert():
+    """Ensure that we can insert IDs and vectors into the database"""
+    with tempfile.TemporaryDirectory() as temp_dir:
```
mtauraso, line 33: NIT: Any reason not to use the tempfile fixture?

tests/hyrax/test_chromadb_impl.py
```diff
@@ -0,0 +1,279 @@
...
+def test_search_by_id_many_shards():
+    """Test search_by_id retrieves nearest neighbor ids when there are many shards"""
+    with tempfile.TemporaryDirectory() as temp_dir:
+        chromadb_instance = ChromaDB({}, {"results_dir": temp_dir})
+        chromadb_instance.shard_size_limit = 5
+        chromadb_instance.min_shards_for_parallelization = 3
+        chromadb_instance.connect()
+        chromadb_instance.create()
+
+        def random_vector_generator(batch_size=1):
```
mtauraso, line 119: This seems ideal to be factored out as a fixture

tests/hyrax/test_chromadb_impl.py
```diff
@@ -0,0 +1,279 @@
...
+            """Create random vectors"""
+            while True:
+                batch = [np.random.rand(3) for _ in range(batch_size)]
+                yield batch
```
mtauraso, line 123: NIT: This seems ideal to be factored out as a fixture

tests/hyrax/test_chromadb_impl.py
```diff
@@ -0,0 +1,279 @@
...
+        chromadb_instance.create()
+
+        ids = ["id1", "id2"]
+        vectors = [np.array([1, 2, 3]), np.array([4, 5, 6])]
+        chromadb_instance.insert(ids, vectors)
+
+        # Search by single vector should return the id1 and id2 in that order
+        result = chromadb_instance.search_by_vector([np.array([1, 2, 3])], k=2)
+        assert len(result[0]) == 2
+        assert np.all(result[0] == ["id1", "id2"])
```
mtauraso, line 162: NIT: These could be written as individual tests under a common fixture with the searchable setup.

mtauraso (approved):
Looks pretty reasonable to me. There are some suggestions for cleanups, but they are minor and I think this can move forward regardless.

tests/hyrax/test_chromadb_impl.py
drewoldag: Ignorance - made the change to use tmp_path in these tests.

tests/hyrax/test_chromadb_impl.py
drewoldag: Done

src/hyrax/vector_dbs/chromadb_impl.py
drewoldag: Oops, no, in this case I don't think we need to check. The Chromadb API handles single or lists of np.arrays. I've updated the interface as I think this should be the standard approach taken for any vector db. I'll update the tests as well.

tests/hyrax/test_chromadb_impl.py
drewoldag: I tried to clean this up a bit, not sure if what I did aligns with your suggestions though.

src/hyrax/verbs/vdb_index.py
drewoldag: Yeah, truly. The approach taken here was intended to minimize the number of files that needed to be opened during the process of loading the vector db. But yes, I'm on board with a more elegant way of doing this work.

PR #296, opened 2025-04-18
Adding column config to visualizer

### LINKED ISSUE #282: Getting Started notebook is broken
Issue is required IDs in the Cifar dataset. Probably two actions should be taken to fix things:

1) We should make sure object_ids end up in the cifar dataset metadata so that visualizer can use them
2) We should move to configuration the required "ra"/"dec" fields in the visualizer so the getting started notebook will work as written.

More context on this slack thread: https://uw-dirac.slack.com/archives/C08F5FLEY5A/p1744404691778099

mtauraso:
- Getting started steps should work now out of the box
- You now must set config['visualize']['fields'] to a list of the metadata fields you'd like the visualizer to use
- End to end tests now also test the visualizer
- Classes deriving from HyraxDataset who do not specify an "object_id" column in their metadata will now have that column pre-filled with the contents of their .ids() member.
- Minor changes to HSCDataSet tests to enable end-to-end testing since metadata is normally disabled on HSCDataSet for unit testing

Closes #282

### DIFF
--- a/docs/pre_executed/mpr_demo.ipynb
+++ b/docs/pre_executed/mpr_demo.ipynb
@@ -13,7 +13,16 @@
    "cell_type": "code",
    "execution_count": 1,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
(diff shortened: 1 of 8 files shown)


aritraghsh09:
@drewoldag If you have time and could review this today, that would be great. It would probably be good to merge this before I start playing around more with the visualizer.

drewoldag (approved):
This looks fine to me. I like the addition of the default `object_id` in data_set_registry

Issue #291, opened 2025-04-18
Running umap with pre-caching leads to errors


aritraghsh09:
Running `umap` with `[data_set][preload_cache] = true` leads to the following error. 

```
Terminating: fork() called from a process already using GNU OpenMP, this is unsafe.
Terminating: fork() called from a process already using GNU OpenMP, this is unsafe.
Terminating: fork() called from a process already using GNU OpenMP, this is unsafe.
Terminating: fork() called from a process already using GNU OpenMP, this is unsafe.
Terminating: fork() called from a process already using GNU OpenMP, this is unsafe.
Terminating: fork() called from a process already using GNU OpenMP, this is unsafe.
Terminating: fork() called from a process already using GNU OpenMP, this is unsafe.
```

My intuition is that this is due to an interaction between the pre-load multithreading and the umap multithreading.

Issue #290, opened 2025-04-18
Running umap with `fit_sample_size = false` produces error


aritraghsh09:
**Bug report**

Running umap with `fit_sample_size = false` produces the following stack trace:- 

```{bash}
Traceback (most recent call last):
  File "/mmfs1/gscratch/dirac/aritrag/miniconda3/envs/hyrax/bin/hyrax", line 8, in <module>
    sys.exit(main())
  File "/mmfs1/gscratch/dirac/aritrag/repos/hyrax/src/hyrax_cli/main.py", line 57, in main
    retval = verb.run_cli(args)
  File "/mmfs1/gscratch/dirac/aritrag/repos/hyrax/src/hyrax/verbs/umap.py", line 44, in run_cli
    return self.run(input_dir=args.input_dir)
  File "/mmfs1/gscratch/dirac/aritrag/repos/hyrax/src/hyrax/verbs/umap.py", line 67, in run
    return self._run(input_dir)
  File "/mmfs1/gscratch/dirac/aritrag/repos/hyrax/src/hyrax/verbs/umap.py", line 94, in _run
    index_choices = rng.choice(np.arange(total_length), size=sample_size, replace=False)
  File "numpy/random/_generator.pyx", line 986, in numpy.random._generator.Generator.choice
TypeError: expected a sequence of integers or a single integer, got 'np.float64(3333.0)'
Exception in thread HSCDataSet-preload-tensor-cache:
Traceback (most recent call last):
  File "/mmfs1/gscratch/dirac/aritrag/miniconda3/envs/hyrax/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/mmfs1/gscratch/dirac/aritrag/miniconda3/envs/hyrax/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/mmfs1/gscratch/dirac/aritrag/repos/hyrax/src/hyrax/data_sets/fits_image_dataset.py", line 421, in _preload_tensor_cache
    for idx, (id, tensor) in enumerate(zip(self.ids(), tensors)):
  File "/mmfs1/gscratch/dirac/aritrag/repos/hyrax/src/hyrax/data_sets/fits_image_dataset.py", line 467, in _lazy_map_executor
    future = executor.submit(self._read_object_id.__func__, self, id)  # type: ignore[attr-defined]
[...]

PR #297, opened 2025-04-19
Umap bugfixes

aritraghsh09:
Closes https://github.com/lincc-frameworks/hyrax/issues/290
When `config_sample_size = False`, `sample_size` used to be a float before. Casting it as an `int` now to solve this bug

Closes https://github.com/lincc-frameworks/hyrax/issues/291
While using multiprocessing, we now use the 'spawn' context to safely create subprocesses after
OpenMP threads have been initialized by any other previous step (e.g., data loader)

### DIFF
--- a/src/hyrax/verbs/umap.py
+++ b/src/hyrax/verbs/umap.py
@@ -2,7 +2,6 @@ import logging
 import pickle
 import warnings
 from argparse import ArgumentParser, Namespace
-from multiprocessing import cpu_count
 from pathlib import Path
 from typing import Optional, Union
 
@@ -68,7 +67,7 @@ class Umap(Verb):
 
     def _run(self, input_dir: Optional[Union[Path, str]] = None):
         """See run()"""
-        from multiprocessing import Pool
+        import multiprocessing as mp
 
         import umap
         from tqdm.auto import tqdm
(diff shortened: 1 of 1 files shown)


src/hyrax/verbs/umap.py
```diff
@@ -111,7 +111,9 @@ def _run(self, input_dir: Optional[Union[Path, str]] = None):
         all_ids = np.array(list(inference_results.ids()))
 
         # Process pool to do all the transforms
-        with Pool(processes=cpu_count()) as pool:
+        # Use 'spawn' context to safely create subprocesses after
+        # OpenMP threads are initialized by data loader
+        with mp.get_context("spawn").Pool(processes=mp.cpu_count()) as pool:
```
mtauraso, line 116: Can you tell me more about why we want 'spawn' here? My very surface-level read of the mp docs says fork is faster, but that may not be the main consideration.

mtauraso (approved):
This should merge. I have a small question, but regardless of the answer this should land on main.

src/hyrax/verbs/umap.py
aritraghsh09: You're right -- 'fork' is typically faster for process startup! The problem is that code elsewhere (during data loading and UMAP itself) is relying on OpenMP for parallelism. When these libraries initialize OpenMP threads, using fork() after that was causing crashes due to unsafe thread state copying (https://github.com/lincc-frameworks/hyrax/issues/291) 

Switching to spawn avoids that by starting fresh interpreter processes, which don’t inherit the OpenMP state. I tested it out, and `spawn` was only slightly slower for my specific use case. 

If this ends up being major bottleneck, we can try to come up with a more elegant solution than just using `spawn` here. I added a comment in the code.

PR #298, opened 2025-04-22
Adding support for dict-based datasets.

mtauraso:
- HyraxCifarDataSet is converted as an example
- Changes to Hyrax are backwards compatible to datasets returning Tensors or tuples of tensors directly.
- Models can now write a to_tensor() static method which converts a dict from whatever dataset to an approprite tensor.
- Torch's batching unifies the first level of dictionary keys so batches are now of type dict with tensor-y values representing the entirety of the batch for each top level dict key. (This is less a change and more just what torch does. We now rely on it)
- Dataset classes now no longer must define a shape() method. the backend simply pulls a sample element and determines the shape when needed. This operation unfortunately now requires both a model and a dataset, but it does not occur very often in code.

### DIFF
--- a/src/hyrax/models/model_registry.py
+++ b/src/hyrax/models/model_registry.py
@@ -3,6 +3,7 @@ from pathlib import Path
 from typing import Any, cast
 
 import torch.nn as nn
+from torch import Tensor
 
 from hyrax.plugin_utils import get_or_load_class, update_registry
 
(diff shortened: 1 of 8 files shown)


mtauraso:
Notes: This needs documentation updates, and possibly a few tests. I think there are some end to end tests that manipulate order which could be added.

Ideally we'd make the interface change across all hyrax dataset/model classes for consistency, and only keep the fallback code around long enough for kbmod and other folks outside the source tree to update.

I'm putting it out for a read to see if this direction still seems like a good idea. I think switching makes us more compatible with data sources and models from the internet, in that it provides a path for dataset authors to make many types of data available, and for model authors to easily select data they care about for inference/training.

aritraghsh09 (commented):
I see that you wrote this is backward compatible, but I wanted to get a few clarifications: 
- Most of my current runs are with https://github.com/lincc-frameworks/hyrax/blob/main/src/hyrax/models/hsc_dcae.py
- Does this model also need a `to_tensor` now? 
   - If so, will my already-trained models fail to infer, or do any of the other steps?
   - If so, do you mind writing a `to_tensor` for that model and throwing it into this PR? 

If this majorly impacts our current crop of models that we already trained (and inferred) for the week of May 6, I would need to do some additional testing before this branch is merged.

mtauraso:
hsc_dae doesn't need a `to_tensor` now because:

1) When it's run with HSCDataSet, the dataset returns tensors anyway, and we avoid all the new code. This is true for inference as well as training.
2) When it's run with HyraxCifarDataSet, the [default to_tensor](https://github.com/lincc-frameworks/hyrax/pull/298/commits/8d9b630aa20372bd8c6d9201612b2d4561fb75eb#diff-d9c9c815f0218382f40c6ee34da8befc183eef1343a32e38e19965290c1350ffR136-R150) function will handle transforming the dict to a tensor.

The only situations where you would have to write a `to_tensor` are:
1) HSCDataSet/FitsFileDataSet is modified to use a dictionary structure, but doesn't use the key "image" for its tensor.
2) You want to use a huggingface dataset off the HF branch which has a pre-defined dictionary structure that Francois decided.

In any event the inputs and outputs to the model `forward()` function are exactly the same before and after the change. The failure mode for `infer` or `train` with this change is that they crash early in the first epoch because they can't perform the dict->tensor conversion.

src/hyrax/models/model_registry.py
```diff
@@ -91,14 +92,74 @@ def hyrax_model(cls):
...
+        kwargs.update({"shape": sample.shape})
+
         original_init(self, *args, **kwargs)
         self.criterion = self._criterion()
         self.optimizer = self._optimizer()
 
     cls.__init__ = wrapped_init
 
-    required_methods = ["train_step", "forward", "__init__"]
+    def default_to_tensor(data_dict):
```
drewoldag, line 136: Is the `default_to_tensor` required if we're going to throw a RuntimeError if the `to_tensor` isn't implemented? i.e. Will the `default_to_tensor` ever be used?

drewoldag (approved):
Looks pretty cool to me.

aritraghsh09 (approved):
Thank you for the clarification! No qualms about merging!

src/hyrax/models/model_registry.py
mtauraso: `default_to_tensor` will be used in situations where the model class doesn't define a `to_tensor`, but the dataset is emitting `dict` values from `__getitem__()`. Running with `HyraxCifarDataset` and any model class that isn't `HyraxAutoencoder` will result in `default_to_tensor` being run.

That being said, its possible for a programmer to write a model class like:
```
@hyrax_model
class MyModel(nn.module):
    def to_tensor(self, data_dict):
        ...
```
This *looks* correct at a glance *and* passes our required method checks (`hasattr(cls, "to_tensor") = True`), but is a serious implementation error.

We require `to_tensor` to be `@staticmethod` because we need to call it in situations where the model class may not have been initialized yet (such as in `.models.model_registry.wrapped_init()`).  

The runtime error on `to_tensor` after we conditionally attach is to catch a non-`@staticmethod` `to_tensor` specifically, and give the user a code sample showing the appropriate method signature.

The inclusion of `to_tensor` in the required method check just after is a safety check against us removing `default_to_tensor` at some future point, to remind anyone editing that the method is assumed to exist on all models after `@hyrax_model` is through with them.

Hopefully that makes the situation more clear? I'm not sure if I really answered your question.

PR #299, opened 2025-04-22
Reducing umap memory footprint

aritraghsh09:
Creates a `config["umap"]["reduce_memory_usage"]` flag that users can use to run the `umap` process only as a single-threaded operation, along with Python garbage collection and other memory-saving behaviour.

`["umap"]["reduce_memory_usage"]` is set to `False` by default, which replicates previous behaviour.  

To give a scale of numbers:- 
- with the flag set to `True`; 100k infer points can be fitted by a umap with ~250GBs of RAM
- Without the flag, this requirement balloons to more than 500GB

### DIFF
--- a/src/hyrax/hyrax_default_config.toml
+++ b/src/hyrax/hyrax_default_config.toml
@@ -226,6 +226,12 @@ inference_dir = false
 # Number of data points used to fit the umap transform.
 fit_sample_size = 1024
 
+# Save the fitted umap as a pickle file 
+save_fit_umap = true
+
+# Use multiprocessing during transforming to umap space (More memory intensive)
+parallel = false
+
 # Name of the umap implementation to use
 name = "umap.UMAP"
 
--- a/src/hyrax/verbs/umap.py
+++ b/src/hyrax/verbs/umap.py
@@ -1,4 +1,6 @@
+import gc
 import logging
+import os
 import pickle
 import warnings
 from argparse import ArgumentParser, Namespace
@@ -6,6 +8,7 @@ from pathlib import Path
 from typing import Optional, Union
 
 import numpy as np
+import psutil
 
 from .verb_registry import Verb, hyrax_verb
 
@@ -95,13 +98,22 @@ class Umap(Verb):
         # If the input to umap is not of the shape [samples,input_dims] we reshape the input accordingly
         data_sample = inference_results[index_choices].numpy().reshape((sample_size, -1))
 
+        self._log_memory_usage("Before fitting umap")
         logger.info("Fitting the UMAP")
         # Fit a single reducer on the sampled data
         self.reducer.fit(data_sample)
+        self._log_memory_usage("After fitting umap")
 
         # Save the reducer to our results directory
-        with open(results_dir / "umap.pickle", "wb") as f:
-            pickle.dump(self.reducer, f)
+        if self.config["umap"]["save_fit_umap"]:
+            logger.info("Saving fitted UMAP Reducer")
+            with open(results_dir / "umap.pickle", "wb") as f:
+                pickle.dump(self.reducer, f)
+
+        # Reclaim Memory
+        del data_sample
+        gc.collect()
+        self._log_memory_usage("After Garbage Collection")
 
         # Run all data through the reducer in batches, writing it out as we go.
         batch_size = self.config["data_loader"]["batch_size"]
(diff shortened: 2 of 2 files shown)


src/hyrax/verbs/umap.py
```diff
@@ -95,13 +99,25 @@ def _run(self, input_dir: Optional[Union[Path, str]] = None):
...
         # Save the reducer to our results directory
-        with open(results_dir / "umap.pickle", "wb") as f:
-            pickle.dump(self.reducer, f)
+        if not reduce_memory_usage:
+            logger.info("Saving fitted UMAP Reducer")
+            with open(results_dir / "umap.pickle", "wb") as f:
+                pickle.dump(self.reducer, f)
+
+        # Reclaim Memory
+        if reduce_memory_usage:
```
mtauraso, line 117: Does this have some downside? Why is it under the flag?

src/hyrax/verbs/umap.py
```diff
@@ -110,36 +126,49 @@ def _run(self, input_dir: Optional[Union[Path, str]] = None):
...
             )
+            for batch_indexes in np.array_split(all_indexes, num_batches)
+        )
 
-            # iterate over the mapped results to write out the umapped points
-            # imap returns results as they complete so writing should complete in parallel for large datasets
-            for batch_ids, transformed_batch in tqdm(
-                pool.imap(self._transform_batch, args),
+        if reduce_memory_usage:
+            # Sequential loop
```
mtauraso, line 143: Nit: each of these loops might read better as their own functions.

src/hyrax/verbs/umap.py
```diff
@@ -95,13 +99,25 @@ def _run(self, input_dir: Optional[Union[Path, str]] = None):
         # If the input to umap is not of the shape [samples,input_dims] we reshape the input accordingly
         data_sample = inference_results[index_choices].numpy().reshape((sample_size, -1))
 
+        if reduce_memory_usage:
```
mtauraso, line 102: These two log stanzas seem generally useful even if reduce_memory_usage is false

src/hyrax/verbs/umap.py
```diff
@@ -95,13 +99,25 @@ def _run(self, input_dir: Optional[Union[Path, str]] = None):
...
         # Fit a single reducer on the sampled data
         self.reducer.fit(data_sample)
+        if reduce_memory_usage:
+            self._log_memory_usage("After fitting umap")
 
         # Save the reducer to our results directory
-        with open(results_dir / "umap.pickle", "wb") as f:
-            pickle.dump(self.reducer, f)
+        if not reduce_memory_usage:
+            logger.info("Saving fitted UMAP Reducer")
```
mtauraso, line 112: Why are these lines not executed when reduce_memory_usage is true? Is the reducer too large to save?

src/hyrax/hyrax_default_config.toml
```diff
@@ -226,6 +226,9 @@ inference_dir = false
 # Number of data points used to fit the umap transform.
 fit_sample_size = 1024
 
+# Run with a reduced memory footprint
+reduce_memory_usage = false
```
mtauraso, line 230: Beyond preserving old behavior, why would the average user want "reduce_memory_usage" set to 'false' by default? 

I'm also thinking that this flag might be better named "parallel" with default of 'false' making the single-threaded umap the default.

src/hyrax/verbs/umap.py
aritraghsh09: Will make them default

src/hyrax/verbs/umap.py
aritraghsh09: Yes, for the larger runs, somehow this increases the memory being used on hyrax by quite a bit; although theoretically speaking, I don't necessarily understand why.

src/hyrax/verbs/umap.py
aritraghsh09: Will make this default behaviour

src/hyrax/hyrax_default_config.toml
aritraghsh09: I agree. Will make this `true` 

With multi-threading, the primary problem is the ballooning memory usage. But if not doing the pickle dump control; this flag can probably be named something like `parallel`

mtauraso: You might actually have 2 configs then, maybe `save_fit_umap` and `parallel`?

mtauraso (approved):
I feel like we're converging on a good implementation. Approving this so you can merge fast once its done.

PR #302, opened 2025-04-23
Incorporating 3D visualization infrastructure into main 

aritraghsh09:
The primary goal of this PR is to prevent me from doing further development in my own little isolated corner. 

It brings in the 3d visualization infrastructure into main without creating a verb or creating additional plumbing to integrate the 3d visualization piece more with hyrax. The verb creation and integration is now [an issue](https://github.com/lincc-frameworks/hyrax/issues/303) and I might need help with that. 

This commit includes:-

1. a Readme describing what needs to be done to boot up a server
2. code to convert umap data to json structure (needed by server) 
3. code to start up a server displaying a webpage hosting the 3d visualizer


aritraghsh09:
Bonus addition: I added `psutil` (which is used by umap memory monitoring) to our dependencies after one of the CI tests (ASV) failed. But that doesn't seem to have solved the issue. @mtauraso can you let me know what you think?

src/hyrax/3d_viz/umap_data.json
drewoldag, line 1: Did you mean to include this file in the PR? Would it be better to save it in zenodo or equivalent? I'm kinda nervous about including large files, or introducing LFS, to the repo.

I know there are limits that Github imposes regarding LFS usage, so if nothing else, we should review those and make sure that we aren't near them. Additionally, if these files aren't required for the code to work, it might make sense to store them elsewhere.

drewoldag (commented):
The code seems fine. I didn't go over the html or js script in detail. I do have concerns about including the .json files in the repo, and I would like to get to the bottom of that before giving this the stamp of approval.

src/hyrax/3d_viz/umap_data_100k.json
drewoldag, line 1: Same question here, did you mean to include this file in the PR? Could it be stored in zenodo?

mtauraso (approved):
Agree generally with Drew's comments.

I do think getting the code into main is important, its a good first step toward making a verb that can run firefly.

src/hyrax/3d_viz/umap_data_100k.json
aritraghsh09: Removed file from the PR

src/hyrax/3d_viz/umap_data.json
aritraghsh09: Removed file

PR #304, opened 2025-04-25
Patch boundary handling in LSSTDataset

mtauraso:
**TL;DR: We trust the lsst pipeline to give us the best patch and error when bounds of that patch are violated by config.**

The problem here is we need to make a cutout of a deep coadd based on the ra/dec of a source, and an arbitrary semi-height/semi-width in angular space. It is possible for the ra/dec and sh/sw extent to specify a box that is arbitrarily close and/or overlapping of any patch or tract boundary. We need to figure out generally correct behavior that will not be surprising to most scientists.

Because each patch's pixels don't line up with each other in patch overlap regions (thanks spherical geometry!), assembling a single cutout with pixels from multiple patch images requires us to reproject and somehow add pixels in some new projection space in order that the cutout pixel values be correct. We would need to become more enmeshed with the data pipeline to ensure correct behavior.  Avoiding this, we came up with the following three possible behaviors:

1. We choose a primary patch for the cutout using the center and the inner bounding box of the patch. There is an unambiguous way to do this in the LSST pipelines API, and that same codepath appears to be how the primary patch for objects near patch boundaries is determined by pipeline code that detects objects. We then impose that the sh/sw cannot leave the outer bounding box of the patch, only choose pixels from the primary patch, and simply error (with a helpful message) if the user-supplied points and extents don't allow this.
2. Like 1, but we infill with nan for regions outside the outer bounding box of the primary patch.
3. We grab all patches that have any overlap with our desired cutout regions, reproject and add everything.

I chose 1 for now, with a possible extension to 2 once there's more nan handling in hyrax, which is on our roadmap because solar-system science needs it anyway.

### DIFF
--- a/src/hyrax/data_sets/lsst_dataset.py
+++ b/src/hyrax/data_sets/lsst_dataset.py
@@ -8,11 +8,6 @@ from .data_set_registry import HyraxDataset
 class LSSTDataset(HyraxDataset, Dataset):
     """LSSTDataset: A dataset to access deep_coadd images from lsst pipelines
     via the butler. Must be run in an RSP.
-
-    Input catalog is hardcoded
-    Cutout code incorrect near the edges of a tract or patch
-
-    Batteries not included, use at your own risk etc.
     """
 
     # BANDS = ["u", "g", "r", "i", "z", "y"]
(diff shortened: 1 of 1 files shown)


aritraghsh09 (approved):
This looks good to merge! 

I have made a [new issue](https://github.com/lincc-frameworks/hyrax/issues/306) to push for the future improvements discussed in this PR

PR #305, opened 2025-04-26
Tests and column name flexibility for FitsImageDataSet

mtauraso:
- Added tests based off Tianqing's 'small_dataset_hscstars'
- "object_id", "filename", and "filter" columns now have configurable names
- Configurable names work in both FitsImageDataSet and HSCDataSet
- "filter" is no longer required in FitsImageDataSet. We fill it in For you with a warning if it's not there.


aritraghsh09 (approved):
These are very helpful features and this should merge; but I have a question:- earlier we were using mostly fake fits which were generated during tests; is there a reason wee are adding real fits files and catalogs now to the tests?

mtauraso:
Yes, there are codepaths the fake fits doesn't test, and those codepaths are more applicable to the user experience of the FitsImageDataSet class.

I also wanted to try real data, because I recall a few sort of random weird things coming up where just choices TQ made in downloading resulted in fits files that didn't work exactly right in HSCDataSet, so the idea here is to get some exposure to weird datasets during our testing.

Note the datasets I added have 10 images, the idea is to keep the data footprint very small, but still hit the important codepaths.

PR #309, opened 2025-05-04
Improved 3D Visualisation Toolkit

aritraghsh09:
1. UI/UX Improvements including a new Selection Mode button

2. Adds a Table that appears when points are selected.
   - Columns can be chosen by the user
   - Disappears when nothing is selected 

3. Adds an Image Panel:- 
   - Initially loads 6 images 
   - Then the user can hit a button to incrementally load more images 
   
  Note: The image file loading, etc. (which was a major effort to get working) is currently implemented on the client-side browser using JS primarily to make it easy to work with static websites (like the currently hosted solution on epyc). But we should move later to a more robust solution that uses astropy for server side processing. 

4. Adds options for the user to choose from a set of different colormaps


aritraghsh09:
Merging this PR without more review as this not tightly coupled with the rest of hyrax just yet

PR #312, opened 2025-05-05
Fix bug in `index` that dissociates id and vector when inserting

drewoldag:
This change fixes a bug that was causing the ids and vectors to lose association.


mtauraso (approved):
🚢 it

PR #313, opened 2025-05-05
Save insert-sorted in addition to id-sorted InferenceDataSet batch index files

drewoldag:
Save an "insert-sorted" batch index file alongside the current "id-sorted" batch index.

This will live alongside the current "batch_index.npy" for a while. Down the line, if/when we discover that having the index available in insert order is valuable, we'll have it, and we won't have to bother users to rerun their inference.


mtauraso (approved):
This is good future-proofing in case we want to use an insertion-order batch index in the future.

PR #316, opened 2025-05-06
Documentation updates 

drewoldag:
It's been a while since we've updated the docs - this won't be the last time...


mtauraso:
Merging these so the docs can go up.

PR #317, opened 2025-05-10
Reducing import time from ~2.7s to ~0.6s

mtauraso:
- Many top of file imports moved to functions
- Some type annotations removed
- Added benchmark test to measure our import time.


mtauraso:
The new benchmark has not shown up here yet: https://lincc-frameworks.github.io/hyrax/#/

I think I need to merge and let the nightly run before it will...

drewoldag (approved):
Looks good. Thanks for adding the start up benchmark.

PR #325, opened 2025-05-20
Visualisation Scaling & Cleaning

aritraghsh09:
- Implements a lognorm-_ish_ scaling in both 2D and 3D visualization tools to ensure faint features also pop out in our display frames
- Also gets rid of some debug-level code in the 2D visualization tool. Keeps some stuff which might be helpful later in case needed. 
- Updates the logic in 2D visualization so that we always attempt to display the UMAP irrespective of the value of the ``return_verb" argument


drewoldag (approved):
Seems reasonable enough. When I saw `!==` in the image-viewer file, I had to look twice...

PR #326, opened 2025-05-20
Polishing the vector db user experience

drewoldag:
- Cleaning up vector database UX.
  - Now it's `h.save_to_database()` instead of `h.index()`.
  - Connect to database without needing to know the database type with `h.database_connection`.
  - Ensure that a user cannot insert duplicate ids into chromadb.
- Added `database_connection` verb
- Renamed `index` to `save_to_database`
- Created a new example notebook.
- Updated tests
- Modified ConfigManager - made the private static method, `read_runtime_config` public so that we can quickly read a config from a result directory without having to invoke the entire ConfigManager machinery (which is really slow now :()

### DIFF
--- a/src/hyrax/verbs/database_connection.py
+++ b/src/hyrax/verbs/database_connection.py
@@ -63,7 +63,10 @@ class DatabaseConnection(Verb):
 
         vector_db_path = Path(vector_db_dir).resolve()
         if not vector_db_path.is_dir():
-            raise RuntimeError(f"Database directory {str(vector_db_path)} does not exist.")
+            raise RuntimeError(
+                f"Database directory {str(vector_db_path)} does not exist. \
+                    Have you run `hyrax.save_to_database(output_dir={vector_db_path})`?"
+            )
 
         # Get the flavor of database (i.e. Chroma, Qdrant, etc) from the config
         # file saved in `vector_db_path`. This ensures that we will use the correct
--- a/src/hyrax/verbs/database_connection.py
+++ b/src/hyrax/verbs/database_connection.py
@@ -1,0 +1,100 @@
+import logging
+from argparse import ArgumentParser, Namespace
+from pathlib import Path
+from typing import Optional, Union
+
+from .verb_registry import Verb, hyrax_verb
+
+logger = logging.getLogger(__name__)
+
+
+@hyrax_verb
+class DatabaseConnection(Verb):
+    """Verb to insert inference results into a vector database index for fast
+    similarity search."""
+
+    cli_name = "database_connection"
+    add_parser_kwargs = {}
+
+    @staticmethod
+    def setup_parser(parser: ArgumentParser):
+        """Stub of parser setup"""
+
+        parser.add_argument(
+            "-d",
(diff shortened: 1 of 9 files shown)


src/hyrax/verbs/save_to_database.py
```diff
@@ -66,15 +61,22 @@ def run(
...
+        from hyrax.config_utils import (
+            create_results_dir,
+            find_most_recent_results_dir,
+            log_runtime_config,
+        )
         from hyrax.data_sets.inference_dataset import InferenceDataSet
         from hyrax.vector_dbs.vector_db_factory import vector_db_factory
 
-        config = self.config
+        config = deepcopy(self.config)
```
mtauraso, line 76: why?

src/hyrax/verbs/database_connection.py
```diff
@@ -0,0 +1,100 @@
...
+        if database_dir is not None:
+            vector_db_dir = database_dir
+        elif config["vector_db"]["vector_db_dir"]:
+            vector_db_dir = config["vector_db"]["vector_db_dir"]
+        else:
+            vector_db_dir = find_most_recent_results_dir(config, "vector-db")
+
+        vector_db_path = Path(vector_db_dir).resolve()
+        if not vector_db_path.is_dir():
+            raise RuntimeError(f"Database directory {str(vector_db_path)} does not exist.")
```
mtauraso, line 66: Should we give a more helpful error in the case where someone runs `h.database_connection()` but has not run `h.save_to_database()` first?

src/hyrax/verbs/save_to_database.py
drewoldag: I stash a copy of the config in the directory where the database is created. I update the config like so: 
```
config["vector_db"]["infer_results_dir"] = str(inference_results_path)
config["vector_db"]["vector_db_dir"] = str(vector_db_dir)
```

If I don't make a deepcopy, then the hyrax-level config is modified with this information. I wanted to avoid doing that since it wouldn't be obvious to the user that it had happened.

drewoldag: Another question comes up though, that I didn't consider. Suppose a user runs `h.save_to_database(input_dir="/a", output_dir="/this/path")` and a copy of the config is stashed in the database directory. 

Then suppose the user wants to newly data to the database and calls `h.save_to_database(input_dir="/b", output_dir="/this/path")`. The original config file will be overwritten.

From the perspective of how the the config is used here, that would be fine. But it would lose the origin of the first batch of data that was provided.

src/hyrax/verbs/database_connection.py
drewoldag: Updated the wording here to prompt the user to run `save_to_database()`

src/hyrax/verbs/save_to_database.py
mtauraso: This makes sense. 

We need to think through how provenance would work in some of these cases.

It also seems like there's some trade off between provenance recording and ease of use.

PR #329, opened 2025-05-20
Basic NaN handling using a quantile-based system

mtauraso:
This is our first run at image based handling of NaN values.

The basic algorithm is that it looks at every image in the batch and replaces NaNs in the image with the 5th percentile value from that same image. The percentile is configurable, and we can add other methods over time.

This is intended as a precursor to having LSST visit images as inputs to hyrax models, but could be immediately usable by anyone with NaNs in their data.

### DIFF
--- a/docs/external_libraries.rst
+++ b/docs/external_libraries.rst
@@ -150,6 +150,30 @@ it's ``x`` input. See the example below, which stacks the g, r, and i fluxes int
 Note that ``to_tensor`` must be defined with ``@staticmethod`` as in the example. The function does not have
 access to the model's data members through the typical ``self`` argument in python.
 
+Another possible use of ``to_tensor`` is to remove or otherwise adjust the input data of your model in ways 
+that are not easily done with a ``torch.transform``. Below is an example ``to_tensor`` function which removes 
+NaN values from input data, replacing them with the value zero. 
+
+.. code-block:: python
+
+    from hyrax.models import hyrax_model
+
+    @hyrax_model
+    class MyModel:
+
+        @staticmethod
+        def to_tensor(batch_dict):
+            """
+            Accepts a batch of tensors which may contain NaN values. Replaces those values with zero.
+            """
+            from torch import any, isnan, nan_to_num
+            if any(isnan(batch)):
+                batch = nan_to_num(batch, 0.0)
+            return batch
+
+Some NaN handling is available automatically in hyrax, via ``config['data_set']['nan_mode']``, but if a 
+customized strategy is desired, the approach above may be preferable.
+
 .. _custom-dataset-instructions:
 
 Defining a dataset class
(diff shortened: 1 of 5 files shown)


drewoldag (approved):
This looks fine for now. I'm excited about this turning into a whole group of different nan-handlers with a nan-handler registry :) Soon Hyrax will be all registries and factories - bwahahaha!

aritraghsh09 (commented):
This looks good to me in general; just a couple of comments:-

1. It doesn't seem like this will add any overhead at all when `["data_set"]["nan_mode"]` is set to `False`. But, do our automated tests indicate anything? 
2. @mtauraso How do you feel about setting the default value of `nan_mode` to `False`? Imputing by quantile is a non-trivial decision, and I think this should be an active choice by users and not default behaviour. If you do decide to keep this as default behaviour, I think there should at least be an info-level logger message; warning users that qauntile based imputation is being done.

mtauraso:
@aritraghsh09 
Agree that NaN handling is subtle, users should probably know what they are doing. We could definitely add some info logging to try to direct people in the right direction. I'm honestly not sure what the right default is for `nan_mode`, I could see it being `False`.

For perf, we don't currently have tests benchmarking this particular codepath, so I'm not sure of the performance impact. I expect as you do that with `nan_mode` set to `False` the impact should be negligible.

mtauraso:
I'm going to go ahead and make `nan_mode = false` by default and add some logging to direct people toward the config. This is probably the safest decision right now, though perhaps not the boldest. Whether this is the moment to be bold on this issue, I have no clue.

Thinking about this decision with respect to our users, I think it hinges primarily on how many people we expect to:
1) Have NaNs in their dataset AND 
2) Care primarily that NaNs are handled over precisely *how* they are handled.
* To the extent this group is smaller we should prefer `nan_mode=False`, preferring to treat them as a niche who must learn some config to get things working the first time.
* To the extent this group is larger we should prefer `nan_mode='quantile'` Or some other generally-good default, preferring to treat this group as "handled by defaults" and treat the user group that cares about the *how* as the niche that must learn more to use the software.

The size of this group depends on two big things: Our user base (nascent/nonexistant at the moment), and how many of our NaN-having users a single default algorithm can cover. We have some control over the latter through what algorithms we offer and make available, but a single default may also be intractable. 

I'd be very interested in information about what astronomy researchers *actually do* with NaNs in their machine learning projects, and *how* they decide what to do. Some open-ended questions:

Are NaNs a hurdle where researchers grab from a small handful of off-the-shelf algos? 
Do researchers write mostly the same 5 algos themselves because there isn't a library? 
Do people mostly write custom NaN handling that is germane to their project? 
How critical do people believe their custom handling routines/algorithm choice are to their results? 
Is the criticality primarily because their ML has non-NaN data, more because of the *particular* algorithim they chose, or something else entirely?

mtauraso:
Now that I've added nan_mode=False by default, I'm going to merge.

PR #330, opened 2025-05-21
Read the original config file from results directory for InferenceDataset

drewoldag:
Just read the config in the results directory instead of hydrating the entire config.

Probably not a huge time savings, but I don't think it's necessary to rehydrate the config file from scratch in this case, since I believe all we want to do is use some values from it to properly instantiate the dataset.


mtauraso (approved):
Seems fine to me. When I wrote this line, I think I was worried about violating encapsulation, but no specific concerns here.

PR #331, opened 2025-05-26
New Features for LSST X Hyrax

aritraghsh09:
1. LSSTDataset: Updates the LSSTDataset to add support for Astropy Tables (i.e., any file format that is supported by `astropy`) 

2. Updates the demo notebook to show the above feature

3. Introduces a new dataset called `DownloadedLSSTDataset` that inherits from `LSSTDataset`
     - This dataset downloads tensors and saves them as `.pt` files for rapid access
     - Uses the same LRU caching strategy as LSSTDataset; but multi-threads it 
     - Creates and updates a manifest file during download, similar to `HSCDataset` 
     - Provides some convenience functions to monitor / help during donwload. 
     

Notes:- 
- Tested that this works on the RSP
- @mtauraso & @drewoldag : If either you think this code will benefit from refactoring or other major structural changes; please feel free to push directly to this branch/PR.

### DIFF
--- a/docs/pre_executed/hyrax_hats_cutouts.ipynb
+++ b/docs/pre_executed/hyrax_hats_cutouts.ipynb
@@ -46,11 +46,11 @@
    "id": "ab7c0163-5b56-47dd-a19f-c1e5604c0ac2",
    "metadata": {
     "execution": {
-     "iopub.execute_input": "2025-04-23T00:16:50.912226Z",
-     "iopub.status.busy": "2025-04-23T00:16:50.911905Z",
-     "iopub.status.idle": "2025-04-23T00:16:50.915794Z",
-     "shell.execute_reply": "2025-04-23T00:16:50.915367Z",
-     "shell.execute_reply.started": "2025-04-23T00:16:50.912206Z"
+     "iopub.execute_input": "2025-05-26T07:50:34.665977Z",
+     "iopub.status.busy": "2025-05-26T07:50:34.665377Z",
+     "iopub.status.idle": "2025-05-26T07:50:34.668812Z",
+     "shell.execute_reply": "2025-05-26T07:50:34.668333Z",
+     "shell.execute_reply.started": "2025-05-26T07:50:34.665958Z"
     }
    },
    "outputs": [],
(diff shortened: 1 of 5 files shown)


src/hyrax/data_sets/lsst_dataset.py
```diff
@@ -17,39 +18,100 @@ def __init__(self, config):
...
+        """Load catalog from astropy table format or pickled astropy table."""
+
+        import pickle
+
+        from astropy.table import Table
+
+        table_path = Path(table_path)
+
+        # Check if it's a pickle file
+        if table_path.suffix.lower() in [".pkl", ".pickle"]:
```
mtauraso, line 81: Do we want to support the pickle file format? 

My understanding is that is unstable for a lot of uses, but not sure if any of that applies here.

src/hyrax/data_sets/downloaded_lsst_dataset.py
```diff
@@ -0,0 +1,616 @@
...
+        with self._manifest_lock:
+            self._save_manifest()
+            self._updates_since_save = 0
+        logger.info("Manifest manually saved")
+
+    @staticmethod
+    def _determine_numprocs_download():
+        """Determine number of threads for downloading."""
+        # TODO:This is a placeholder for actual logic to determine number of threads.
+        return 1
```
mtauraso, line 575: This could be a config?

src/hyrax/data_sets/downloaded_lsst_dataset.py
drewoldag, line 1: Going to leave a similar comment here as I did on `lsst_dataset`. There's lots of if/else logic sprinkled throughout this class to handle either `AstropyTables` or `HATS catalogs`. At this point it's fine but I sense that in the future we'll refactor this into multiple classes that the different data formats independently.

drewoldag (approved):
Overall this is fine. I've left a couple comments expressing structural concern, but I accept that addressing those comments right now is probably out of scope, and it's more beneficial to think about them more later

src/hyrax/data_sets/lsst_dataset.py
aritraghsh09: For larger catalogs, I/O times (especially writing) for Astropy `.fits` tables are very bad. Pickles have been at least 10X quicker in my testing, and that's why I have been using them for my work.

That's the reason I am adding support here. It works fine in my testing. But if it leads to problems later on downstream, we can get rid of this.

docs/pre_executed/hyrax_hats_cutouts.ipynb
```diff
@@ -46,11 +46,11 @@
    "id": "ab7c0163-5b56-47dd-a19f-c1e5604c0ac2",
```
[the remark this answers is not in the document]
aritraghsh09: See reply to previous comment about astropy table

src/hyrax/data_sets/downloaded_lsst_dataset.py
aritraghsh09: This is already an argument (`max_workers`) for `download_cutouts`; and this is only a placeholder backup similar to the `HSCDataSet` multi-threading placeholder. 

The plan was to develop the code later on so that this configures on its own (due to the weird non-linear behaviour of download speed v/s number of threads); but if we don't reach that point; we can probably turn it into a config.

src/hyrax/data_sets/downloaded_lsst_dataset.py
```diff
@@ -0,0 +1,616 @@
...
+                    # Manifest says file exists but it doesn't, reset entry
+                    self._update_manifest_entry(idx, None, "")
+                    synced_count += 1
+
+        if synced_count > 0:
+            logger.info(f"Synced {synced_count} manifest entries with filesystem")
+            self.save_manifest_now()
+
+    @staticmethod
+    @functools.lru_cache(maxsize=128)
```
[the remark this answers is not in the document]
aritraghsh09: So, here's how the caching is working here and it should address the primary concern:- 

1. `_request_patch_cached` , because of being a `@staticmethod`, should be attached to the class-level method and not any specific instance of it. 
2. Every thread (whether main thread or worker thread) should call `DownloadedLSSTDataset._request_patch_cached()` - the same static method with the same cache. As far as I can understand,  `functools.lru_cache` is thread-safe by design and uses internal locks to handle concurrent access. 
3. Given this, the cache should save the last 128 patches and should be hit across from all the threads. 

In my current testing, if the number of workers is selected appropriately, this cache is hit 95% of the time. If as I scale up, this stops being the case, I will push some more changes in line with the patch-sorting logic we discussed in June 2, 2025's Hyrax Sync meeting.

src/hyrax/data_sets/downloaded_lsst_dataset.py
```diff
@@ -0,0 +1,616 @@
...
+        # Retry failed downloads
+        a.download_cutouts(force_retry=True)
+
+        # Access cutouts (loads from cache)
+        cutout = a[0]  # Single cutout
+        cutouts = a[0:10]  # Multiple cutouts
+
+    File Organization:
+    - Cutouts saved as: cutout_{object_id}.pt or cutout_{index:04d}.pt
+    - Manifest saved as: manifest.fits (Astropy) or manifest.parquet (HATS)
```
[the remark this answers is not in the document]
aritraghsh09: This is currently not configurable by the user; you get back what you pass. If you passed `.fits` (either native or pickled) originally, you get back `.fits`; if you give Hyrax a HATS catalog, you get back parquet.

src/hyrax/data_sets/downloaded_lsst_dataset.py
aritraghsh09: This is an issue now: https://github.com/lincc-frameworks/hyrax/issues/336

src/hyrax/data_sets/lsst_dataset.py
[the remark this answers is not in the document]
aritraghsh09: This is an issue now https://github.com/lincc-frameworks/hyrax/issues/336

PR #333, opened 2025-05-30
Flatten vectors and check for tuples

drewoldag:
Flattening vectors before inserting into vector db, checking if batch is tuple before de-NaN-ifying.

- The verb `save_to_database` was not flattening the vectors before attempting to insert them into the database, now it does.

- Additionally, in the `_handle_nan` function, if a batch is provided that is a tuple, the `torch.any(torch.isnan(batch))` logic would fail, complaining that the batch wasn't a tensor. I added a little type check-and-convert with the assumption that getting to a working solution is better for now, even if it's not the idealized solution.


mtauraso (approved):
This should go in.

Is there an easy way to write a test for input/output format to database which would catch the behavior change here?

PR #339, opened 2025-06-04
Fixing bug in `_handle_nans` treatment of tuple inputs

drewoldag:
Implementing single dispatch for nan-handling. Might be over doing it but it makes nice separation. Added and updated tests.

The root cause was that I was overriding the value of `batch` with `batch[0]` when the input type of `batch` was a tuple. Thus, when batch was returned, it had lost all the other elements of the tuple. 

This is fixed now. The singledispatch approach allows slightly cleaner treatment of logic necessary. This is mostly to address the case of recombining the values from `_handle_nan_quantile` and any future nan handling approaches taken. Really I just didn't want to have piles of `if isinstance(batch, tuple)` all over the place. 

Also added a test to make sure this is working as expected for our dataset types, and updated the HyraxLoopback model to handle tuples appropriately. (It's starting to make me wonder more if HyraxLoopback should be a real model just to help people get their pipelines set up - @mtauraso originally pitched that idea IIRC)

### DIFF
--- a/tests/hyrax/test_nan.py
+++ b/tests/hyrax/test_nan.py
@@ -5,6 +5,7 @@ from torch.utils.data import Dataset
 
 import hyrax
 from hyrax.data_sets import HyraxDataset
+from hyrax.pytorch_ignite import _handle_nans
 
 
 class RandomNaNDataset(HyraxDataset, Dataset):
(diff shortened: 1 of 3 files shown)


tests/hyrax/test_nan.py
```diff
@@ -123,3 +124,32 @@ def test_nan_handling_off(loopback_hyrax_nan):
 
     result_nans = tensor([any(isnan(item)) for item in inference_results])
     assert any(result_nans)
+
+
+def test_non_handling_off_returns_input(loopback_hyrax_nan):
```
mtauraso, line 129:
```suggestion
def test_nan_handling_off_returns_input(loopback_hyrax_nan):
```

mtauraso (approved):
LGTM, minor typo in suggest

PR #340, opened 2025-06-05
Adding chromadb shard size and vector size configs

drewoldag:
Adding configuration for shard size and warning for large vectors to chromadb_impl.

The user now has easier control over the shard size of the chromadb vector database. Additionally, a warning is emitted if the user is trying to insert vectors with >= 10k elements. And the 10k value is configurable in the config as well so they can easily turn off the warnings by setting this to false.


mtauraso (approved):
LGTM

PR #342, opened 2025-06-06
Adding HyraxLoopback and HyraxRandomDataset classes.

drewoldag:
This PR promotes the random datasets and the loopback model that were used for testing to fully supported datasets and model. 

The goal here is to provide users will simple place holders. If they are designing a model and waiting for data to download, they can use the random dataset to ensure that their model handles the expected shape and aspects of their data. If they are developing a new dataset, they can use the loopback model as a very fast model for debugging. 

For the most part the tests have been updated to make use of the new random datasets and loopback model. 

- [x] Break apart HyraxRandomDataset so that iterable version has correct behavior around len(dataset).
- [x] Documentation out the nose.
- [x] Double check on behavior of the loopback model and simplify it if possible
- [x] Reduce the pytest conftest.py file and use the built in model and dataset classes in the tests


mtauraso (approved):
LGTM.

PR #343, opened 2025-06-11
Better documentation for verbs that return a dataset object

mtauraso:
Coming out of a conversation with Peter, Verbs returning a dataset object and the object itself aren't well linked or fleshed out in our docs. This is an attempt to remedy the situation.


psferguson:
I like it!

PR #347, opened 2025-06-23
More benchmarks for vector db - search

drewoldag:
Fixed a benchmark for `index` timing that I should have updated when index was renamed to `save_to_database`. Added a timing benchmark for `database_connection`.


mtauraso (approved):
🚢 it

PR #359, opened 2025-07-07
Train returns a model

mtauraso:
- Moved train to new verb infrastructure
- Train now returns the trained model object for use in notebook
- Example notebook covering direct examination of model, onnx, and pytorch weights files
- Fixed: Tests won't fail on OSX's default  NOFILE rlimit of 256
         files per processif the test process can change the value
         itself.
- Fixed: User working directory configs (./hyrax_config.toml) no longer
         read when tests are being run.


aritraghsh09 (approved):
Looks good to me!

PR #361, opened 2025-07-08
Updating landing page with more structure and new info

aritraghsh09:
Updating landing page with:-

1. More guidance on the structure of documentation
2. New header image with new mascot
3. Slightly updated language


drewoldag (approved):
This looks ok to me. I'm slightly apprehensive about the tag line "A framework for scalable ML and unsupervised discovery in astronomy". Specifically calling out "unsupervised discovery". Should be fine for now, perhaps it gives a concrete hook for users to grab onto to help get people started.

PR #362, opened 2025-07-09
HyraxQL implementation

drewoldag:
# HyraxQL

I wanted to put this together to see how some of the pieces would fit together. This PR should be interpreted as a potential implementation/techspec/starting point. 

### Why I implemented graphql instead of using something out of the box
There were two main contenders, graphql-core and graphene. Graphene is a wrapper over graphql-core (because gql-core was so challenging to use they needed a separate package to make it usable) Graphene was ok, but required the user to write the data request as a json string. We could implement this as a dictionary that then gets passed through json.stringify (or whatever the function is called) 

There was a lot of functionality that Graphene provided that was unnecessary for our use case (like modifying the original data, as one would do when modifying data in a db as part of an API request)

Finally if we roll our own graphql resolver, we can make it _look_ like a pytorch dataset, and so we can continue to use all of the machinery that we've already built, and the machinery provided by pytorch ignite that puts the data loader on the best available hardware. 

So, while I love graphql, the python implementations seemed like they aren't exactly what we need, and the functionality that we do need is accomplished in a relatively small class. 

- [ ] Write inferred `model_data` back to the configuration file so that it is persisted with the rest of the configuration
- [ ] Determine if `InfereceDataset.__getitem__` really needs to be returning a PyTorch tensor, of if numpy is fine.


src/hyrax/data_sets/fits_image_dataset.py
```diff
@@ -333,6 +335,40 @@ def __len__(self) -> int:
...
+
+        Returns
+        -------
+        torch.Tensor
+            The image at the given index as a PyTorch Tensor.
+        """
+        object_id = self.get_object_id(idx)
+        return self._object_id_to_tensor(object_id)
+
     def __getitem__(self, idx: int):
```
drewoldag, line 372: Shall we update this to be:
```
def __getitem(self, idx: int) -> dict:
    return {
        'data': {
            "object_id": self.get_object_id(idx),
            "image": self.get_image(idx),
            "index": idx
        }
    }
```

src/hyrax/models/model_registry.py
```diff
@@ -92,53 +92,23 @@ def hyrax_model(cls):
...
         self.criterion = self._criterion()
         self.optimizer = self._optimizer()
 
     cls.__init__ = wrapped_init
 
     def default_to_tensor(data_dict):
-        if isinstance(data_dict.get("image"), Tensor):
-            if "label" in data_dict:
-                return (data_dict["image"], data_dict["label"])
+        data = data_dict.get("data")
```
drewoldag, line 103: Here we explicitly assume that the default dict returned from any dataset class `__getitem__` will be of the form:
```
{"data": {<field_n>: <value_n>}}
```

src/hyrax/pytorch_ignite.py
```diff
@@ -50,6 +52,59 @@ def __len__(self) -> int:
...
 
 
+def is_iterable_dataset_requested(data_request: dict) -> bool:
+    """This function checks each of the datasets included in the data_request.
+    If any of them are iterable-style datasets, we return True.
+    """
+
+    is_iterable = False
+    for _, dataset_definition in data_request.items():
+        if DATA_SET_REGISTRY[dataset_definition["dataset_class"]].is_iterable():
```
drewoldag, line 62: Note: using the `classmethod` version of `is_iterable()` here as well.

src/hyrax/verbs/infer.py
```diff
@@ -58,11 +57,9 @@ def run(self):
         # Create a tensorboardX logger
         tensorboardx_logger = SummaryWriter(log_dir=results_dir)
 
-        data_set = setup_dataset(config, tensorboardx_logger)
-
-        model = setup_model(config, data_set)
-        if data_set.is_map():
-            logger.info(f"data set has length {len(data_set)}")  # type: ignore[arg-type]
+        model, dataset = setup_model(config, tensorboardx_logger)
```
drewoldag, line 60: Revert this change. Also, update the logger.info statement to use capitalization.

src/hyrax/verbs/train.py
```diff
@@ -55,11 +54,10 @@ def run(self):
         tensorboardx_logger = SummaryWriter(log_dir=results_dir)
 
         # Instantiate the model and dataset
-        data_set = setup_dataset(config, tensorboardx_logger)
-        model = setup_model(config, data_set)
+        model, dataset = setup_model(config, tensorboardx_logger)
```
drewoldag, line 57: Just revert this change after reverting the stuff in `pytorch_ignite.py`.

tests/hyrax/test_inference_dataset.py
drewoldag, line 1: @mtauraso There are a bunch of changes in this file, so I would appreciate a close look when it's time to review this PR to make sure that I haven't broken InferenceDataset in a fundamental way.

src/hyrax/data_sets/fits_image_dataset.py
drewoldag: I went ahead and did made this change.

aritraghsh09 (commented):
I think we have reviewed this enough through meetings and conversations that I am happy with these chnages. But I will leave it to @mtauraso to do a closer look!

src/hyrax/data_sets/hsc_data_set.py
```diff
@@ -25,7 +25,7 @@ class HSCDataSet(FitsImageDataSet):
 
     _called_from_test = False
 
-    def __init__(self, config: ConfigDict):
+    def __init__(self, config: ConfigDict, data_location=None):
```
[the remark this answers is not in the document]
drewoldag: This is a sensible argument, however, our user base is small enough at this point that we can affort to roll forward and address backward compatibility issues as they arise with our users.

src/hyrax/data_sets/data_set_registry.py
```diff
@@ -100,12 +101,26 @@ def __init__(config):
...
+            for col in self._metadata_table.colnames:
+                method_name = f"get_{col}"
+                if not hasattr(self, method_name):
+                    setattr(self, method_name, MethodType(_make_getter(col), self))
+
         self.tensorboardx_logger = None
 
-    def is_iterable(self):
+    @classmethod
+    def is_iterable(cls):
```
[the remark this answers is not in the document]
drewoldag: These need to be class methods so that we can check them before creating an instance of a given dataset class.

src/hyrax/data_sets/data_provider.py
```diff
@@ -0,0 +1,532 @@
...
+        Parameters
+        ----------
+        config : dict
+            The Hyrax configuration that defines the data_request.
+        """
+
+        self.config = config
+        self.data_request = generate_data_request_from_config(self.config)
+
+        self.validate_request(self.data_request)
```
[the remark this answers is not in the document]
drewoldag: `validate_request` will ultimately be removed. So we can skip the try/except block here.

PR #365, opened 2025-07-10
Fixed logging for fits_image_dataset

diegomiura:
## Change Description

- [ ] My PR includes a link to the issue that I am addressing

## Solution Description

As you can see below, `FitsImageDataSet` has 5 objects, but it would say that it only processed 4.

Screenshot 2025-07-10 at 12 35 45 PM

The logging would print the {index} amount of objects, which start from 0 and is always 1 off from the true number of objects processed. Now it is {index+1}.


drewoldag:
This PR is looking just fine, but it's failing the pre-commit checks - https://github.com/lincc-frameworks/hyrax/actions/runs/16207839285/job/45762022627?pr=365 

Ruff is trying to enforce a particular code style. In this case, it's just wanting spaces around the operators :)

drewoldag (approved):
Looks great, thanks!

PR #366, opened 2025-07-12
Update pytorch_ignite.py to track any metric returned by `train_step`

drewoldag:
Currently the create_trainer and create_validator functions expect that the `train_step` method in the model class will return a "loss" value. 

The user should be free to return and track any metrics produced in the `train_step` method. This change enables that. 

As long as the returned value from `train_step` is a dictionary, the metrics will be recorded in TensorBoard and MLFlow. 

For reference the return dict should look something like this: 

```
return {
    'metric_0': metric_0_value,
    ...,
    'metric_n': metric_n_value,
}
```


mtauraso (approved):
LGTM

PR #368, opened 2025-07-15
New Models with Slight Modifications in config file

aritraghsh09:
- Updates some config variables for better consistency across our models
- Renames HSCDCAEv2 to ImageDCAE -- a new more general model with a proper latent space bottleneck
- Introduces a v2 for the HyraxAutoencoder with more flexibility than our baseline version


src/hyrax/models/hyrax_autoencoderv2.py
```diff
@@ -0,0 +1,196 @@
...
+    def _eval_decoder(self, x):
+        x = self.dec_linear(x)
+        x = x.reshape(x.shape[0], -1, self.conv_end_h, self.conv_end_w)
+        x = self.decoder(x)
+        x = CenterCrop(size=(self.image_width, self.image_height))(x)
+        return x
+
+    def forward(self, batch):
+        # When we run on a supervised dataset like CIFAR10, drop the labels given by the data loader
+        x = batch[0] if isinstance(batch, tuple) else batch
```
drewoldag, line 128: Should we try to simplify this, and modify the `to_tensor` method so that it will return just the image instead of the `(image, label)` tuple? 

If we made the change to `to_tensor` then we could remove the logic on line 146 too.

src/hyrax/models/image_dcae.py
```diff
@@ -0,0 +1,233 @@
...
+            x = f.interpolate(x, size=self.original_size, mode="bilinear", align_corners=False)
+
+        x = self.final_activation(self.decoder1(x))
+
+        return x
+
+    def forward(self, x):
+        """Forward pass - returns latent representation for anomaly detection."""
+        # Dropping labels if present
+        x = x[0] if isinstance(x, tuple) else x
```
drewoldag, line 164: Since `to_tensor` is returning just the image, you can drop this logic. Likewise, on line 204 in `train_step`.

drewoldag (approved):
Looks pretty good. Left a couple comments, but nothing major, just a little bit of clean up.

src/hyrax/hyrax_default_config.toml
```diff
@@ -79,22 +79,29 @@ mask = false
...
+# The activation function of the final layer.
+final_layer = "tanh"
 
 
 [criterion]
 # The name of the built-in criterion to use or the import path to an external criterion
 name = "torch.nn.CrossEntropyLoss"
 
+# Whether to "sum" or "mean" loss across channels. Only used by HyraxAutoencoderV2
+band_loss_reduction = "mean"
```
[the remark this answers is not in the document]
aritraghsh09: I will be using this for the entire family of HSC and LSST models that we train. But we should keep an eye on this; and revisit this at some point above. For configs that are only used by one model; they should live under the [model.model_name] universe

src/hyrax/models/image_dcae.py
aritraghsh09: Updated!

PR #383, opened 2025-08-12
Read inference_dir as `None` if it's `false` in the config

drewoldag:
The run method of visualize was taking the `inference_dir` value directly from the config (it defaults to `False`) and passing that to instantiate the InferenceDataset. 

InferenceDataset expects a Path or str (or None) as the data types for `result_dir`. 

The change here will set `input_dir` to `None` if it is `false` in the config.


mtauraso (approved):
🚢 it

PR #388, opened 2025-08-18
WIP - Investigating use of parquet as output format for InferenceDataset

drewoldag:
So far it seems pretty good. And seems like it will reduce a lot of code. Making use of `pyarrow` right now. Tried out `fastparquet` but it was having trouble accepting numpy arrays as input. 

The biggest hurdle that I see right now is that we have to flatten and reshape the numpy arrays. They can be at most 1D. What is currently implemented uses parquet metadata to store the original shape of the input tensor (assuming that all tensors were the same dimensions originally) in `InferenceDatasetWriter`. 

When reading the data in with `InferenceDataset`, we reshape the tensors before returning them. 

Some open questions are:
- Should we embed any existing metadata from the original dataset along side? 
- Per-column metadata is supported, so perhaps the original shape information should be on that column instead of the global metadata
- Is this an opportunity to rename the "tensor" column to something like model_output, since it's not a tensor in the PyTorch sense... 
- Not a question just a statement, HyraxQL is going to fit so nicely along side this.

### DIFF
--- a/docs/pre_executed/random_parquet.ipynb
+++ b/docs/pre_executed/random_parquet.ipynb
@@ -49,6 +49,16 @@
   {
    "cell_type": "code",
    "execution_count": null,
+   "id": "845aa3da",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "h.save_to_database()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
    "id": "d7f79cb1",
    "metadata": {},
    "outputs": [],
--- a/docs/pre_executed/random_parquet.ipynb
+++ b/docs/pre_executed/random_parquet.ipynb
@@ -1,0 +1,363 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "a72367cd",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "[2025-08-18 16:26:09,122 hyrax:INFO] Runtime Config read from: /Users/drew/code/hyrax/src/hyrax/hyrax_default_config.toml\n"
+     ]
+    }
+   ],
+   "source": [
+    "from hyrax import Hyrax\n",
+    "h = Hyrax()\n",
+    "# h.config['data_set']['name'] = 'HyraxRandomDataset'\n",
+    "# h.config['data_set']['random_dataset']['size'] = 10_000\n",
+    "# h.config['data_set']['random_dataset']['shape'] = (3, 41, 41)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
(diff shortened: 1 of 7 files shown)


mtauraso:
Thoughts:
* I like @hombit 's suggestion in slack to use proper tensor types, in that it'll hopefully help people who take their inference data to-go (that is out of hyrax and read the files themselves). That being said I think you have the right idea maintaining an interface where tensors go in with a shape and come out with that same shape. 
* How hard is it to do a read fall-back? I know several of our investigators have data in the old format. If we can put hyrax in a mode where we read both formats but no longer write the old one that would be ideal for them.
* For Metadata I'm on the fence. We went with the pointer-based strategy on the thought that some users of an InferenceDataset (umap, similarity search, etc) did not need the metadata, but other users (visualize) did. The idea was to only create the old datasets's metadata when needed and to provide provenance back to the original dataset. This comes at the cost of maintaining indirection and that it is brittle to a user moving inference datasets off the system where the original dataset path is valid. Certainly embedding metadata would be less brittle, but it may be more challenging to find out "where did this metadata come from?"
* Name change tensor -> model_output 👍

drewoldag:
I've pulled the original .npy code back into InferenceDataset for backward compatibility, and I've updated `save_to_database.py`, which I believe is the only other place where backward compatibility is a concern. 

I've updated the name from `tensors` to `model_output`.

The biggest remaining concern is that Pyarrow, as far as I can tell, requires you to open and read into memory an existing file if you intend to append to it. Apparently `fastparquet` supports appending directly to a file without reading into memory. However, it's not clear if `fastparquet` would support the numpy arrays that @hombit was telling us about earlier today. 

So the question becomes, is the time penalty of (opening + appending + writing) x N worth the value of being able to export the inference results and play with them outside of Hyrax? Or is that time penalty so bad that we would want to use `fastparquet` to do the writing. _OR_ am I missing something obvious, and either you can append with PyArrow or you can persist multidimensional numpy arrays with `fastparquet`?

drewoldag:
The usage pattern here is roughly:
1) At run time we instantiate a `InferenceDatasetWriter` object
2) Batches of input data of a specified size (often 512) are passed through a model for inference
3) When inference of the batch is complete a call is made to the InfereceDatasetWriter.write_batch method
4) We expect the batch to be save in a single write operation

`ParquetWriter` seems like a good choice, but it will take a fair amount to massage the structure into a context manager to ensure that we can close the file. 

Seems like figuring out an efficient way to write the data is going to be the biggest hurdle for this PR...

hombit:
Year, I see that it can be hard to refactor everything for that. Also, 512 rows could be too small for a single row-group, but I don't know how large each row is. It is better to use byte size as guides, e.g. ~100MB per row group

drewoldag:
Going to close out this PR - it was experimental and more substantial than originally thought. I'll add a link to this in the design doc that we have for investigating file formats for InferenceDataset.

PR #391, opened 2025-08-26
Changing `weights_filepath` to `weights_filename`

drewoldag:
Updating the `[train]` configuration file key name from `weights_filepath` to `weights_filename`.

Since the config parameter is actually used as a file name and not a path, this naming is more correct. There is also an `[infer][model_weights_file]` but that one is actually a path, not a file name. Considered changing that to `model_weights_filepath` but taking small steps here.


aritraghsh09 (approved):
This looks good to me with one suggested check:- 

- We should check in case any notebook other than the training notebook is also using `weights_filepath` and edit that appropriately.

PR #403, opened 2025-09-03
Fix benchmark bug and qdrant vector db implementation bug

drewoldag:
Found the bug in the qdrant benchmark tests that also revealed a bug in the qdrant impl code. Fixed them both and updated unit tests.


benchmarks/vector_db_benchmarks.py
drewoldag, line 1: In this benchmark, the `shape` and `vector_size` need to be the same. But they were hardcoded to different values.

maxwest-uw (approved):

PR #404, opened 2025-09-03
Implement simple benchmark to request data from dataset

drewoldag:
This benchmark configures a HyraxRandomDataset with a (3, 64, 64) "image" with a total of 4096 samples. It then randomly picks 128 indexes and records timing metrics for retrieving all 128 samples by index. 

We can use this same metric to see if there are substantial differences when we implement the HyraxQL approach that will introduce the DataProvider class in between the user and the dataset.

### DIFF
--- a/benchmarks/data_request_benchmarks.py
+++ b/benchmarks/data_request_benchmarks.py
@@ -11,7 +11,6 @@ class DatasetRequestBenchmarks:
 
     def setup(self):
         """Prepare for benchmark by defining and setting up a random dataset"""
-        self.h = Hyrax()
         self.tmp_dir = tempfile.TemporaryDirectory()
         self.input_dir = Path(self.tmp_dir.name)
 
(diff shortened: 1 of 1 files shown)


benchmarks/data_request_benchmarks.py
```diff
@@ -0,0 +1,36 @@
...
+
+from hyrax import Hyrax
+
+
+class DatasetRequestBenchmarks:
+    """Timing benchmarks for requesting data from the Hyrax random dataset"""
+
+    def setup(self):
+        """Prepare for benchmark by defining and setting up a random dataset"""
+        self.h = Hyrax()
```
drewoldag, line 14:
```suggestion

```

benchmarks/data_request_benchmarks.py
```diff
@@ -0,0 +1,35 @@
...
+        self.ds = self.h.prepare()
+
+        self.indexes = np.random.randint(0, num_vectors, size=128, dtype=int)
+
+    def time_request_all_data(self):
+        """Benchmark the amount of time needed to retrieve all the data from
+        the random dataset
+        """
+        for indx in self.indexes:
+            self.ds[indx]
```
gitosaurus, line 35: I don't see anything in here or with a decorator that does timing.  Should this have a different name or docstring? I recognize that this is under a `benchmarks/` directory, so there may be some implicit machinery that I don't see.

gitosaurus (approved):
LGTM!

benchmarks/data_request_benchmarks.py
drewoldag: Yeah, this is run automatically by ASV, which will look at the method names for the prefix `time_` for timing or `peakmem_` for memory usage.

PR #405, opened 2025-09-03
Removes references to "table.sub_table" configuration settings when inappropriate

drewoldag:
We added a couple of configuration settings where we wanted the table name to include `.`s, which requires the use of double quotes, as in `["torch.optim.Adam"]`. Which is then accessed as `h.config["torch.optim.Adam"]`

However, we started using this incorrectly when we wanted a subtable, as in `["vector_db.chromadb"]` and then access it as `h.config["vector_db.chromadb"]`, when what we really want is `h.config["vector_db"]["chromadb"]`.

This PR cleans up the places where we incorrectly used the double quotes when creating subtables in the config files, as well as the assorted places in the code where we utilize the config values.


aritraghsh09 (approved):
This looks good to me!

PR #453, opened 2025-09-26
Fix Parser help when config file is specified early on cli. Present error on nonexistant config.

mtauraso:


hombit (changes requested):
Thank you, it actually works now! Could you please also change the documentation to put `-c` after the verb, because it doesn't work as it is specified

Image

hombit (approved):
Thank you!

PR #462, opened 2025-09-29
Use of `fetch_dataset_class` instead of accessing DATA_SET_REGISTRY

drewoldag:
This should automatically load any dataset classes that are requested as libpaths. We were already doing this for models, but for some reason, we weren't doing this for dataset classes. 

Also removed `DataProvider.validate_request`.

Updated unit tests.

### DIFF
--- a/src/hyrax/data_sets/data_provider.py
+++ b/src/hyrax/data_sets/data_provider.py
@@ -182,6 +182,7 @@ class DataProvider:
             dataset_class = dataset_definition.get("dataset_class")
             if not dataset_class:
                 logger.error(f"Model input for '{friendly_name}' does not specify a 'dataset_class'.")
+                raise RuntimeError(f"Model input for '{friendly_name}' does not specify a 'dataset_class'.")
 
             # It's ok for data_location to be None, some datasets
             # (e.g. HyraxRandomDataset) may not require it.
--- a/src/hyrax/data_sets/data_provider.py
+++ b/src/hyrax/data_sets/data_provider.py
@@ -3,7 +3,7 @@ from typing import Any
 
 import numpy as np
 
-from hyrax.data_sets.data_set_registry import DATA_SET_REGISTRY
+from hyrax.data_sets.data_set_registry import fetch_dataset_class
 
 logger = logging.getLogger(__name__)
 
@@ -67,8 +67,6 @@ class DataProvider:
 
         self.config = config
         self.data_request = generate_data_request_from_config(self.config)
-
-        self.validate_request(self.data_request)
 
         self.prepped_datasets = {}
         self.dataset_getters = {}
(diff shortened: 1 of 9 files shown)


src/hyrax/models/model_registry.py
```diff
@@ -168,12 +168,18 @@ def fetch_model_class(runtime_config: dict) -> type[nn.Module]:
...
+            "e.g. 'HyraxCNN' or 'my_package.my_module.MyModelClass'."
+        )
+
     try:
-        model_cls = cast(type[nn.Module], get_or_load_class(model_config, MODEL_REGISTRY))
-    except ValueError as exc:
-        raise ValueError("Error fetching model class") from exc
+        model_cls = cast(type[nn.Module], get_or_load_class(model_name, MODEL_REGISTRY))
+    except Exception:
+        raise
```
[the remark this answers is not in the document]
drewoldag: Makes sense, updated.

src/hyrax/data_sets/data_provider.py
```diff
@@ -172,42 +170,6 @@ def is_map(self):
         """DataProvider datasets will always be map-style datasets."""
         return True
 
-    # ^ Since the getter methods are dynamically created when a dataset class is _instantiated_,
-    # ^ we can't really validate that the requested fields exist until after instantiation.
-    # ^ And instantiation doesn't happen here, it happens in `prepare_datasets`.
-    # ^ See: https://github.com/lincc-frameworks/hyrax/issues/419
-
-    @staticmethod
-    def validate_request(data_request: dict):
```
drewoldag, line 181: This method has gotten smaller and smaller over time. Now that all of the validation logic is contained in either the `fetch_dataset_class` utility function or the `prepare_datasets` method in this class, it was time to fully remove it.

src/hyrax/data_sets/data_set_registry.py
drewoldag, line 1: Took the opportunity here to rename any instance of `data_set` to `dataset`. Also changed the function to accept a class_name string instead of a dictionary. This is because, with HyraxQL, there are potentially many different class name strings in the config, and the logic to pick those out already exists in `data_provider`, we don't want to repeat it here.

src/hyrax/plugin_utils.py
drewoldag, line 1: Updated `get_or_load_class` to accept a string instead of a dict. (Originally the class name string was always in the same place in the configuration dictionary. But with HyraxQL, that isn't the case.

We have different code paths now depending on whether a registry dictionary was provided or not. This is an attempt to make the logic a bit more clear - happy to take feedback on this one, if there's a preferred way to make it easier to read.

tests/hyrax/conftest.py
```diff
@@ -8,7 +8,6 @@
 from torch.utils.data import Dataset, IterableDataset
 
 import hyrax
-import hyrax.data_sets
 from hyrax.data_sets import HyraxDataset
```
dougbrn, line 11: given the data_set -> dataset naming push here, is it fine that these are still data_sets? I imagine changing these would cause potential issues for users, so understand if this was preserved for that reason

dougbrn (approved):
Looks good to me, just one question that I don't think is blocking in any capacity

PR #468, opened 2025-10-02
WIP - Attempting to make ConfigManager easier to use

drewoldag:
WIP - Attempting to make ConfigManager easier to use so that we can call it multiple times during runtime.

If we can make use of staticmethods in ConfigManager to easily refresh the state of the configuration so that _ALL_ defaults are fully resolved just prior to executing `train`, `infer`, etc. would be very useful. 

This would allow a user to instantiate hyrax, _then_ edit the config, and have confidence that the desired default configs will exist. As in this contrived notebook cell:
```
from hyrax import Hyrax
h=Hyrax()
h.config["model"]["name"] = "external_hyrax_example.example_model.ExampleModel"

h.train()  
```
Currently `h.train()` will break because the default config from ExampleModel isn't loaded in the runtime config.

### DIFF
--- a/src/hyrax/config_utils.py
+++ b/src/hyrax/config_utils.py
@@ -217,8 +217,14 @@ class ConfigManager:
             self.user_specific_config
         )
 
-        self.overall_default_config = TOMLDocument()
-        self._merge_defaults()
+        # 1) merge all the external library config dictionaries together
+        self.external_default_config = self.merge_external_default_configs(self.external_library_config_paths)
+        # 2) merge the external library configs on top of the hyrax defaults
+        #! ? Should we just create the hyrax_default_config inside this staticmethod?
+        self.overall_default_config = self.merge_default_configs(
+            self.hyrax_default_config, self.external_default_config
+        )
+        # 3) merge the user config on top of the overall defaults
         self.config = self.merge_configs(self.overall_default_config, self.user_specific_config)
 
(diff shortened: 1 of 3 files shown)


src/hyrax/hyrax.py
```diff
@@ -94,6 +94,27 @@ def __init__(self, *, config_file: Optional[Union[Path, str]] = None, setup_logg
...
+        Parameters
+        ----------
+        key : str
+            The dotted key to set, e.g. "model.name"
+
+        value : Any
+            The value to set the key to.
+        """
+        self.config_manager.set_config(key, value)  # type: ignore[arg-type]
+        self.config = self.config_manager.config
```
mtauraso, line 116: If we froze this dict (recursively) we could prevent the `[]` method of setting?

Not sure if we want that, but this seems like maybe the easiest place to add that trip-wire.

mtauraso (approved):
Looks good to me. Only comment is very minor.

src/hyrax/hyrax.py
drewoldag: Hmm, that's an interesting thought. I would think that would give the effect that we're after. Perhaps I'll keep this PR as is, and open a fresh issue for an investigation into that approach.

drewoldag: https://github.com/lincc-frameworks/hyrax/issues/472

PR #474, opened 2025-10-07
Updating built-in models that are not already using `data_sample`

drewoldag:
Some of the built-in models were still using the `shape` parameter. (About half weren't doing anything with that parameter)

This PR updates the models to make sure that they are all using `data_sample` and not `shape`. Plus whatever logic is required to make sure that the new data_sample is being used correctly.

### DIFF
--- a/src/hyrax/models/hsc_autoencoder.py
+++ b/src/hyrax/models/hsc_autoencoder.py
@@ -13,7 +13,7 @@ class HSCAutoencoder(nn.Module):  # These shapes work with [3,258,258] inputs
     This autoencoder is designed to work with datasets that are prepared with Hyrax's HSC Data Set class.
     """
 
-    def __init__(self, config, shape):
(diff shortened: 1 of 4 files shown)


aritraghsh09 (approved):
This looks good to me. 

I have just one non-blocking comment: I remember putting in the `shape` parameter in some of the models because otherwise the logger would complain. I am assuming it's still the same and we can't just remove `data_sample` from the models that are not using them?

drewoldag:
@aritraghsh09 you're correct, we assume that the model __init__ method will accept a config dictionary and a data_sample (in whatever form is returned from the model's `to_tensor` method) 

So, yes, if you define your model like this, you'll get an error message when we try to instantiate the model.
```
class Model():
    def __init__(self, config):  # <-- error here - we automaticallly pass "config" and "data_sample" to the model.
        #pass

PR #485, opened 2025-10-14
Attach final metrics to model returned from `h.train()`.

drewoldag:
This PR does a couple of clean up/bug fix tasks in addition to the interesting work.

1. Removes unnecessary calls to `to_tensor` in autoencoder models
2. Cleans up naming from `data_set` to `dataset` in one pytorch-ignite function
3. Attaches final metrics to the model output from `h.train()`.

With point 3 we can now do something like this: 
```
h = Hyrax()
m = h.train()
m.final_training_metrics['loss']
>> 321.4354
```

The reason this is important is that this unblocks the use of `Optuna` as a tool for searching the hyperparameter space.

Here's some code that does a simple Optuna optimization:

```
import hyrax
import optuna

h = hyrax.Hyrax()
h.config["model"]["name"] = "HyraxAutoencoder"
h.set_config('train.epochs', 1)

def objective(trial):
    base_channel_size = trial.suggest_categorical("base_channel_size", [8, 16, 32, 64])
    h.set_config("model.base_channel_size", base_channel_size)
    m = h.train()
    return m.final_training_metrics['loss']  # Access the final loss from the model object

study = optuna.create_study(direction="minimize")
study.optimize(objective, n_trials=10)

trial = study.best_trial
print(f"Best trial: Value: {trial.value}, Params: {trial.params}")
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,7 @@ dependencies = [
     "pooch", # Used to download data files
     "onnx", # Used to export models to ONNX format
     "onnxruntime", # Used to run ONNX models
+    "onnxscript", # Used to convert PyTorch models to ONNX format
     "plotly", # Used in 3d visualization
     "psutil", # Used for memory monitoring
     "tqdm", # Used to show progress bars


mtauraso (approved):
🚢 it

PR #487, opened 2025-10-16
Enabling distinct splits to be defined for training

drewoldag:
Imposes expectation that `train` and `infer` splits will be defined separately.

Users are required to have `model_input.train.<friendly_name>` and `model_inputs.infer.<friendly_name>`
If just those two, then we'll assume that the user want to do a percentage-based dataset split for training.

If `model_inputs.train.<friendly_name>` and `model_inputs.validate.<friendly_name>` are present, then we assume that the user has train/validation data in different directories or is split in some other way, so we won't do the percentage-based splitting.

Still need to do
- [x] Respond to PR comments
- [ ] Add more tests to ensure backward compatability
- [ ] Looks like benchmarks are broken again

### DIFF
--- a/docs/notebooks/model_input_1.ipynb
+++ b/docs/notebooks/model_input_1.ipynb
@@ -53,15 +53,18 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "model_inputs_definition = {\"data\": {\"dataset_class\": \"HyraxRandomDataset\"}}\n",
-    "\n",
+    "model_inputs_definition = {\n",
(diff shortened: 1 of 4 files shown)


aritraghsh09 (commented):
This looks (in line with our last conversation and) good to me!

mtauraso (approved):
Minor nit to take copilots suggestions of:
- Correct the spelling of "testing" in the filename
- Remove the commented out code

Don't really agree with any of its other changes, generally this looks good to merge otherwise.

PR #490, opened 2025-10-16
Adding `onnxscript` dependency.

drewoldag:
See the failing workflow https://github.com/lincc-frameworks/hyrax/actions/runs/18552893012/job/52883692299 as an example. These workflows started failling yesterday. Python 3.9 seems unaffected because it is no longer supported by onnx, and onnxruntime


mtauraso (approved):
🏆

PR #492, opened 2025-10-18
Allow users to provide an externally defined collate function to PyTorch's DataLoader

drewoldag:
This PR adds a new default config parameter, `[data_loader][collate_fn] = false`. Using `false` will cause PyTorch to use the default collate function. Otherwise the user is expected to provide a libpath like `package.module.function` which will be passed as a callable to the DataLoader instance.

If/when we provide a default Hyrax collate function, we'll need to update the `load_collate_function` but for now, we're assuming it's either 1) use the default or 2) the user provides their own collate function.

### DIFF
--- a/src/hyrax/pytorch_ignite.py
+++ b/src/hyrax/pytorch_ignite.py
@@ -166,7 +166,9 @@ def load_collate_function(data_loader_kwargs: dict) -> Optional[Callable]:
     Optional[Callable]
         The collate function if specified, else None
     """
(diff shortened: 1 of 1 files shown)


dougbrn (approved):
This looks good to me, has it been tested in any way?

dougbrn:
I can give it a shot once this is merged! (Or work off the branch if this is going to hang around for a bit, just let me know)

PR #494, opened 2025-10-22
Reduce verbosity of model logging

drewoldag:
When running train or infer, the logging for which model is being trained is verbose and unhelpful. All we really need is the model name.

## Change Description

- [ ] My PR includes a link to the issue that I am addressing


dougbrn (approved):
Thank you!

PR #533, opened 2025-11-12
Explicitly use `fork` in InferenceDatasetWriter

drewoldag:
Specifying that multiprocessing should use "fork" in InferenceDatasetWriter.


dougbrn (approved):
Looks good to me! One suggestion, considering the rabbit hole this spawned by only being Pool(), I would love a comment here highlighting the importance of this addition

PR #535, opened 2025-11-13
Model updates rebased to main

aritraghsh09:
This PR is a result of the LSST-Hyrax Sprint Week.

This brings in a lot of new changes with regards to 

- How DonwloadedLSSTDataset Works
- Updates to 2D Visualization
- Updates to 3D Visualization
- Adds the `SimCLR` Model to Hyrax
- Some changes to how metadata with NaNs are handled in `data_set_registry.py`

### DIFF
--- a/tests/hyrax/test_downloaded_lsst_dataset.py
+++ b/tests/hyrax/test_downloaded_lsst_dataset.py
@@ -198,15 +198,15 @@ def test_subset_catalog_preserves_manifest_and_loads_correct_cutouts():
         # Mock the apply_transform method to return tensors as-is for testing
         with mock.patch.object(dataset, "apply_transform", side_effect=lambda x: x):
             # dataset[0] should return cutout for object 1003
-            cutout_0 = dataset[0]
+            cutout_0 = dataset[0]["data"]["image"]
             assert cutout_0.item() == 1003.0, f"dataset[0] should contain object 1003, got {cutout_0.item()}"
 
             # dataset[1] should return cutout for object 1007
-            cutout_1 = dataset[1]
+            cutout_1 = dataset[1]["data"]["image"]
             assert cutout_1.item() == 1007.0, f"dataset[1] should contain object 1007, got {cutout_1.item()}"
 
             # dataset[2] should return cutout for object 1009
-            cutout_2 = dataset[2]
+            cutout_2 = dataset[2]["data"]["image"]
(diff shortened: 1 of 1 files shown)


src/hyrax/data_sets/data_set_registry.py
```diff
@@ -272,7 +277,16 @@ def metadata(self, idxs: npt.ArrayLike, fields: list[str]) -> npt.ArrayLike:
...
 
-        return self._metadata_table[idxs][columns].as_array()
+        result = self._metadata_table[idxs][columns].as_array()
+
+        # Convert masked arrays to regular arrays with NaN for masked values
+        import numpy as np
+        import numpy.ma as ma
+
+        if ma.isMaskedArray(result):
+            result = ma.filled(result, np.nan)
```
drewoldag: Should this be in the child dataset class? This feels like it could be one of those magical things that happens without the knowledge of the user or explanation.

src/hyrax/verbs/visualize.py
```diff
@@ -80,24 +80,28 @@ def run(
...
         if self.config["visualize"]["display_images"]:
             fields += [self.filename_column_name]
 
         # If no input directory is specified, read from config.
         if input_dir is None:
             logger.info("UMAP directory not specified at runtime. Reading from config values.")
-            input_dir = (
-                self.config["results"]["inference_dir"] if self.config["results"]["inference_dir"] else None
-            )
+            input_dir = self.config["results"]["inference_dir"]
```
drewoldag: We typically have a third fallback here, something like:
1. Use `input_dir` the user provided when calling h.visualize
2. If no `input_dir`, use `config['results']['inference_dir']`
3. If no results.inference_dir use `find_most_recent_results()` to get the most recent umap results. 

Is there a particular reason we don't do that here?

drewoldag (approved):
I left 2 comments in this PR, but I believe that neither should be prevent merging this PR. I took that approach that the 3d viz is what it is, and I didn't review it thoroughly. I looked over the stuff in data_sets, and glossed over what is included in tensor_cache_mixin. Changes in `models` looks fine. `visualize` had a comment that is non-blocking, and I know that next on our agenda is a big refactor of visualize anyway.

mtauraso (approved):
Generally agree this should go in.

I looked over the object ID stuff and it seems about what I would expect, as well as a surface level read of everything that isn't 3d viz.

There is a lot here that we need to improve, but I think we also know what parts we need to improve and they're all linked up to major refactors.

PR #540, opened 2025-11-20
Add semi-hidden feature to allow recording metrics at the end of each training epoch

drewoldag:
This change add a new event handler to the training engine that will look for a model method named `log_epoch_metrics`. If found, it will call the method on the model at the end of each training epoch. 

It is expected (though not enforced) that `log_epoch_metrics` will return a dictionary similar to what `train_step` returns. We will then log each key of the dictionary under `training/training/epoch/<foo>` in TensorBoard or `training/epoch/<foo>` in MLFlow. 

The anticipated use case is that model developers will accumulate values in `train_step` over the course of an epoch, then use those values to calculate a metric at the end of the epoch before resetting the accumulators. 

Just a simple example after implementing `log_epoch_metrics` in HyraxAutoencoder that returns 3 "metrics" with random values over the course of training for 5 epochs.
Screenshot 2025-11-20 at 12 04 43 PM

### DIFF
--- a/src/hyrax/pytorch_ignite.py
+++ b/src/hyrax/pytorch_ignite.py
@@ -717,6 +717,17 @@ def create_trainer(
         logger.debug(f"Epoch {trainer.state.epoch} run time: {trainer.state.times['EPOCH_COMPLETED']:.2f}[s]")
         logger.debug(f"Epoch {trainer.state.epoch} metrics: {trainer.state.output}")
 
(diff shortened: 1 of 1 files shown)


dougbrn (approved):
This looks good for rapid turnaround on applecider incubator, I think this should evolve into something more tunable within the train_step eventually so as to avoid the train_step needing modification on top of the definition of this special method.

PR #581, opened 2026-01-06
Updating docstrings so that sphinx will build the documentation.

drewoldag:
Small edits to docstrings:
* whitespace
* type hints

Autoapi was trying and failing to import _version.py. I was unable to instruct it to ignore that file, so instead I have suppressed that warning. Which may or may not be a great thing.


mtauraso (approved):
Minor nit we probably want to turn back on "-W" since it protects us from broken internal links.

Otherwise this looks reasonable.

PR #587, opened 2026-01-10
Some tidying up of introductory documentation

gitosaurus:
## Change Description

Small initial PR to fix some build and syntax issues, and to activate the Hyrax verbs and sample notebooks sections.


drewoldag (approved):
This looks fine for now. I would like to use `Common workflows` as the eventual landing area for the notebooks that are not showing specific science cases. I think that there's a high probability that the majority of the notebooks currently in docs/notebooks.rst are out of date. 

So perhaps as a group we can work through those, identify the ones that we want to keep/update and let the rest of them go. (Not part of this PR though :))

PR #591, opened 2026-01-15
Removed return type hint

drewoldag:
More than anything else, this makes the red squiggly underlines in my IDE go away.


mtauraso (approved):
🚢 it

PR #593, opened 2026-01-15
Data Caching for tensors

### LINKED ISSUE #508: Implement a data cache for training
Currently HSCDataset implements a data caching layer for faster data access after the first epoch of training. This should be generalized so that any dataset can take advantage of it.

This probably doesn't belong in DataProvider, and should likely be someplace after to_tensor. If it's placed after to_tensor, or rather the cached data is the output of to_tensor, we can sidestep a potentially large amount of work. 

This should be implemented in such a way that the user can enable/disable it via config. It should likely be disabled by default. If it is enabled by default, then the user may run into OOM errors due to the cache without understanding why. 

If the user has to take an explicit action to enable the cache, then we can provide hints/instruction at the time to let them know that OOM is a possibility.

### LINKED ISSUE #509: Implement pre-caching
This feature would fill the cache ahead of time, or at least in parallel with processing batches, so that the trainin cache is filled before the first training epoch has finished. 

HSCDataset implements this as well, but we could make this dataset independent. 

It's also possible that PyTorch DataLoader can handle this for us with a built in pre-fetch setting.

mtauraso:
Generalize Data Cache to all tensors from all datasets:
* Caching occurs on the return from DataProvider.resolve_data, this means we're caching numpy dicts, but does mean some post-processing steps will be re-run as a matter of course.
* TensorboardX logger has been factored out to a global singleton to simplify logging of cache misses and hits (and clean up several other places where we had to pass it everywhere)
* Caching counts the data size in a bottoms-up manner accounting for objects which share memory. This should allow us to surface accurate information about cached data size in future
* We are now strict about not allowing datasets to return pytorch objects, and DataProvider only accepting single integers as indexes (rather than pass-through multi-indexing).

This probably needs some tests. It is not exercised as well as one would like with existing tests.


drewoldag (approved):
Overall this looks good to me. My only additional comment is about the default value of `['data_set']['use_cache']`. Since that is a bit of magic that could potentially result in OOM, perhaps it makes sense to set that and the pre_load to False? 

If those are False, we could provide some logging and documentation that lets the user know that performance would be better if they set those values to True, but would also run the risk of OOM if they didn't allocate enough memory.

PR #595, opened 2026-01-16
Adding links to the most useful of the old reference docs 

mtauraso:
This makes the old docs easily findable while we are doing the docs sprint.


gitosaurus (approved):
LGTM

PR #596, opened 2026-01-17
Adding Core concepts page

drewoldag:
Core concepts with links to "what you need to provide", "config system" and "data flow"


docs/required_input.rst
```diff
@@ -0,0 +1,71 @@
+Required inputs
+===============
+
+Hyrax covers a lot of the "plumbing" required to run machine learning experiments,
+however there are still a few key places where a user needs to provide input for
+Hyrax to know what to do.
```
gitosaurus: I think these lines are redundant and can be replaced with the next paragraph.

maxwest-uw (approved):

docs/required_input.rst
```diff
@@ -0,0 +1,71 @@
...
+there are three essential pieces of information that a user must provide:
+
+#. The model to use for training or inference
+#. The data that will be provided to the model
+#. The function needed to prepare the data for the model
+
+.. admonition:: Hyrax design choice
+
+   Why do I need to provide data *and* tell Hyrax how to prepare the data for
+   the model? That seems like extra work!
```
gitosaurus: Have users given feedback that they expected provide + prepare to be the same thing? (Do they think the preparation should be automatic?)

docs/required_input.rst
```diff
@@ -0,0 +1,71 @@
...
+   This increases reusability and allows users to quickly swap out datasets and models
+   without needing to make changes to the code in either.
+
+Instructions for Hyrax come in two forms - simple configuration settings and user-defined code.
+
+
+Define the model
+------------------
+At it's core, Hyrax is a framework for training and running machine learning models.
+To do this, a user must provide the model that Hyrax will use for training or inference!
```
gitosaurus:
```suggestion
To do this, a user must provide the model that Hyrax will use for training or inference.
```

I think this and the next paragraph can be combined.

docs/required_input.rst
```diff
@@ -0,0 +1,71 @@
...
+In order to convert the output of a dataset into a form that is suitable for a model,
+Hyrax requires a user-defined function that will prepare the data for the model.
+
+As a contrived example, consider a dataset where each data point is a composed of
+three 2d arrays: "science", "mask", and "variance".
+To be flexible, the dataset class that provides data from disk would return each
+array individually.
+The model may expect that a single flattened array as input.
+If the model accepted a single flattened array as input, the user would need to
+define a `to_tensor` function that takes in the structured data sample and stack
```
gitosaurus:
```suggestion
define a `to_tensor` function that takes in the structured data sample and stacks
```

gitosaurus (approved):
LGTM! Minor feedback only.

docs/required_input.rst
```diff
@@ -0,0 +1,71 @@
...
+Instructions for Hyrax come in two forms - simple configuration settings and user-defined code.
+
+
+Define the model
+------------------
+At it's core, Hyrax is a framework for training and running machine learning models.
+To do this, a user must provide the model that Hyrax will use for training or inference!
+
+There's no reasonable default model that Hyrax could provide, so this is an essential piece
+of information that a user must provide.
```
dougbrn: I think it would be helpful here to provide a small code cell that shows a dummy model definition, as small as possible, I took a rough stab here:
```
from hyrax.models import hyrax_model
import torch.nn as nn
import torch.nn.functional as F

@hyrax_model
class Model(nn.Module):
    def __init__(self) -> None:
        super().__init__()
        self.conv1 = nn.Conv2d(1, 20, 5)
        self.conv2 = nn.Conv2d(20, 20, 5)

    def forward(self, x):
        x = F.relu(self.conv1(x))
        return F.relu(self.conv2(x))

    def train_step(self, batch):
        decoded = self.forward(batch)
        loss = self.criterion(decoded, labels)
        self.optimizer.zero_grad()
        loss.backward()
        self.optimizer.step()
        return {"loss": loss.item()}

    def to_tensor(data_dict):
        # see below!
```

A link to an eventual tutorial that shows how this would be created would also be great, and if that's the plan don't worry about this. The core of my suggestion is that this is a great point to introduce what a model looks like to Hyrax, with the strength for some users being that it looks a lot like a Pytorch model.

dougbrn (approved):
I think this looks great, in my opinion it would benefit a lot from some short example snippets, or links to other pages with examples (now or sometime later in the doc sprint). Fully optional though and will approve!

docs/required_input.rst
drewoldag: Removed these lines

docs/required_input.rst
drewoldag: Going to pull this out, if someone does ask, we'll bring it back in.

docs/required_input.rst
drewoldag: Done

PR #629, opened 2026-01-24
Fix _get_available_bands_from_manifest to find complete band entries

aritraghsh09:
We previously had a hacky solution to determining the reasonably consistent band structure in the manifest file. I am now attepmpting to define this more rigorously:- 

- While downloading files, `DownloadedLSSTDataset` always saves the tensor with as many bands as the number of bands requested (e.g., 6 for u, g, r, i, z, y) and just pads with NaNs for the bands not available and skips that band in the `downloaded_bands` entry in the manifest file

- To detect the band structure used during download (needed to filter correctly), we now look for the first 5 entries where as many bands were downloaded  as the number of tensor filters being saved (i.e., a full set download -- no bands missing) 

- We make sure all the 5 are consistent 

- We go with the common structure as our detected band structure 

In terms of primary changes:
- I updated the `get_available_bands_from_manifest` to have the new logic
- I put in a new test to check for this intended behaviour now.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,6 +85,7 @@ dev = [
     "sphinx-togglebutton",
     "sphinx-rtd-theme",
     "lsdb", # Used to test lsst dataset classes
+    "cdshealpix <= 0.7.1",
 ]
 
 [build-system]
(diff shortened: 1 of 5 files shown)


mtauraso (approved):
Only real request is a minor change to an error message. Other than that it looks fine.

PR #638, opened 2026-01-27
Removing Python 3.10 support

mtauraso:
Not adding python 3.14 because onnx runtime doesn't support it without going to pre-release versions.


drewoldag (approved):
Looks good!

gitosaurus (approved):
LGTM!

PR #641, opened 2026-01-27
Remove python 3.10 compatibility pin

mtauraso:
This pin somehow stayed in when I removed 3.10. This PR removes it. CI passing will show it was the right move.


gitosaurus (approved):
LGTM!

PR #661, opened 2026-02-03
Fix bug in Pydantic validation

gitosaurus:
## Solution Description

The collect_additional_datasets model validator modified the input dict in-place, which is not good practice.  When reusing such an input (e.g., adding 'infer' after setting 'train'), the empty 'other_datasets' would cause DataProvider to fail with "No datasets were requested in data_request".

While simply preventing the in-place mutation fixes the bug, and I have that in a separate commit, I decided to completely remove this part of validation entirely, since we already know that we want lighter validation.

## Code Quality


drewoldag (approved):
This is reasonable enough to go ahead with. We discussed additional tweaks to the way that "other_datasets" should work in person, which will be in a separate PR.

PR #664, opened 2026-02-03
Introducing verb-specific batch functions in models to fix the `torch.no_grad` issue with the `test` verb

### LINKED ISSUE #662: Fix `test` verbs usage of `torch.no_grad`
The `test` verb attempts to make use of `train_step`, while also setting the torch.no_grac context manager. 

These two things are mutually exclusive because in `train_step` we typicially call `loss.backward()` which expects gradients to have been tracked, and it calls `self.optimizer.step()` which updates the model weights. 

This agreed upon approach is now: 
1) Create a separate model method for `test` (and `validate`)
2) Create a separate method for `infer` as well.
3) Name all of these functions `*_batch`, i.e. `train_batch`, `validate_batch`, `test_batch`, `infer_batch`.
4) Update the existing built in models so that they implement those correctly
5) Update the pytorch_ignite engine machinery so that it calls the correct model methods.

drewoldag:
Renaming all instances of `train_step` to `train_batch` for clarity.
Introducing`validate|test|infer_batch` methods to all existing models.
Updated engine creation code to look for appropriate model methods when creating the engine. i.e. the trainer engine looks for `train_batch`, the validation engine looks for `validate_batch`.
Search/replaced instances of `train_step` with `train_batch` in the docs and tests.

### DIFF
--- a/docs/data_flow.rst
+++ b/docs/data_flow.rst
@@ -24,7 +24,7 @@ The ``prepare_inputs`` function is responsible for taking in the output of the `
 Model Input and Output Pipeline
 -------------------------------
 
-The data is then either sent to ``pytorch`` ignite for training or to ``onnx`` for inference. In the case of training, the data is converted into a tensor and passed into the model's ``train_step`` function. The model processes the data and returns the output predictions. If the user wishes to perform data augmentations, these can be set up in the model's ``train_step`` function as well.
+The data is then either sent to ``pytorch`` ignite for training or to ``onnx`` for inference. In the case of training, the data is converted into a tensor and passed into the model's ``train_batch`` function. The model processes the data and returns the output predictions. If the user wishes to perform data augmentations, these can be set up in the model's ``train_batch`` function as well.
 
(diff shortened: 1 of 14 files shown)


gitosaurus (approved):
LGTM!

aritraghsh09 (approved):
Reviewed in-person! This looks good to me!

drewoldag:
Your observation is correct, in the built in models, it is the case that validate_batch and test_batch are equivalent. You probably noticed that there is at least one example where test_batch simply calls validate_batch. 

I would assume that the majority of the time, that a user would also have equivalent code in both validate and test batch, but maybe not always.

The statement from Aritra was that this approach feels like it maps well to verbs, as well as the existing mental model for ML where each of these steps would have it's own specific function to call. 

To your point, if a user attempts a train/validate run, but doesn't have a `validate_batch` function, when we see that their model doesn't have a `validate_batch` function, we direct them to that specific chunck of the docs, and show them an example of a model that does have a validate_batch function. This check already happens early to avoid the footgun problem.

PR #673, opened 2026-02-04
Moving precache threads to config

mtauraso:


src/hyrax/hyrax_default_config.toml
```diff
@@ -285,6 +285,11 @@ use_cache = true
 # Warning: Only suitable for situations where the entire dataset fits in system memory
 preload_cache = false
 
+# Number of threads to use for preload. Preload is most effective when your dataset has many small files
+# and your storage layer has high latency or is IOPS limited. Avoid writing your dataset class this way,
+# but if you do then a large number of preload threads will maximize io throughput
```
drewoldag: I'm not sure that general users are going to understand what you mean with these last two lines. How will a user konw if their storage layer has high latency or is IOPS limited. What are they to avoid when writing their dataset class? What is a "large" number of threads?

drewoldag (approved):
What I see here seems fine. I like that you can easily adjust the threads now. I am a little concerned that a general user reading the comment in the default config might not be able to get out of it what they need.

PR #693, opened 2026-02-11
Fixing MLFlow Tracking Directory

aritraghsh09:
In the last two weeks at some point, we had erroneously changed where the MLFlow results were being stored. 

We want the MLFlow results to be stored,not inside individual results directories, but at the root. This was the behaviour before. Not sure when it got changed.


drewoldag (approved):
This seems fine, especially if this unblocks you.

PR #702, opened 2026-02-13
Lr scheduler notebook

SamSandwich07:
Adds an notebook demonstrating the use of an LR scheduler


drewoldag (changes requested):

docs/pre_executed/using_lr_schedulers.ipynb
```diff
@@ -0,0 +1,400 @@
+{
```
drewoldag: <p>In the same way that you called out that the model performance was worse with gamma=0.9, you could do so here. Noting that the goal is to demonstrate the use of different learning rate schedulers more than to tune the model for optimal performance.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lincc-frameworks/hyrax/pull/702/discussion/#comment-2830778686'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

drewoldag (approved):
This is looking very good. Thanks for the effort!

drewoldag (changes requested):
The second image name changed.

drewoldag (approved):
Good to go

PR #707, opened 2026-02-17
Fixing up model wrapping and tracking of metrics code, learning scheduler

mtauraso:
This is a follow-up on #706 which was a targeted fix for the bug reported in dirac slack: https://uw-dirac.slack.com/archives/C08F5FLEY5A/p1771350453909469

This is the more complete/thorough fix.

To avoid putting data/calling functions on the wrong model, the create_* methods now have two local variables "model" and "wrapped_model"  In the case where there is no wrapping done by idist.auto_model() these are the same.  I've tried to update the relevant accesses, but I want both @drewoldag and @SamSandwich07 to sign off before I merge.

I've also removed local variables of scheduler and optimizer in favor of model.scheduler and model.optimizer which are always on the inner model.

This has also exposed a larger issue that our CI doesn't have real GPUs, and we would have caught the original bug, and perhaps several other issues where we memoize data onto the model if we had CI with even an old and cheap GPU.


drewoldag (approved):
This looks good to me.

SamSandwich07 (commented):
We talked about this last week, looks good to me.

SamSandwich07 (approved):

PR #710, opened 2026-02-19
'data' key not required; can be any friendly name.

gitosaurus:
Discovered while testing pre-executed notebooks.

## Solution Description

Previous testing incorrectly assumed that in a data request, the data source name had to be called 'data' if there weren't multiple sources.  This is not true.  Corrected both test and logic.

## Code Quality


drewoldag (approved):
This looks correct to me. I made a similar change in PR #690. I'll pull these changes in when they land on main.

PR #713, opened 2026-02-19
`visualize` verb should find all its inputs from config.

### LINKED ISSUE #709: visualize verb does not read new inference data sets
**Bug report**

The `visualize` verb still uses an `InferenceDataset` directly.  Now that results are written to a `ResultsDataset` by default, `visualize` needs to work with those. On `main`, at present, it fails with an exception, as it is not ready to read the Lance database that is now the default.

**Environment Information**

When executing `docs/pre_executed/visualize.ipynb`:

<details>
<summary>Traceback</summary>

FILL IN YOUR STACK TRACE HERE

RuntimeError: <hyrax-root>/results/20260218-114640-umap-SVxP is corrupt and lacks a batch index file.

**Before submitting**
Please check the following:

- [X] I have described the situation in which the bug arose, including what code was executed, and any applicable data others will need to reproduce the problem.
- [X] I have included information about my environment, including the version of this package (e.g. `hyrax.__version__`)
- [X] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a description of what I expected instead.
- [X] If I have a solution in mind, I have provided an explanation and/or pseudocode and/or task list.

gitosaurus:
Rather that assuming that the dataset provides its own breadcrumbs to its input, use the session-wide config to locate that data and its necessary metadata.

## Change Description

Closes #709 .

## Code Quality

### DIFF
--- a/src/hyrax/verbs/visualize.py
+++ b/src/hyrax/verbs/visualize.py
@@ -20,6 +20,13 @@ class Visualize(Verb):
     cli_name = "visualize"
     add_parser_kwargs = {}
 
+    # Dataset groups that the Visualize verb knows about.
+    # REQUIRED_SPLITS must be present in the data request configuration.
+    # OPTIONAL_SPLITS are used when present but do not cause an error if absent.
+    # Note: "umap" results may also be required in the future.
+    REQUIRED_SPLITS = ("infer",)
+    OPTIONAL_SPLITS = ()
+
     @staticmethod
     def setup_parser(parser: ArgumentParser):
         """CLI not implemented for this verb"""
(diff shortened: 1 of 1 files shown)


src/hyrax/verbs/visualize.py
```diff
@@ -106,10 +107,16 @@ def run(
...
         # Get the umap data and put it in a kdtree for indexing.
-        self.umap_results = InferenceDataSet(self.config, results_dir=input_dir, verb="umap")
+        self.umap_results = load_results_dataset(self.config, results_dir=input_dir, verb="umap")
         logger.info(f"Rendering UMAP from the following directory: {self.umap_results.results_dir}")
 
-        available_fields = self.umap_results.metadata_fields()
+        # Build a DataProvider from the live config for metadata access.
+        # This avoids implicit coupling between result datasets and their original data sources.
+        data_request = generate_data_request_from_config(self.config)
+        infer_request = data_request.get("infer", next(iter(data_request.values())))
```
drewoldag: If you look in PR #690 you'll see that verbs (excluding visualize) define a `REQUIRED_DATA_SPLITS` and `OPTIONAL_DATA_SPLITS` tuples that define the expected splits, in this case, it would be: `REQUIRED_DATA_SPLITS = ("infer",)` though it brings up two questions:
1) Should visualize expect "infer"
2) Should the variables be called something like `REQUIRED_DATA_GROUPS` and `OPTIONAL_DATA_GROUPS`?

src/hyrax/verbs/visualize.py
```diff
@@ -154,7 +161,7 @@ def run(
                     all_indices = list(range(len(self.umap_results)))
 
                     # Extract metadata for the specified column
-                    metadata = self.umap_results.metadata(all_indices, [self.color_column])
+                    metadata = self.metadata_provider.metadata(all_indices, [self.color_column])
                     self.color_values = metadata[self.color_column]
```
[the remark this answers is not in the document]
drewoldag: I don't know enough about Lance to know if this is true or not, but if nothing else it seems like this should be given a comment and warning as a place where index mismatches may occur?

drewoldag (approved):
This seems reasonable enough to me. I agree with most of the comments here. Once addressed, should be good to go.

src/hyrax/verbs/visualize.py
gitosaurus: @copilot Implement this suggestion, following the approach in PR #690; expect "infer" but add a comment that "umap" may also be required.

PR #717, opened 2026-02-21
Working with `ResultsDataset` common workflow notebook

drewoldag:
Adding first draft of common workflow notebook for working with ResultsDataset.


mtauraso (approved):
I think this is good to go. I had some comments on the notebooks that i think are more that there is old stuff still there which ReviewNB didn't line up correctly.

Overall we should merge this, and polish the notebooks after.

PR #733, opened 2026-02-27
[AUTO-MERGE] Removed unused input argument in `create_results_writer`

### LINKED ISSUE #724: Clean up `pytorch_ignite.py:create_save_batch_callback`
I believe (I may be incorrect here) that `result_factory.py:create_results_writer` doesn't require a dataset as input at all. Which, I believe means `pytorch_ignite.py:create_save_batch_callback` doesn't require a dataset as input.

At this point, the object_id values that are written out are taken from the batch that was processed which includes a "object_id" key.

We should also clean up the `create_results_writer` function definition to remove the `original_dataset` parameter.

drewoldag:
There was an unused input parameter on `create_results_writer`, so I've removed that and chased down the various places it was used.


mtauraso (approved):
Looks good

PR #735, opened 2026-02-27
[AUTO-MERGE] Removed `Dataset` from ResultDataset class inheritance

### LINKED ISSUE #732: `ResultDataset` doesn't need to inherit from PyTorch `Dataset`
I don't think that there's any reason for ResultDataset to inherit from PyTorch dataset. It should be safe to remove that parent class. 
Just do that an double check that nothing breaks.

drewoldag:
Just removed `Dataset` from the list of parent classes, and removed the import line.


mtauraso (approved):
🚢 it

PR #745, opened 2026-03-03
[AUTO-MERGE] Testing to see if we can remove the numba pinning.

### LINKED ISSUE #583: Understand why we need to pin our numba dependency for python 3.11
Between Dec 30 and Dec 31, one of our dependencies changed and smoke tests began failing for python 3.11 only. 

It appears that this was due to difficultly resolving a good version of numba, and uv was falling back to a _very_ old version 0.53.1 (released in 2021) This had many down stream affects. 

Pinning the numba version in hyrax's pyproject.toml seems to help, but it's not clear why we need to do this. Ideally we would identify which of our dependencies is causing the problem and fix that.

drewoldag:
## Change Description
Closes #583
Removed the pin from the pyproject.toml file to see if the CI would pass, and indeed it does. I believe that we can simply remove this dependency pin now.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,6 @@ dependencies = [
     "tqdm", # Used to show progress bars
     "qdrant-client", # Vector database for similarity search
     "colorama", # Used to color terminal output
-    "numba==0.63.1",
     "lancedb", # Used for columnar storage of inference results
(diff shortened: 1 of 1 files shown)


mtauraso (approved):
No notes

PR #749, opened 2026-03-04
Checkpointing Notebook

### LINKED ISSUE #743: How to load a checkpoint
Notebook showing how to use checkpoints in hyrax.

SamSandwich07:
Adds a notebook detailing when checkpoints are created and how to load them to resume training.


drewoldag (commented):
Left a few comments on the notebook. Overall looking like a good start. The biggest request is for a tensorboard image at the end. 

And of course, when it's ready, add it to the Common Workflows page :)

drewoldag (approved):
Two small comments with suggested changes. Also, looking at the notebook, it does seem odd that the the config file used to initialize Hyrax is named "working_with_result_data". 

If you're motivated, and since that config is only used in 1 other place, now would be a good time to rename that .toml file to something like `getting_started_config.toml`.

Just let me know if you decide _not_ to make that rename, and I can take care of it.

drewoldag (approved):
Looks great. thanks for renaming the config!

PR #751, opened 2026-03-04
Removed nan-handlers for torch tensors. Updated the unit tests.

### LINKED ISSUE #737: Remove nan handling functions that operate on tensors
There are a few nan handler functions in data_provider.py that are specific to torch tensors. Now that we've agreed on all data being numpy up to the point of being passed to the model, we don't need the nan handlers for tensors. Let's just remove them.

drewoldag:
## Change Description
Closes #737 
Removed nan handlers for torch tensors and updated the associated test file. 
The unit tests were simplified as well. There was an unused (or hardly used) dataset class that would produce data with nans, but that was previously packed into the HyraxRandomDataset. So I removed the old `RandomNanDataset` that was exclusively used for testing this feature. This is ok because HyraxRandomDataset can produce nans in it's random data and exercise the same functionality.

### DIFF
--- a/src/hyrax/data_sets/data_provider.py
+++ b/src/hyrax/data_sets/data_provider.py
@@ -29,44 +29,24 @@ def _handle_nans_numpy(batch, config):
     return _handle_nans_logic_numpy(batch, config)
 
 
-# Conditionally register torch.Tensor if torch is available
-try:
-    import torch
-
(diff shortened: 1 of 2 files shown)


mtauraso (approved):
LGTM 🚢 it

PR #753, opened 2026-03-05
[AUTO-MERGE] Polished existing vector database notebook, added to common workflows

### LINKED ISSUE #730: Demo how to create and query a vector database
Check over the existing notebooks, we may have all the pieces already.

drewoldag:
## Change Description
Closes #730

The vector database notebook was mostly ready to go, I just polished some of the words. Also added to common workflows page. 

FWIW it's a pre-executed notebook, and we could probably export the inference results to a zipped file on zenodo to save ourselves a bunch of cells at the beginning of the notebook. This could be an approach taken for a lot of notebooks...


aritraghsh09 (approved):
Looks good!

Note for the future for all our notebooks: Because we say we are an astronomy framework, we should really move away from CIFAR to something astronomy-related for all our notebooks, but we can do that later once we have basic working docs we are happy with!

SamSandwich07 (commented):
Left a couple questions which I am wondering about. Also, would it be good to include a sentence or two about when a user might want to use ChromaDB vs Qdrant? Or is that something that is already elsewhere or the user is expected to know?

PR #772, opened 2026-03-11
Adding excalidraw diagrams.

drewoldag:
I created a bunch of diagrams in excalidraw a while ago, and didn't realize that I could save them not-online. 

I've moved them into the repo so that we can use them for documentation or in talks.


aritraghsh09 (approved):
These look great! 

The only comment I have is that we should reuse the figures from the paper when possible (only relevant for 1 diagram in this set probably)

PR #774, opened 2026-03-11
Cleaned up HyraxCifarDataset

### LINKED ISSUE #773: Tidy HyraxCifarDataset
This is one of the first dataset classes that we created. Then we expanded it to be a demonstrator of map and iterable style datasets. 

After we removed iterable dataset support, we left the lingering Cifar base class. And after HyraxQL we left some lingering PyTorch relics. 

We should clean up HyraxCifarDataset so that it:

1. Doesn't inherit from pytorch dataset
2. Isn't split into two classes
3. Doesn't define an unnecessary metadata table
4. Doesn't implement unnecessary functions (__get_item__, get_index)

drewoldag:
Removed unnecessary class and methods.
Updated doc string.
Ran all the unit tests, and a bunch of notebooks that use this dataset class.

## Change Description
Closes #773


SamSandwich07 (approved):
Looks good to me

PR #783, opened 2026-03-13
[Auto-merge] Trace Mode

mtauraso:
- Instrumented DataProvider, Engine verb, Dataset class functions, and Model class functions
- Instrumentation only active when trace=1 passed to enabled verbs (train, infer, test, engine)
- Prepare verb moved to class based verb model.
- When in trace mode verbs output a TraceResults which has the data trace.
- TraceResult objects can be easily printed and inspected via __str__,__repr__,__getitem__, and getattr type access.


mtauraso:
@copilot 
Please extend the existing tests on this PR to also access function parameters captured in a trace.
I'd like tests that verify this works by name and also by number.
You do not need to be comprehensive, adding a spot check on each of the existing verb tests that have trace mode enabled will do.

drewoldag (commented):
Still reviewing the trace.py code. But wanted to go ahead and submit these comments now instead of later.

src/hyrax/trace.py
```diff
@@ -0,0 +1,811 @@
...
+    3. "collate" stage
+    Each data column is combined into a single batch tensor in this stage. If your dataset defines a custom
+    collate function (e.g. time-series data with different lengths) it will be evaluated in this stage.
+    Any NaN handling that is configured into hyrax is also performed in this stage.
+
+    4. "prepare_inputs" stage
+    The ML Model's prepare_inputs function is called in this stage and converts the data dictionary
+    containing each column of batched tensor data into a single batch tensor that will form the input to
+    the model's evaluation functions. If the model is doing supervized learning, the output will be
+    a tuple of two tensors (inputs, labels)
```
drewoldag:
```suggestion
    a tuple of tensors (inputs_0, [inputs_1, ..., inputs_n], labels)
```

drewoldag (commented):
Overall this is looking very good. I've worked through trace.py and am leaving comments here. Next I'll play around with the functionality just using the getting started notebook to see it in action.

src/hyrax/trace.py
```diff
@@ -0,0 +1,811 @@
...
+
+    def instrument_dataset_getter(self, dataset, getter, friendly_name, field_name):
+        """
+        Instrument a dataset get_* function. Called by DataProvider to insert shims before
+        any betters are called
+        """
+        trace_def = TraceDef(
+            disp_name=f"{friendly_name}__get_{field_name}",
+            func_name=f"get_{field_name}",
+            params_to_capture={"index": 0},
```
drewoldag:
```suggestion
            params_to_capture={"index": 1},
```

drewoldag (commented):
I'm feeling pretty good about what is here. I've made code suggestions that fix the `UNSUPPORTED TYPE` thing that I was seeing. Going to play with this a little more.

src/hyrax/trace.py
```diff
@@ -0,0 +1,811 @@
...
+            "evaluation": TraceStage(),
+        }
+
+    def __getitem__(self, ref):
+        return self.stages[ref]
+
+    def _valid_keys(self):
+        return list(self.stages.keys())
+
+    def reduce_len(self, cls):
```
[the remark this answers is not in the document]
mtauraso: I'm really not sure either, but we should keep it in mind.

When I was dealing with pytorch ignite's infinite iterator support I came across the notion of setting epoch size independently on an infinite iterator.

If it ends up being a strict length-limit API to data provider, we could definitely call that API here.

src/hyrax/trace.py
mtauraso: I need to make sure this actually works.

drewoldag (approved):
One suggestion in a comment.
Also, one question regarding printing the trace - at the risk of cluttering the printed trace, what do you think of adding a `'input:` and `output:` label for each traced function? 

i.e.: 
```
Trace Stages {
	dataset_getter: [
		data__get_image(index) -> image duration=0.00617 ms
			input: index = 0
			output: image = <numpy.ndarray shape=(1, 32, 32) hash=26660822658842624 device=cpu>
		
		data__get_object_id(index) -> object_id duration=0.00577 ms
			input: index = 0
			output: object_id = '23'
		
	]

PR #793, opened 2026-03-17
Basic CitationCompass integration

mtauraso:
Codex generated this pull request, but encountered an unexpected error after generation. This is a placeholder PR message.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69b894dd70d88331890588106bfd003e)


jeremykubica:
I don't think the command 

```
python -m citation_compass --config citation_compass.toml
``` 

in your PR will work. Currently citation compass doesn't have a main, so it can't be run. It also doesn't process toml files. We could add this to citation-compass if it would be useful though.

PR #802, opened 2026-03-18
s/DataSet/Dataset/g

mtauraso:
Codex generated this pull request, but encountered an unexpected error after generation. This is a placeholder PR message.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69b9d61b6c6c83318c62c1f8768c77e2)


mtauraso (commented):
Honestly this looks really complete to me. No notes.

drewoldag (approved):
Lots and lots of tiny changes, but all looks good. The only comment I had was on the last file in the list.

drewoldag (changes requested):
One thing this PR hasn't touched is the `[data_set]` table name in the default_hyrax_config.

drewoldag:
ok, with the understanding that the configuration `[data_set]` rename will not happen as part of this PR, I can switch this back over to being approved. 

I believe at the end of our conversation thread that the consensus was that we'll attempt to do the configuration name change in a later PR. And that it would be really nice to have versioned configuration names.

drewoldag (approved):
Repeating my last comment in the PR thread - we're in alignment to leave the configuration `[data_set]` rename out of this PR and address it later.

PR #804, opened 2026-03-19
HyraxQL's `dataset_config` now accepts a full sub-config

### LINKED ISSUE #576: Allow for dataset redirection in the model_inputs definitions
**Feature request**
Allow for dataset redirection in the model_inputs definitions. 

If we implement some fallback logic as follows like for external developers of datasets will be much better:
When instantiating a dataset class (or overriding config values) using DataProvider check if the value of `dataset_class` is a valid import path like `<package>.<module>.<class_name>`. If so everything behaves as it currently does. 

If the value is not a valid import path, assume that it is a path to a table/sub-table in the toml file (like what is shown below). If that table/sub-table exists, assume there is a magic key in that sub-table like `module_name` (or equivalent) that will provide the valid import path for creating the class instance, and the location to place the dataset class config parameters..

```
[data_set.acPhotoEventsDataset]
module_name = "applecider.datasets.photo_dataset.PhotoEventsDataset"
manifest_path = "../../../data/photo_events/manifest_train.csv"
stats_path = "../../../data/photo_events/feature_stats_day100.npz"
horizon = 100.0 # days
ideal_class_distribution = [0.3,0.1,0.1,0.3,0.1]
use_oversampling = true

[model_inputs]
[model_inputs.train.data]
dataset_class = "data_set.acPhotoEventsDataset" # <<<<< The magic is here <<<<<
data_location = "../../../data/photo_events/train/"
fields = ["photometry", "label", "mean", "std"]
[model_inputs.train.data.dataset_config]
manifest_path = "../../../data/photo_events/manifest_train.csv"
```

Taking an approach like this also means that in the dataset class code, referencing the config keys is much easier. 
_Before_
`self.config['data_set']['applecider.datasets.photo_dataset.PhotoEventsDataset']`

_After_
`self.config['data_set']['acPhotoEventsDataset']`

drewoldag:
## Change Description
Modified `dataset_config` syntax to accept a full sub-configuration.
Closes #576 

## Solution Description
See the `dataset_config` section below. Note that you can pass a full set of nested dictionaries there. If the key is a known Hyrax built-in dataset class you don't need to build the full nested dictionary structure. If it's an external class like "galaxy10_dataset" shown here, the user will build the full nested dictionary that is present in the external default_config.toml file. A copy of this example's toml file is shown below as well.

``` python
data_request = {
    "train": {
        "data": {
            "dataset_class": "external_hyrax_example.datasets.galaxy10_dataset.Galaxy10Dataset",
            "data_location": "/home/drew/code/external_hyrax_example/docs/pre_executed/data/galaxy_10",
            "fields": ["image", "label"],
            "split_fraction": 0.8,
            "dataset_config": {
                "external_hyrax_example": {
                    "galaxy10_dataset": {
                        "channels_first": True,
                    },
                },
                "HyraxRandomDataset": {
                    "size": 42,
                },
            },
        },
    },
}
```

``` toml
[external_hyrax_example]

[external_hyrax_example.galaxy10_dataset]
# When false, the images will be in (N, H, W, C) format.
# When true, the images will be transposed to (N, C, H, W) format.
channels_first = false
```

### DIFF
--- a/src/hyrax/data_sets/data_provider.py
+++ b/src/hyrax/data_sets/data_provider.py
@@ -533,7 +533,7 @@ class DataProvider:
             tmp_config = {}
             for k, v in dataset_definition["dataset_config"].items():
                 if k in DATASET_REGISTRY:
-                    tmp_config["data_set"] = {k: v}
+                    tmp_config.setdefault("data_set", {})[k] = v
                 else:
                     tmp_config[k] = v
 
--- a/src/hyrax/data_sets/data_provider.py
+++ b/src/hyrax/data_sets/data_provider.py
@@ -434,16 +434,19 @@ class DataProvider:
         If no ``dataset_config`` is provided in the ``dataset_definition`` dict,
         the original base config will be returned unmodified.
 
-        Example of a dataset definition dictionary:
+        Data request dictionary examples:
 
+        1) Requesting a built-in Hyrax dataset, "MyDataset"
         .. code-block:: python
 
             "my_dataset": {
                 "dataset_class": "MyDataset",
                 "data_location": "/path/to/data",
                 "dataset_config": {
-                    "param1": "value1",
-                    "param2": "value2"
+                    "MyDataset": {
+                        "param1": "value1",
+                        "param2": "value2"
+                    }
                 },
                 "fields": ["field1", "field2"]
             }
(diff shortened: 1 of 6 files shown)


src/hyrax/data_sets/data_provider.py
```diff
@@ -478,17 +517,17 @@ def _apply_configurations(base_config: dict, dataset_definition: dict) -> dict:
...
         if "dataset_config" in dataset_definition:
-            tmp_config = {
-                "data_set": {dataset_definition["dataset_class"]: dataset_definition["dataset_config"]}
-            }
+            tmp_config = {}
+            for k, v in dataset_definition["dataset_config"].items():
+                if k in DATASET_REGISTRY:
+                    tmp_config.setdefault("data_set", {})[k] = v
+                else:
+                    tmp_config[k] = v
```
drewoldag: This is the heart of the work. Everything else in this file is just docstring and comment updates.

aritraghsh09 (commented):
It took me a second to understand exactly what was going on inside `dataset_config`, given the extra level of nesting needed for the external dataset class. I don't know if it's possible to make it 
```
"dataset_config": {
    "external_hyrax_example.galaxy10_dataset": {
                        "channels_first": True,
                },
    "HyraxRandomDataset": {
                    "size": 42,
                },
      }
```
This is more intuitive for me; but as I wrote that it felt like it was a violation of some `.toml` rule 😂

mtauraso (approved):
🚢 it

PR #807, opened 2026-03-23
Tensorboard and mlflow demo notebook

### LINKED ISSUE #771: How to start tensorboard and mlflow from a notebook
Just two lines that show how to crank up Tensorboard and MLFlow from a notebook. 

Also worth pointing out that there would need to be some port forwarding somethingsomething that needs to happen if you're using MLFlow on a remote system.

SamSandwich07:
Adds a notebook detailing how to use TensorBoard and MLFlow


drewoldag (commented):
This is looking pretty good. Just a couple of comments to resolve and it should be good to go.

drewoldag (approved):
Just one comment. Remove that one line and you're good to go.

PR #809, opened 2026-03-23
[AUTO-MERGE] Now passing a data sample with batch size = 1 with initializing a model.

### LINKED ISSUE #795: Pass an entire batch as data_sample to the model
I'm unclear about the best approach here, but let's strive for a consistent state.

Right now when we instantiate a model, it receives a data sample that is a single data point. i.e. not a batch. If it's an image, the shape would be `(C, W, H)` for instance. This data sample is passed through the `prepare_inputs` function as well. Generally `prepare_inputs` handles both `(C, W, H)` and `(batch_size, C, W, H)` equally well. 

However, there are scenarios where those two things are different, for instance, if `prepare_inputs` was transposing dimensions of an input from `(W, H, C)` to `(C, W, H)` - the code will need special logic to handle the case when the data is provided as a batch vs as a single data point. 

So the question at hand is how should we handle this? It seems like the obvious choice would be to always pass a batch when instantiating a model (instead of a single data point) - it is conceivable that a users model architecture _may depend_ on the batch size in the same way that it may depend on the pixel size of input images.

drewoldag:
## Change Description
We now pass the data sample through DataProvider.collate in order to provide a properly shaped data sample to the model to work with. 
Closes #795 

Also updated the models that are currently making use of `data_sample` so that they unpack the shapes correctly.


mtauraso (approved):
Looks good. Only change I'd consider is stylistic,  on the key line where you call `prepare_inputs`, `collate` and `sample_data` it would be great to have them be three different lines.

Sometimes this makes stack traces better, but its fine to merge as is.

SamSandwich07 (approved):

PR #814, opened 2026-03-24
Once inside Hyrax, all "object_id" values are strings.

### LINKED ISSUE #380: Require dataset ids be returned as strings
This keeps cropping up - generally we recommend that ids be returned as strings, but we should just require it and build some unit tests around it. 

The mental model of where the data is and what form it's in is complicated enough - reducing the number of variations (i.e. number vs. string) will help in the long run.

drewoldag:
## Change Description
Use DataProvider as the pinch point to ensure that primary_id_field values are always cast to strings.

This closes out a rather old bug from before the time when we really landed on all ids should be strings. I believe that most built in dataset classes return string ids at this point, but the intent of this change is to ensure that the "object_id" key that is returned on the outside of the data sample is converted to a string.

Note: I intentionally _do not_ convert the internal object id into a string because that seems like it would be surprising to the user.

An example input and return value:
```python
class MyDataset():
    def get_object_id(self, index) -> int:  # note this returns an integer
        return 1  # just an example

data_request = {
    'data': {
        'dataset_class': 'MyDataset',
        'fields': ['object_id'],
        'primary_id_field': 'object_id',
     }
}

# returned_value
{
    'data': {
        'object_id': 1,  # <-- Note: type is the same as what is returned from dataset class get_object_id method.
    },
    'object_id': '1',  # <-- Note: type is coerced to string
}
```

## Solution Description
- `DataProvider.get_object_id()` and the top-level `returned_data["object_id"]` assignment in `resolve_data()` now cast the primary ID to `str`.
- The `returned_data` local variable type annotation and the return type of `resolve_data()` have both been updated to `dict[str, dict[str, Any] | str]` to accurately reflect the mixed structure (nested field dicts per dataset, plus the top-level string `"object_id"` key).
- `DataProvider.ids()` typing and documentation updated to reflect string IDs.
- A regression test was added to verify that the top-level `object_id` is coerced to a string even when the dataset's `get_object_id` method returns an integer.

## Code Quality

---

💡 You can make Copilot smarter by setting up custom instructions, customizing its development environment and configuring Model Context Protocol (MCP) servers. Learn more [Copilot coding agent tips](https://gh.io/copilot-coding-agent-tips) in the docs.

### DIFF
--- a/src/hyrax/datasets/data_provider.py
+++ b/src/hyrax/datasets/data_provider.py
@@ -588,7 +588,7 @@ class DataProvider:
             tensorboardx_logger.log_duration_ts(f"{prefix}/cache_hit_s", start_time)
             return cached_data
 
-        returned_data: dict[str, dict[str, Any]] = {}
+        returned_data: dict[str, dict[str, Any] | str] = {}
 
         for friendly_name, fields in self.requested_fields.items():
             getters = self.dataset_getters[friendly_name]
--- a/src/hyrax/datasets/data_provider.py
+++ b/src/hyrax/datasets/data_provider.py
@@ -568,7 +568,7 @@ class DataProvider:
         """
         return [self.get_object_id(idx) for idx in range(len(self))]
 
-    def resolve_data(self, idx: int) -> dict:
+    def resolve_data(self, idx: int) -> dict[str, dict[str, Any] | str]:
         """This method requests the field data from the prepared datasets by index.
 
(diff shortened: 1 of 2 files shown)


SamSandwich07 (approved):
Looks good to me

mtauraso (approved):
Thinking through edge cases here I'm not finding the reason to not do this.

PR #819, opened 2026-03-25
Remove req that `collate` function to only consume/return friendly name `data`. Add common workflow notebook

### LINKED ISSUE #746: How to write a collate_fn (or how to work with ragged data)
Example notebook showing basic operational setup of a dataset class for ragged data (e.g. time series)

Should feature `collate_fn` heavily.

### LINKED ISSUE #815: Update collate_fn in DataProvider to not rely on "data" friendly name
There are many comment warnings throughout DataProvider:collate that warn about a custom collation function expecting and returning "data" as the friendly name. 

It doesn't seem like that should be a requirement. The DataProvider:collate function should handle whatever friendly name is used in the data request, and the custom collation function in the dataset class should be able to ignore whatever the friendly name is and simply step one layer deeping into the provided list of dictionaries, do the collation, and return final dict without a surrounding 'data'. i.e. not this: `{'data': {<returned_dict>}}`, just this `{<returned_dict>}`.

drewoldag:
## Change Description
Closes #815 

This PR removes the requirement that custom collation functions return a dictionary of collated data nested under a "data" dictionary. 

Now a custom collation function will receive a list of data sample dictionaries, and it should return a single dictionary with the sample data collated as appropriate. 

I've also added a common workflow notebook with an example collation function attached to the Hyrax random dataset that shows what a collation function might look like.


SamSandwich07 (approved):
Notebook makes sense to me

mtauraso (approved):
Looks reasonable to me.

PR #822, opened 2026-03-25
Remove an extraneous sentence from the learning rate scheduler notebook

SamSandwich07:
There was a confusing sentence that I forgot to delete at the end of the learning rate scheduler notebook which was missed on the first pass. I deleted it now.


drewoldag (approved):
Approved

PR #828, opened 2026-03-26
Centering all figure captions

aritraghsh09:
Centers all figure captions. 

Can be overridden per-figure with local directives


drewoldag (approved):
Looks good to me.

PR #830, opened 2026-03-26
converts Prepare to be a subclass of Verb

SamSandwich07:
Changes Prepare to be a subclass of Verb and moves it to the `verbs` directory


drewoldag (changes requested):
I think that two files snuck into the PR that shouldn't be there. Once those are removed it will be good to go.

drewoldag (approved):
Good to go.

PR #848, opened 2026-03-28
Mega Notebook PR

aritraghsh09:
This turned into too big of a PR. Here's what it does 

- Redoes the Getting Started Section:-
    - Splits it into four subsections covering installation, Hyrax Hello World, Astronomy Supervised, and Astronomy Unsupervised. The Astronomy Unsupervised is what used to exist in our old science examples notebook. 
    
- Redoes the Science Section. 
     - Adds a detailed unsupervised discovery notebook and time-domain classification example using light curve data. 

- Also does some aesthetic changes on these pages to make them pleasing overall.


mtauraso (approved):

drewoldag:
Just FYI - I reviewed the readthedocs rendered version

These are more stylistic:
- Swap the first two paragraphs on the landing page. Let the “what is Hyrax” section be the intro paragraph, and then embed what is currently in the first paragraph in a “Why Hyrax?” box. 
- On the installation page:
  - Move the warning about bleeding edge to the “Install from source” section.
  - It’s probably just me, but I’m not a fan of “bleeding edge”, we could use the term “pre-release”?
- I think we could combine the Install with Pip and Install in a conda environment sections. “Install in a virtual environment”, Then the “tip” (most astronomers use conda), then the block creating the env and installing with pip. Then the note about Hyrax officially supports 3.11,…

These seem more important so that we provide consistent guidance:
- Astronomy Supervised page:
  - The note admonition at the top renders oddly (ok, maybe this one isn't super important 😄)
  - In the dataset class, discourage the use of .get() from the config to avoid multiple default values
  - If guard in collate for sequence. It complicates things, but it's what we're advising for now.

PR #849, opened 2026-03-30
Supporting non-index aligned multimodal datasets

drewoldag:
## Change Description

The thrust here is that we will be able to support multimodal data by performing what amounts to a left outer join between datasets.

A primary dataset will define a `primary_id_field` in its data request. This string will be used to join against other datasets.
Secondary datasets _can_ define a `join_field`. Like a `primary_id_field`, this field will return a string id.
For each secondary dataset, `DataProvider` will create a dictionary of `secondary_id_str` : `index`.
Each of these dictionaries will be saved to disk after creation so that they only need to be created once.
When resolving data, `DataProvider` will use the primary dataset id string to look up the corresponding index for the secondary dataset, before requesting data from the secondary dataset.
If a secondary dataset does not have matching data, we return None - this is questionable because it impacts collation.

If a secondary dataset _does not_ define a `join_field`, then we assume that the datasets are index aligned, and process like normal.

Initial work here is done by Claude. Prompted to investigate a way to support non-index aligned datasets.

## Solution Description

- Add `join_field` to `DataRequestConfig` (mutually exclusive with `primary_id_field`) to declare key-based joins for secondary datasets.
- Implement join-map construction, caching, joined lookup behavior in `resolve_data`, and `collate` support for missing secondary rows via `__matched` masks.
- Add join-specific tests and a design spec, plus `.gitignore` entry for join cache artifacts.
- Cap `ThreadPoolExecutor` `max_workers` to `min(4, os.cpu_count())` when building join maps in parallel, to avoid unbounded thread creation when many secondary datasets are configured.

## Code Quality


drewoldag:
I don't want to push this much farther at this point. I know that there's more work to do here, but in the interest of avoid speculative work, I know that I should not try to push ahead.

mtauraso (approved):
This seems basically fine.

I think we might want to add some documentation which specifies how to use it; however, this probably fits best in some sort of multi-modal example for dataprovider, and if we don't have it... we might want to write such a doc as a separate change.

PR #862, opened 2026-04-04
Hexbin visualization tool

### LINKED ISSUE #214: Create a DataSet Interface for visualization
Similar to an interface to splitting, we need an interface between visualization and datasets that surfaces two types of information:
* Metadata relevant to the astronomical objects in the dataset, suitable for lookup in a telescope catalog or an aggregator like vizier/aladin
* A visualization of the input tensor

Generally the fibad (and fibad config) specific parts of this should be factored out, with an eye toward creating an easy and minimal implementation task for someone writing a dataset class.

### LINKED ISSUE #219: Add visualizer support for images or graphs from datasets
Depends on #214

### LINKED ISSUE #314: Implement Permanent Solution for Visualise Module's Usage of Metadata
During investigation of an ordering bug in how metadata returns index calls (https://github.com/lincc-frameworks/hyrax/issues/311); we implemented a quick fix via https://github.com/lincc-frameworks/hyrax/pull/308

However, the following tasks remain:-
1.  `visualize.py` has a lot of code bloat that was added for debugging (e.g., for production deployment, we don't really need to be comparing the object ids we are constructing via the two independent methods). Remove/reduce this code as appropriate. 
2.  `visualize.py` currently uses metadata, to facilitate which we hackily comment out the removal of the `filename` and `filter` column in Lines 270 and 271 of 'fits_image_dataset.py`. A permanent solution for this is described by @mtauraso:- 
     - "We could make the visualizer not use the metatadata interface, put the lines back in, and instead use `FitsImageDataSet`'s `_filter_filename(self, object_id):` to lookup `filenames`. If `_filter_filename` is called as a generator, it'll iterate through the filenames and do the path concatenation that visualizer code is presently doing."

### LINKED ISSUE #507: Decouple visualization and LSSTDataset
PR #356 contains both 2d and 3d visualization work. We should remove the 3d part (Issue #503) and then specifically address the connection between the 2d visualizer and the LSST dataset.

Likely this work should include something like convenience functions that allow for simply visualizing a particular data item given an index.

For instance, implementation of a default "to_image" (name TBD) function on the dataset class that can return or render a visualization of some kind that can be used by the visualization tool for quick display. Additionally/alternatively instead of a default, on the parent class, just helper functions (like what was suggested by @hombit) that a user could choose to use to quickly render images of their data.

### LINKED ISSUE #571: Zooming after selection changes highlighted points
**Bug report**
When you zoom in/out after selecting with lasso, the highlighted points will change.

To reproduce:
1) Display UMAP data with h.visualize
2) Lasso select a small number of data points (~5)
3) Zoom in
4) Note that the highlighted points change as you zoom.
5) Also note that the table _does not_ change

Note that zooming out and back in does not necessarily result in the original points being rehighlighted.

**Environment Information**
WSL, Working with Liams SALT3TRAIN dataset.

### LINKED ISSUE #572: Visualize assumes data points
**Bug report**
```
display_pane, vis_verb = h.visualize(...)
selected = vis_verb.get_selected_df()
```

Here, `get_selected_df` will assume that some data points have been selected. If not, it wil produce a runtime exception. 
There should be a guard in the function that will emit a helpful message.

### LINKED ISSUE #573: Visualize doesn't help identify available metadata columns
**Feature request**
If a metadata column is requested via `h.config["visualize']['fields']` that does not exist, a warning message is logged, but it would be more helpful to display the fields that are actually available.

drewoldag:
Experimenting with holoviews to implement a new hexbin-based visualization tool. Currently only in notebook stage. But handles O(150M) simulated 2d points pretty well!

TODO: 
- [x] Verify that it still works with 100M-300M data points
- [x] Try dynamic color mapping for hexbins

Do we want?
- A light/dark mode
- Specifying a UMAP directory from function call/automatically run the latest UMAP to match behavior of the current visualization?

Main hexbin plot displaying 300M datapoints
Screenshot 2026-04-17 at 3 21 47 PM

Zoomed into a smaller section with a lasso selection (~300k data points selected)
Screenshot 2026-04-17 at 3 23 03 PM

Detail of the selected data points table
Screenshot 2026-04-17 at 3 23 22 PM

### Older screen shots
Screenshot 2026-04-16 at 4 09 47 PM

Screenshot 2026-04-16 at 4 10 12 PM


src/hyrax/verbs/visualize_v2.py
```diff
@@ -0,0 +1,875 @@
...
+        if _primary_ds_name is None:
+            raise RuntimeError(
+                "No sub-dataset in data_request.visualize has a 'primary_id_field'. "
+                "Set 'primary_id_field' on the dataset that contains the UMAP 2D coordinates."
+            )
+        reduced_dim_dataset = self.datasets["visualize"].prepped_datasets[_primary_ds_name]
+        reduced_dim_results = reduced_dim_dataset.__get_all__()
+
+        # ── Build DataFrame from UMAP 2D results ─────────────────────────────
+        points_array = np.array([np.asarray(pt) for pt in reduced_dim_results])
```
drewoldag: Replace with np.stack(reduced_dim_results).astype(np.float32, copy=False)

Seems like this for-loop could be expensive.

src/hyrax/verbs/visualize_v2.py
```diff
@@ -0,0 +1,949 @@
...
+
+                _update_table(
+                    sel,
+                    progress_callback=lambda pct: setattr(_progress_bar, "value", pct),
+                    should_abort=lambda: _gen[0] != my_gen,
+                )
+
+                export_btn.name = "Export table to selected_points"
+                export_btn.disabled = False
+                export_all_btn.disabled = False
```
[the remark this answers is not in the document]
drewoldag: This is done

src/hyrax/verbs/visualize_v2.py
```diff
@@ -0,0 +1,949 @@
...
+                        f.cancel()
+                    return
+                rows[futures[future]] = future.result()
+                if progress_callback and (done % _progress_step == 0 or done == total - 1):
+                    progress_callback(int((done + 1) / total * 100))
+
+            display_df = pd.DataFrame(rows).reset_index(drop=True)
+            _table_df[0] = display_df.copy()
+            selection_table.value = display_df
+            _truncated = len(sel) > max_table_rows
```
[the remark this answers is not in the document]
drewoldag: Fixed

src/hyrax/verbs/visualize_v2.py
```diff
@@ -0,0 +1,949 @@
...
+
+        # The primary dataset (identified by primary_id_field) holds the UMAP coordinates.
+        _primary_ds_name = self.datasets["visualize"].primary_dataset
+        if _primary_ds_name is None:
+            raise RuntimeError(
+                "No sub-dataset in data_request.visualize has a 'primary_id_field'. "
+                "Set 'primary_id_field' on the dataset that contains the UMAP 2D coordinates."
+            )
+        reduced_dim_dataset = self.datasets["visualize"].prepped_datasets[_primary_ds_name]
+        reduced_dim_results = reduced_dim_dataset.__get_all__()
```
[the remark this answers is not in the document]
drewoldag: Done

drewoldag:
I believe that this is ready for review/playing with by other folks. I have resolved the "obvious" performance bugs, and have gone around a few times with agents asking them to look for additional perf improvement opportunities. 

I took another look at the strange way that the HSC images are being rendered. The issue stems from the fact that in HSCDataset when we get an image with `get_image` is ia automatically transformed with `tanh`. Where as in the original visualization code, the images are loaded from disk directly and then passed through `make_lupton_rgb()` without any transforms being applied. 

If we remove the transform within the data_request as shown here, the lupton images look more reasonable:
```python
data_request = {
    "visualize": {
        "results": {
            "dataset_class": "ResultDataset",
            "data_location": "/Users/drew/code/hyrax/docs/pre_executed/results/20260406-134617-umap-2zH3",
        },
        "hsc_data": {
            "dataset_class": "HSCDataset",
            "data_location": "../../data/hsc_8asec_1000",
            "dataset_config": {
                "data_set": {
                    "transform": False,  # <--------Remove the default transform
                }
            }
        },
    }
}

h.set_config("data_request", data_request)
```

Screenshot 2026-05-01 at 1 29 48 PM

Additional thoughts:
1. We should remove the default transformations since those are clearly helpful for some use cases, but are likely unexpected for many. It would probably make more sense to have those transformations and normalizations applied in the `prepare_inputs` function instead of directly on the dataset class. 
2. Once people are happy with the visualization, we can do the work to change the name so that this can be the default visualization and we can mark the old as deprecated.

src/hyrax/verbs/visualize_v2.py
```diff
@@ -0,0 +1,1037 @@
...
+            "Plasma": "plasma",
+            "Magma": "magma",
+            "Cividis": "cividis",
+            "Turbo": "turbo",
+            "Blues": "Blues",
+            "Greens": "Greens",
+            "Purples": "Purples",
+            "Reds": "Reds",
+            "YlOrRd": "YlOrRd",
+            "YlGnBu": "YlGnBu",
```
mtauraso, line 266: I have so much trouble taking these color names seriously.

src/hyrax/verbs/visualize_v2.py
```diff
@@ -0,0 +1,1037 @@
...
+        initial_y_range = (y_lo - y_pad, y_hi + y_pad)
+
+        # ── Adaptive hexbin callback ──────────────────────────────────────────
+        range_xy = RangeXY()
+        _cmap_stream = Params(_cmap_slider, ["value"], rename={"value": "cmap_idx"})
+
+        # Cache the last rasterized element so colormap changes skip re-rasterization.
+        _hex_cache: dict = {"range_key": None, "element": None}
+
+        def make_hexbin(x_range, y_range, cmap_idx=0):
```
mtauraso, line 300: Very clever to reduce the dependence on holoviews axes in the interactive loop.

src/hyrax/verbs/visualize_v2.py
```diff
@@ -0,0 +1,1037 @@
...
+        _prev = {"bounds": (0, 0, 0, 0), "geometry": None}
+
+        def selection_overlay(bounds, geometry):
+            try:
+                return _selection_overlay_impl(bounds, geometry)
+            except Exception:
+                logger.error("selection_overlay raised an exception", exc_info=True)
+                return Rectangles([]).opts(apply_ranges=False) * Polygons([]).opts(apply_ranges=False)
+
+        def _selection_overlay_impl(bounds, geometry):
```
mtauraso, line 756: Clever to separate out selection this way to a separate object.

src/hyrax/datasets/result_dataset.py
```diff
@@ -224,30 +224,54 @@ def __getitem__(self, idx: Union[int, np.ndarray]):
...
+
+        flat_np = data_column.values.to_numpy(zero_copy_only=False)
+
+        # Reshape to original tensor shape
+        tensors = flat_np.reshape(-1, *self.tensor_shape).astype(self.tensor_dtype, copy=False)
 
         # Return single tensor or array of tensors
-        return tensors[0] if is_single else np.array(tensors)
+        return tensors[0] if is_single else tensors
+
```
mtauraso, line 254: I dislike all of this, but understand why its here.

I recall us talking at some point about range retrieval on datasets, which seems more relevant now.

src/hyrax/datasets/result_dataset.py
mtauraso, line 206: I do feel like "these are only needed by viz" should be documented somewhere nearby these function.

mtauraso (approved):
Overall approve. Left some comments, understand this is going in as sort of yet-another-prototype... not the architectural end of visualization.

I think only one of my comments is actionable, and its to add something to a docstring.

src/hyrax/datasets/result_dataset.py
drewoldag: Yeah, it would be cool if we could have some kind of proxy set up so that if a user passes a range/split to a dataset, and the dataset can handle that, we pass it through, but otherwise we intercept and emit requests for indexes one at a time.

drewoldag:
Before merging, I need to remove the three example and utility notebooks.

docs/pre_executed/generate_synthetic_100M.ipynb
drewoldag, line 1: @copilot can you remove this notebook from the PR please?

docs/pre_executed/results_perf_testing.ipynb
drewoldag, line 1: @copilot can you remove this notebook from the PR please?

drewoldag:
@copilot please add the vis_v2_experimentation.ipynb notebook back. Only results_perf_testing.ipynb and generate_synthetic_100M.ipynb should be removed

PR #868, opened 2026-04-09
[Auto-merge] Add NestedPandasDataset

mtauraso:
### Motivation
- Provide a lightweight Hyrax dataset that can read nested Parquet files produced by `nested-pandas` and expose nested columns via dot-notation for downstream inference and examples.

### Description
- Add `NestedPandasDataset` which wraps `nested_pandas.read_parquet`, supports `data_location`, `read_kwargs`, `max_samples`, and dynamically registers `get_<field>` accessors for nested and top-level columns.
- Export `NestedPandasDataset` from `hyrax.datasets` and add a default config block in `hyrax_default_config.toml` for `read_kwargs` and `max_samples`.
- Add `nested-pandas` to `pyproject.toml` dependencies and include an example notebook `docs/notebooks/nested_pandas_time_series_with_hyrax.ipynb` demonstrating usage with nested light-curve data.
- Add unit tests `tests/hyrax/test_nested_pandas_dataset.py` covering reading, nested field getters, `read_kwargs`, `max_samples`, and required `data_location` behavior.

### Testing
- The PR adds pytest unit tests in `tests/hyrax/test_nested_pandas_dataset.py` which create a `NestedFrame`, write a parquet file, instantiate `NestedPandasDataset`, and assert expected getters and length.
- No automated test run was executed as part of this rollout.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d7e3c08390833193074a155e31835e)


src/hyrax/hyrax_default_config.toml
```diff
@@ -361,6 +361,15 @@ number_invalid_values = 0
...
 invalid_value_type = "nan"
 
+
+
+[data_set.NestedPandasDataset]
+# Extra keyword arguments passed directly to the selected nested_pandas reader.
+read_kwargs = {}
+
+# Maximum number of rows to load. Set to false for all rows.
+max_samples = false
```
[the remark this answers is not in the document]
mtauraso: @copilot Please fix with an arg to read_parquet if one exists, otherwise remove the max_samples flag entirely. We will just read the whole file all the time.

mtauraso:
@copilot Can you also fix all the linter errors. Ruff is having some issues with this change.

src/hyrax/datasets/nested_pandas_dataset.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def _value_for_field(self, idx: int, field: str):
+        if "." not in field:
+            return self.nested_frame.iloc[idx][field]
+
+        nested_col, sub_col = field.split(".", maxsplit=1)
+        if nested_col not in self.nested_frame.columns:
+            return self.nested_frame.iloc[idx][field]
+
+        nested_value = self.nested_frame.iloc[idx][nested_col]
```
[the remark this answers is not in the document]
mtauraso: @copilot I would like you to try to fix this function; however, I have totally different feedback.

I think this function is unnecessary because nested pandas supports putting the dot of a nested field inside the bracket operator.

I don't think its necessary for us to parse the field name at all here, and the call to this function may be simply eliminated at the call site with an appropriate oneliner using nested pandas bracket operator.

Please attempt this approach and verify it works with the existing tests.

mtauraso:
@copilot Take a look at `docs/conf.py` There is an exclusion list for classes that we use to avoid certain classes of doc errors.

If you change directory into `docs` and run `make html` you should get the same error that is happening in CI. See if you can alter conf.py to fix the error.

dougbrn (approved):
Looks very cool, approving, but my one comment is that whether by looking at the PR, or trying the branch RTD site, it's not clear to me that the added notebook is navigable on the docs, so I would add links to it where appropriate.

docs/conf.py
```diff
@@ -43,7 +43,14 @@
...
 
 templates_path: list[str] = []
-exclude_patterns = ["_build", "**.ipynb_checkpoints", "archived_notebooks/*"]
+exclude_patterns = [
+    "_build",
+    "**.ipynb_checkpoints",
+    "archived_notebooks/*",
+    # This notebook requires a running Hyrax pipeline and cannot be executed
+    # in the CI/ReadTheDocs build environment.
+    "notebooks/nested_pandas_time_series_with_hyrax.ipynb",
```
drewoldag, line 52: I think that we can just put the notebook in pre-executed, right? At least one cell has to have been run in order for pandoc _not_ to try to rerun it when building the documentation though.

src/hyrax/hyrax_default_config.toml
```diff
@@ -361,6 +361,12 @@ number_invalid_values = 0
 # "nan", "inf", "-inf", "none" or a float value.
 invalid_value_type = "nan"
 
+
+
+[data_set.NestedPandasDataset]
+# Extra keyword arguments passed directly to the selected nested_pandas reader.
+read_kwargs = {}
```
drewoldag: I might recommend making this another subtable in toml since that will map nicely and users won't have to write dictionaries in toml. This new code might be the first instance of this approach. I think that the way we would usually do it is like this:

``` toml
[data_set.NestedPandasDataset]
[data_set.NestedPandasDataset.read_kwargs]
thing_1 = 'foo'
thing_2 = 'bar'
```

I'm open to counter arguments.

src/hyrax/datasets/nested_pandas_dataset.py
drewoldag, line 1: This actually looks really tidy.

src/hyrax/datasets/nested_pandas_dataset.py
```diff
@@ -0,0 +1,53 @@
...
+class NestedPandasDataset(HyraxDataset):
+    """A minimal Hyrax wrapper around ``nested_pandas.read_parquet``."""
+
+    def __init__(self, config: dict, data_location: Path | str | None = None):
+        if data_location is None:
+            raise ValueError("A `data_location` must be provided.")
+
+        self.data_location = str(data_location)
+        settings = config.get("data_set", {}).get("NestedPandasDataset", {})
+        self.read_kwargs = settings.get("read_kwargs") or {}
```
drewoldag: I'm generally worried about the `.get(...)` usage around the config. I would like to avoid it if possible, just so that general users don't get the idea that they can provide _other_ default values.

docs/notebooks/nested_pandas_time_series_with_hyrax.ipynb
```diff
@@ -0,0 +1,141 @@
...
+    "            \"data_location\": str(output_path),\n",
+    "            \"primary_id_field\": \"object_id\",\n",
+    "            # Request only nested columns using nested-pandas dot-notation.\n",
+    "            \"fields\": [\"lightcurve.time\", \"lightcurve.flux\", \"lightcurve.flux_err\", \"lightcurve.band\"],\n",
+    "        }\n",
+    "    }\n",
+    "}\n",
+    "\n",
+    "prepared = h.prepare()\n",
+    "print(f\"Loaded rows: {len(prepared['infer'])}\")"
```
drewoldag, line 77: Minor point - perhaps move the Hyrax instantiation stuff to a new cell, just to separate the synthetic data production from the Hyrax stuff.

docs/notebooks/nested_pandas_time_series_with_hyrax.ipynb
```diff
@@ -0,0 +1,141 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import matplotlib.pyplot as plt\n",
+    "import numpy as np\n",
+    "\n",
+    "time = np.asarray(prepared[\"infer\"][0][\"lc\"][\"lightcurve.time\"])\n",
+    "flux = np.asarray(prepared[\"infer\"][0][\"lc\"][\"lightcurve.flux\"])\n",
+    "flux_err = np.asarray(prepared[\"infer\"][0][\"lc\"][\"lightcurve.flux_err\"])\n",
+    "band = np.asarray(prepared[\"infer\"][0][\"lc\"][\"lightcurve.band\"])\n",
```
drewoldag: These feel kind of wordy, perhaps: 
``` python
data = prepared["infer"][0]["lc"]
time = data["lightcurve.time"]
...
```

src/hyrax/datasets/nested_pandas_dataset.py
```diff
@@ -0,0 +1,53 @@
...
+                "NestedPandasDataset requires the `nested-pandas` package. "
+                "Install it with `pip install nested-pandas`."
+            ) from err
+
+        return npd.read_parquet(self.data_location, **read_kwargs)
+
+    def _all_available_fields(self) -> list[str]:
+        fields = list(self.nested_frame.columns)
+        if hasattr(self.nested_frame, "get_subcolumns"):
+            fields.extend(self.nested_frame.get_subcolumns())
```
drewoldag, line 37: Given what I see in the example notebook, am I correct in my understanding the the getter method names would look like this:

```python
def get_lightcurve.time():
    ...
```

Are dots valid in python method names?

drewoldag (commented):
Mostly minor comments, but one real question about the sub-columns and using dots in the getter method names.

src/hyrax/hyrax_default_config.toml
mtauraso: I think you're right.I'm not sure what to actually put there though.

src/hyrax/datasets/nested_pandas_dataset.py
mtauraso: agree.

src/hyrax/datasets/nested_pandas_dataset.py
mtauraso: So apparently... they are... but you can't call them with normal syntax.

We are metaprogramming so it "just works"

Nonetheless it is awkward
```
dataset = prepared["infer"].prepped_datasets['lc']
dir(dataset)
```
yields:
```
[...
 'get_lightcurve',
 'get_lightcurve.band',
 'get_lightcurve.flux',
 'get_lightcurve.flux_err',
 'get_lightcurve.time',
 'get_object_id',
 ...]
```

And this returns a proper numpy array as one would expect:
```
getattr(dataset, "get_lightcurve.flux")(0)
```

The rough part, is that if we do any translation of "." I think it has to be up in DataProvider if we want people to be able to put "." in their data defnitions, calls to __getitem__ etc...

mtauraso: Apparently the idiomatic python thing is to have "get_lightcurve" return an object where fundamentally ".flux()" works as well as "()" e.g. a callable class instance.

This is another thing that would probably require surgery inside DataProvider to implement....

src/hyrax/hyrax_default_config.toml
drewoldag: I think that if you leave it as just a subtable with no key/values that you'll get an empyt dict like what you're after. 

```toml
[data_set.NestedPandasDataset]
[data_set.NestedPandasDataset.read_kwargs]
# message here saying to add nested kwargs here if needed.
# or a commented out, common kwarg someone might use? 
```

src/hyrax/datasets/nested_pandas_dataset.py
drewoldag: Ok, so all that being said, are we basically good to go? Given that internal hyrax machinery will be used to call the getters and we don't need to do anything special to to the metaprogramming to make e.g. ".flux()" act like "()"?

mtauraso: Yeah, I'm gonna merge it now.

Issue #428, opened 2025-09-22
Investigate the use of `lance` file format for storing InferenceDataset output


drewoldag:
Currently model output is saved to disk with InferenceDatasetWriter that produces several .npy numpy files along with a manifest. 

We would like to move to a more familiar file format, and .parquet has been considered, but the random access read times are not very good. 

Our understanding is that the lance (`pylance`) data format may support performant random access reads. 

The output for this ticket should be:
- Create timing benchmarks for writing and random access read times for InferenceDatasetWriter and InferenceDataset.
- Create a jupyter notebook to compare performance of the current .npy implementation against a lance implementation.

drewoldag:
Assigning this to @gitosaurus just in case it's helpful to have an issue for tracking any of the investigation.

gitosaurus:
With the merging of #699, this is mostly done.  LanceDB is now the default in Hyrax, and we have a `ResultDataset` class that is used by default.  Not closing this issue yet until we're sure that the conversion script and deprecation plan are complete.

One thing to note is that using `async` / `await` with LanceDB, while possible, led to a complaint when run in a Jupyter notebook, so turning them on isn't trivial.  However, Lance uses Rust async I/O under the hood and appears sufficiently performant in its most straightforward usage at present.

PR #872, opened 2026-04-13
Add script to convert .npy inference results to Lance format

drewoldag:
Adds scripts/convert_results.py with convert() and verify() functions
to migrate legacy InferenceDatasetWriter .npy output to the new
ResultDataset Lance format. Verifies every row bitwise after conversion.

Also adds tests/hyrax/test_convert_results.py covering round-trips,
multi-dimensional tensors, in-place vs separate output dir, and guard
conditions (missing batch_index, already-converted directory).

Both files are marked for removal once the .npy format is fully
deprecated (tracking issue #428).

https://claude.ai/code/session_01V22KAdUXExCws1AvLH7WP8


aritraghsh09:
@drewoldag I see there is a test, but have you tested this out on an old `.npy` file that you have from Hyrax?

tests/hyrax/test_convert_results.py
```diff
@@ -0,0 +1,258 @@
...
+        all_ids.extend(ids)
+        all_batch_nums.extend([batch_num] * len(ids))
+
+    # Write batch_index.npy sorted by id (matches InferenceDatasetWriter behaviour)
+    index_dtype = np.dtype([("id", id_dtype), ("batch_num", np.int64)])
+    batch_index = np.zeros(len(all_ids), index_dtype)
+    batch_index["id"] = all_ids
+    batch_index["batch_num"] = all_batch_nums
+    batch_index.sort(order="id")
+    np.save(result_dir / "batch_index.npy", batch_index, allow_pickle=False)
```
mtauraso: Mildly concerning to me that this test fixture doesn't use InferenceDatasetWriter directly...

Seems like tests depending on it would be much more valid if id did, and IIRC InferenceDatasetWriter doesn't need much in the way of dependencies.

Is there a hard blocker to using InferenceDatasetWriter this way?

scripts/convert_results.py
```diff
@@ -0,0 +1,273 @@
...
+        raise RuntimeError(
+            f"Row count mismatch after conversion: wrote {total_written} records "
+            f"but expected {total_expected} from batch_index.npy. "
+            f"One or more batch files may be missing or corrupt."
+        )
+
+    print(f"Conversion complete — {total_written} records written to {output_dir / LANCE_DB_DIR}")
+
+
+def verify(input_dir: Path, output_dir: Path) -> None:
```
mtauraso, line 141: This seems kinda reasonable as a spot-check... but its also used in the tests.

The test I want to see (but isn't written afaict) is one where convert an inference dataset written with `InferenceDatasetWriter` and verify that when `ResultDataset` reads either the pre or post-convert dataset it gets the same values in the same order.

Not reimplementing the read/write code for the batched numpy files is a *feature* of this test not a bug.

Reading this, it looks like the design cuts against the spirit of the test I have in mind.

I also have questions about why verify is re-implementing the read code for batched numpy files. I understand such a reimplementation *might* be more performant in a migration scenario, as a spot check... but it seems like there's another way to write this function that would create a lot more confidence in the correctness of the migration.

mtauraso:
Overall my feedback rounds to "Why are we reimplementing batch numpy read/write code?"

I acknowledge that the code written is probably correct; however, it seems like if the goal is filesystem conversion correctness as assessed by parity at the level of the dataset abstraction.... we should simply be using the existing dataset abstractions a lot more than this code actually does.

Happy to be told I'm missing something here about what would happen if we did more code re-use.

drewoldag:
@aritraghsh09 yes, I tested this locally by checking out the branch you've been working on, generating some result files and then converting to them with this script.

drewoldag:
@mtauraso it's a good suggestion to use InfereceDatasetWriter for the tests, I went ahead and made that update. And added a test to make sure that index alignment is maintained as well. 

Using InferenceDataset to read the .npy files in `convert` is maybe a different story. To create an instance of InferenceDataset, we call setup_dataset on the original dataset, which is _probably fine_ assuming that we have access to the original dataset that created the inference output, but it might be missing too, and if that's the case, I think that InferenceDataset won't fully instantiate, right? 

That's why I was leaning toward the more naive approach of just reading the batch files directly. 

But, to reciprocate, I'm happy to be told that it's silly and to just use to the InferenceDataset for reading :)

scripts/convert_results.py
drewoldag: I updated the tests so that they use InferenceDatasetWriter to output the test .npy files. I also added a test that is meant to verify the insertion order is the same between the original and converted datasets. 

As mentioned in the comment in the conversation page, the InferenceDataset(Reader) requires the original input dataset to be present. While it's likely going to be available, if it's not, then I think that InferenceDataset doesn't fully initialize. 

Again, tell me I'm wrong if I'm misunderstanding here :)

mtauraso:
I had to read a little bit to convince myself, but I think InferenceDataset is pretty easy to set up in a testing context, so long as you pass the path you want it to read.

There's an example in test_inference_dataset.py the `inference_dataset` fixture does repeated write/read cycles so tests can analyze the stability of the read/write operations. The setup of InferenceDataset there is pretty minimal, and I think would work here with only minor adjustments.

The key thing is that InferenceDataset simply doesn't have a lot of configuration. You do need an original dataset for metadata, but you can use HyraxRandomDataset, and whether that matches up for testing doesn't really matter for this test because the original dataset is only used in the metadata codepaths in inference dataset, not any of the read/lookup cood.

mtauraso (approved):

drewoldag:
ok, before I merge this I'll try again with InferenceDataset for reading in the test.

PR #878, opened 2026-04-15
Add versioned config schema migrations

### LINKED ISSUE #810: data_set table in config rename to dataset / Versioning config
Coming out of #802 We noticed that there is still a `data_set` section in the config despite the canonical naming being dataset all one word.

Changing this risks invalidating prior configs, which users depend on for reproducibility. 
We have also in the past had breaking config changes, with similar effects.

I propose that we introduce a version key under [general] that is set in the default config. It's a monotonically increasing number.  When we make major breaking config changes, we flag them in hyrax under whatever number they changed in.

The number would be an int starting at 1, with configs that don't define it being implicitly version zero (everything up until now).  This gives us a forward compatibility path and a way to keep reading people's projects.

drewoldag:
Introduces a top-level `config_version` scalar in hyrax_default_config.toml
plus a migration registry in src/hyrax/config_migrations.py so that future
config schema changes (renamed tables, moved keys, value transforms) can
evolve without breaking users on older config files. On load, ConfigManager
walks the registered migrations from the user's declared version up to
CURRENT_CONFIG_VERSION before the existing merge, emits a DeprecationWarning
per transform that actually fires, and refuses configs from a newer Hyrax
with a clear pip install -U hint.

Additionally, a registry of name changes is generated from all of the defined migrations. If a user calls `h.set_config(key, value)` and we find `key` in the list of name changes, we'll _warn_ the user, but will allow the set_config to continue. Here I'm assuming that it's better to warn and not block because the user may have some special reason for doing what they are doing.

The v1 -> v2 migration retires the long-dangling `[model_inputs]` ->
`[data_request]` rename: the default config's dead deprecation comment is
removed, the `model_inputs` entry drops out of PYDANTIC_VALIDATED_KEYS, and
the verb-registry fallback is gone. Users whose configs still reference
`[model_inputs]` now get a one-time DeprecationWarning and keep working.

Documentation in HYRAX_GUIDE.md, CLAUDE.md, and the configuration system
reference describes the developer workflow for adding a new migration.

https://claude.ai/code/session_01Jop15LHgatYCecyyeNfWu4


src/hyrax/verbs/verb_registry.py
```diff
@@ -76,7 +75,7 @@ def validate_data_request(self) -> None:
         if not self.REQUIRED_DATA_GROUPS and not self.OPTIONAL_DATA_GROUPS:
             return
 
-        data_request = self.config.get("data_request") or self.config.get("model_inputs")
+        data_request = self.config.get("data_request")
         if not data_request:
             return
```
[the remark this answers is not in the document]
drewoldag: Thanks for flagging. This falsy-skip is pre-existing behavior — this PR only touched the line to drop the `model_inputs` fallback. An empty `[data_request]` table is already a user error surfaced elsewhere in the stack (dataset construction fails without required fields like `primary_id_field`), so tightening this to `data_request is None` would be a behavioral change outside the scope of config versioning. Leaving as-is for this PR; happy to revisit in a follow-up if the team wants stricter enforcement here.

---
_Generated by [Claude Code](https://claude.ai/code)_

drewoldag:
The first draft of this was looking pretty good. I've requested some changes to alert the user when they try to `h.set_config(...)` for a deprecated value. Currently choosing to just warn, but not block the user. 

Also trying to simplify the dev burden by reducing that amount of code that is needed when writing a migration.

drewoldag:
TODO: Review the existing code for instances of `model_inputs` and remove those since the config versioning should be working now.

drewoldag:
The core part of this is now in place, thanks for the early feedback @mtauraso - I've moved the migrations into a separate directory. Also the file and function names are no longer important, all that matters is getting the `from_version` tag in the decorator correct. (Though reasonable file names will help make it easier to know what the `from_version` should be 😄) 

I have yet to remove the old `model_inputs` code references. I'll do that in my next commit.

src/hyrax/config_migrations/migrations/001_rename_model_inputs_to_data_request.py
mtauraso, line 1: Not to nit too much, but a load bearing part of the activerecord-style migration files is both the version-ordered enumeration (which is present) as well as descriptive naming of migrations.

I'd want this to be called like "01_rename_model_inputs.py" and similar for the function names within.
We should exercise the flexibility this pattern affords us with naming :D

src/hyrax/hyrax_default_config.toml
```diff
@@ -1,3 +1,7 @@
+# DO NOT MODIFY config_version MANUALLY!!!
+# Hyrax will migrate older configs forward on load — do not edit this by hand.
+config_version = 2
```
mtauraso: There's a potential programming failure here where someone correctly adds a migration... but fails to increment this value.

The config system needs to know this at runtime, and this value should be output whenever we write a config, so we can read it in from a user config in the process of migration. I wonder if we can derive this value at runtime in a consistently correct manner, and then put it all the places it needs to go and thereby eliminate this type of oversight error.

HYRAX_GUIDE.md
```diff
@@ -209,6 +209,36 @@ defaults, not by Pydantic schemas.
...
+and self-registers via the `@migration_step` decorator. `CURRENT_CONFIG_VERSION`
+is auto-derived from the highest registered migration — do not bump it manually.
+
+**When you rename or restructure a config key, you must:**
+
+1. Create `src/hyrax/config_migrations/vN_to_vN_plus_1.py`. Decorate the
+   migration function with `@migration_step(from_version=N, key_renames={...})`.
+   Import the decorator and helpers (`rename_table`, `move_key`) from
+   `hyrax.config_migrations._machinery`.
+2. Add `from hyrax.config_migrations import vN_to_vN_plus_1  # noqa: F401` to
```
mtauraso: If we're going with descriptive names we should probably tell the robots to do that too.

mtauraso (approved):
Overall approve. I have some comments, but they are things we can address down the road if needed.

src/hyrax/config_migrations/migrations/001_rename_model_inputs_to_data_request.py
drewoldag: Done, renamed this file and updated the docstrings and README

src/hyrax/hyrax_default_config.toml
drewoldag: Removed from `hyrax_default_config`

HYRAX_GUIDE.md
drewoldag: Updated.

PR #881, opened 2026-04-21
adding umap loading logic and tutorial

### LINKED ISSUE #758: Umap should be able to save and load it's model
It turns out we _do_ save the reducer that is trained in the `umap` verb, but we don't expose a way to specify a particular saved reducer to use.

And since UMAP is very configuable, we should be saving a copy of the config along with the rest of the output.

Graciaaa3:
## Change Description
Closes #758 

## Solution Description
Adding a new argument that takes in the path to a pre-existing umap reducer model. This allows users to reuse models instead of creating new one each time. 

Updated the tutorial notebook `using_umap.ipynb` to include this addition. 

Expanded the CLI argparse list of CLI arguments to include the new argument.  

## Code Quality


src/hyrax/verbs/umap.py
```diff
@@ -99,17 +111,48 @@ def _run(self, input_dir: Union[Path, str] | None = None):
...
-        # Fit a single reducer on the sampled data
-        self.reducer.fit(data_sample)
-        self._log_memory_usage("After fitting umap")
-
-        # Save the reducer to our results directory
-        if self.config["umap"]["save_fit_umap"]:
-            logger.info("Saving fitted UMAP Reducer")
-            with open(results_dir / "umap.pickle", "wb") as f:
-                pickle.dump(self.reducer, f)
+        if model_path is not None:
```
drewoldag: One additional improvement that would be nice here is to have a new config value added to `hyrax_default_config.toml`. That would allow as user to specify a model_path in the config to be picked up here. 

Basically the logic just before the `if` statement would be something like:
```python
if model_path is None: 
    model_path = self.config['umap']['model_path']

if model_path:
    # check validity of the path
    ... # the rest of the current code.
```

drewoldag (commented):
One quick comment on this - happy to chat about it some more in person :) Overall looking pretty good. I need to take a closer look at the notebook, but the umap.py file is pretty close!

docs/pre_executed/using_umap.ipynb
```diff
@@ -333,6 +329,166 @@
...
+   "cell_type": "markdown",
+   "id": "21370ecd",
+   "metadata": {},
+   "source": [
+    "## 6. Loading pre-exisiting UMAP model\n",
+    "\n",
+    "\n",
+    "When running the `umap` verb, the trained reducer is automatically saved as *umap.pickle* file as part of the results. We can reuse a previously trained UMAP reducer model instead of creating a new one from configuration to ensure consistent results and faster workflow.\n",
+    "\n",
+    "To do this, pass the saved model path via the `model_path` argument:="
```
drewoldag:
```suggestion
    "When running the `umap` verb, the trained reducer is saved in the results directory as *./umap.pickle*. We can reuse a previously trained UMAP reducer model to ensure consistent results and faster workflow.\n",
    "\n",
    "To do this, pass the saved model path via the `model_path` argument:="
```

docs/pre_executed/using_umap.ipynb
drewoldag, line 1: There is a table at the end of this notebook that lists the config values that can be set. I would recommend adding `model_path` to that table.

drewoldag (approved):
Overall this looks good. I had a couple of comments on the notebook. If you don't mind addressing those, I would appreciate it. I'll go ahead and mark this as approved so that once the notebook comments are addressed, you can go ahead and merge.

PR #882, opened 2026-04-22
Speed up AI container startup.

mtauraso:
### Motivation
- Avoid installing Hyrax's full dependency set inside constrained agent/container environments by providing a prebuilt Python venv artifact.
- Centralize and reuse a single environment artifact produced from `main` to ensure consistent dev dependencies across agent sessions.
- Clean up older artifacts to prevent clutter and reduce storage/lookup ambiguity.

### Description
- Add GitHub Actions workflow `/.github/workflows/build-agent-env-artifact.yml` that sets up Python 3.11, creates a venv, runs `pip install -e '.[dev]'`, runs `pip check`, packs the venv into a tarball, uploads it as artifact `hyrax-agent-venv-main`, and deletes older artifacts with the same name.
- Update `agent_scripts/claude_code_web_setup_container.sh` to download the artifact zip via the GitHub Actions API, extract the contained tarball into a configurable `VENV_DIR`, activate the venv, and install the repo in editable dev mode; add `set -euo pipefail` and robust repo root detection.
- Update `agent_scripts/codex_setup_container.sh` with the same artifact-download-and-extract flow, activation of the prebuilt venv, and `set -euo pipefail`, and make artifact repo/name and venv dir configurable via `HYRAX_*` env vars.
- Remove previous local venv creation steps and direct pip installs in these scripts so they rely on the prebuilt artifact instead.

### Testing
- No automated tests were executed as part of this PR; CI has not been triggered by this change yet.
- The new workflow includes a `pip check` step which will validate installed package dependencies when the workflow runs on CI.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e907fdf1808331aa3e6cc8405bcd5d)

### DIFF
--- a/.github/workflows/build-agent-env-artifact.yml
+++ b/.github/workflows/build-agent-env-artifact.yml
@@ -26,8 +26,11 @@ jobs:
       - name: Build conda env
         shell: bash -l {0}
         run: |
-          conda install -y conda-pack pip
+          conda install -y conda-pack pip setuptools
           pip install '.[dev]'
+          conda list
+          conda install -y setuptools
+          conda list setuptools
           pip check
 
       - name: Pack conda env
--- a/.github/workflows/build-agent-env-artifact.yml
+++ b/.github/workflows/build-agent-env-artifact.yml
@@ -1,7 +1,7 @@
 name: Build agent conda environment artifact
 
 on:
-  pull_request:
+  push:
     branches: [ main ]
   workflow_dispatch:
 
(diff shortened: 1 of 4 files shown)


drewoldag (approved):
This seems pretty cool, can we use the packed conda environment for anything else? build-docs, or pre-commit? I suppose the risk is that the deps have changed and the packed venv might be out of date with pre-commit or build-docs?

Either way, this part seems quite useful.

PR #893, opened 2026-04-24
Fix test_torch_load_with_map_location on macs.

mtauraso:
Issue was that nn.Module `load_state_dict` uses the properties of the model rather than the properties of the incoming state dict to determine device affinity by default.

With `assign=True` `load_state_dict` uses the properties of the incoming dict, which is what we want, since the state dict has already been deserialized into device memory by `torch.load` using `map_location`


drewoldag (approved):
Looks good

PR #900, opened 2026-04-25
Fix claude code and codex startup scripts.

mtauraso:
More robust setup scripts for claude code and codex venvs.  Current version on main is broken, and only appeared to work due to caching of container environments.

If you want to try this out, follow the steps in the README on your chosen platform and then make sure your branch is set to this PR branch. First startup will be a cold-start, but subsequent startups will be fast. You can see the venv on $PATH, and the agent should be able to do virtually any dev command requiring our python dependencies.

Cold start will take about 5 minutes, Warm starts give 30-90s startup times that seem more dependent on the container and LLM backends than anything we're doing.

Both Codex and Claude code are robustly *in* the venv by default.


drewoldag (approved):
ok, give it a shot

PR #905, opened 2026-04-30
Add `over_write` to `h.set_config`

### LINKED ISSUE #821: h.set_config(..., ..., over_write=True)
**Feature request**
The `set_config` Hyrax method is explicitly meant to merge the new setting into the existing settings. Most of the time this is exactly what you want to do. 

However, there are times, such as when setting the data_request, that it's not useful to merge the config, instead it would be more useful to simply overwrite it. 

Here's the current behavior:
``` python
data_request = {'train': 'data_0': {...}}
h.set_config('data_request', data_request)

print(h.config['data_request'])
>> data_request{ train{ 

data_request = {'train': 'data_1': {...}}
h.set_config('data_request', data_request)

print(h.config['data_request'])
>> data_request{ train{ data_0 { ... }, data_1 { ... } } }
```

The expectation would be that train would contain only `data_1` at the end of the process, but we see that data_1 and been merged into `train` data_group.

Thus I propose that we could implement a `over_write` flag in set_config. 

```
def set_config(dotted_string: str, new_value: any, over_write=False)
```
When `over_write==True` the value at `dotted_string` in the config is completely replaced by the `new_value`. If `over_write==False` then we use the current logic.

Graciaaa3:
## Change Description
Closes #821 

## Solution Description
Adding `over_write=False` to `set_config()` to support overwriting the highest level existing values provided. Already update the docstring to address the change. I will add some tests to test the new parameter.

Currently I kept the workflow logic the same as before but added this new parameter to all the function gets called along the path. `merge_configs` will skip recursive calls when it was set to True, so directly setting the value to what is given.

## Code Quality

### DIFF
--- a/docs/notebooks/config_basics.ipynb
+++ b/docs/notebooks/config_basics.ipynb
@@ -154,7 +154,9 @@
    "id": "1b550a85",
    "metadata": {},
    "source": [
-    "This is useful when you want to completely replace a config section instead of partially updating it, especially for complex nested settings where old fields should not be preserved. See documentation for more information: :meth:`hyrax.hyrax.set_config`."
+    "This is useful when you want to completely replace a config section instead of partially updating it, especially for complex nested settings where old fields should not be preserved. \n",
+    "\n",
+    "See the documentation for more information: [set_config](https://hyrax.readthedocs.io/en/latest/autoapi/hyrax/hyrax/index.html#hyrax.hyrax.Hyrax.set_config)"
    ]
   },
   {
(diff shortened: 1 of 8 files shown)


drewoldag (commented):
This is looking good. Just to summarize the additional requested changes: 

1. Rename `ConfigManager.set_config` -> `ConfigManager._set_config`
2. Add a super tiny example of using `over_write = True` to the config_basics.ipynb notebook after the "Updating config values" section.

drewoldag (approved):
This is looking good to me. Nice work.

Issue #841, opened 2026-03-27
Per column collate functions.


mtauraso:
This work comes out of an observation about the current `collate()` interface hyrax exposes to datasets. Since a dataset's `collate` function is always called, it must be resilient to the user only specifying some of the columns, meaning that the `collate` function needs to essentially if-guard every column it processes, and it must process all columns the dataset defines.

Given that most astronomers only define a `collate` function for one or two columns on their dataset, the idea would be to allow a dataset to define `collate` functions per-column and fall back on default collation for all the other columns.

For example:
```
class MyDataset:
...
  def get_light_curve(idx):
    ...
  def collate_light_curve(samples:list[np.ndarray]):
    ...
``` 

One could also imagine default collate functions for particular types of data built into hyrax, and accessed with decorators like this:

```
class MyDataset:
...
  @collate_as_light_curve
  def get_light_curve(idx):
    ...
```
But we need to make sure that we know enough about the data formats astronomers use to de-risk this decorator step, so we are only building the per-field version on this ticket.

This work is just step two on the overall path:
1. One giant collate function (what we have now)
2. Support for per-field collate functions that have names that mirror the getters. get_light_curve + collate_light_curve 
3. Decorators once we start to detect a pattern.

Each of these should be easily backward compatible. And could coexist within Hyrax.

However, 1 and 2 shouldn’t coexist in the same dataset class, and this should be enforced some time during the loading of the dataset.

SamSandwich07:
Solved with #917 and #921

PR #917, opened 2026-05-02
Support for field-dependent collate functions

SamSandwich07:
Adds support for field-dependent collate functions, addressing step 2 in #841. Users define `collate_{field}` akin to `get_{field}` and `data_provider.py` handles constructing a dataset-level collate function from the field-level collate functions


mtauraso (approved):
This looks good to me.

Maybe for a follow up, we should think about how this interacts with trace mode.

Given the bookkeeping of the functions in DataProvider, there should be a relatively simple way to make the per-field collate functions available in traces.

SamSandwich07:
One question before I merge, should there be a RuntimeError thrown if the statement `if all(s == shapes[0] for s in shapes)` is false on line 1044? Even in the original default logic, if it's false then the code moves on to
```python
if isinstance(values, list):
    batch_dict[friendly_name][field] = np.array(values)
```
but I think `values` is always a list so then if the arrays in `values` aren't all the same length, this just returns a numpy array of lists, something like `array([list([1, 2]), list([1, 2, 3])], dtype=object)`. That's why I removed the redundant `if isinstance(values, list)` code when I moved this logic to a function, but this same problem still exists so it feels like if the arrays in `values` are different lengths then we should return an error.

On that note, why does `if isinstance(values, list)` even exist if `values` is always going to be a list?

src/hyrax/datasets/data_provider.py
```diff
@@ -484,12 +493,26 @@ def prepare_datasets(self):
...
                     )
+                field_collate_fn = getattr(dataset_instance, f"collate_{field}", None)
+
+                # error if dataset collate is defined along with field dependent collate
+                # or if field dependent collate is defined for some fields and not others
+                if callable(field_collate_fn):
+                    if dataset_collate:
+                        raise RuntimeError(
+                            f"Dataset '{friendly_name}' declares both global collate function "
+                            f"and field-dependent collate function for field '{field}'"
```
drewoldag: I would be ok with expanding this runtime error message even more to say that Hyrax expects one of 1) no collation fuctions 2) dataset `collate` function or 3) field-specific collate functions. Feel free to chose better wording. 

It make sense to add or expand our docs, specifically this one: https://hyrax.readthedocs.io/en/stable/notebooks/custom_dataset_collation.html 

So we could reference that doc page as well in the this error message.

src/hyrax/datasets/data_provider.py
```diff
@@ -484,12 +493,26 @@ def prepare_datasets(self):
...
+                        )
+                    self.field_collate_functions[friendly_name][field] = field_collate_fn
+                elif not dataset_collate:
```
drewoldag: can this just be `else`?

drewoldag (commented):
This seems good. I would like to see some unit tests for this one now, since there's a good amount of logic here in the `DataProvider.collate` function.

mtauraso:
I think there's sort of two questions here:

1) Why was the previous code written as it was?
I don't have a super-clear answer, but we're iterating over `batch_dict` and changing it in the original. I think the idea was to only make changes if needed. The contract at this point is that we are outputting numpy types, so while it works to re-write everything as numpy... why bother if it's already numpy?

2) What error behavior should we have when you pass in data of different lengths, but 1d?

In generally we should try to give helpful error messages. The error case as written before and after the diff in default collate covers the fatal error of "We can't collate this at all". The more subtle "You are passing arrays of different length in 1d" condition isn't really handled cleanly.

I'm not sure I have a solid opinion on what's best for the user around number 2. Warning them seems reasonable. I'm  pretty sure we shouldn't have a fatal error though.

Overall a lot of the code in this function suffers because we don't have a clean way to ensure that the dataset is meeting its contract. I think a lot of these concerns become simpler if we can check a dataset, and know it is outputting a particular way.

drewoldag (approved):
Ok, that test seems like a reasonable starting point to ensure that the collate_* functions are picked up and executed as expected.

Issue #919, opened 2026-05-08
Update and expand documentation for per-field collation


drewoldag:
We should expand our current documentation to explain per-field collation functions, and change or explain in the science example notebooks how custom collation can be used. 

Here are a couple of specific locations, there may be others. 
https://hyrax.readthedocs.io/en/latest/notebooks/custom_dataset_collation.html
https://hyrax.readthedocs.io/en/latest/pre_executed/supervised_lightcurve_transients_gs.html#Define-a-custom-dataset

This topic might be complex enough that it could use a couple of diagrams to clarify what's going on. Happy to brainstorm on this one.

SamSandwich07:
Solved with #921

PR #920, opened 2026-05-12
Update custom_data_collation.ipynb to show how to use field-level collation

SamSandwich07:
Update the `custom_data_collation.ipynb` notebook. One step in addressing #919.

### DIFF
--- a/docs/notebooks/custom_dataset_collation.ipynb
+++ b/docs/notebooks/custom_dataset_collation.ipynb
@@ -99,7 +99,7 @@
     "`if \"image\" in samples[0]`). Without these guards the collate function will\n",
     "fail whenever a user omits a field from their request.\n",
     "\n",
-    "In production the method would live in the dataset class file. Attaching it\n",
+    "In production, the function should be a method on the dataset class. Attaching it\n",
     "dynamically here, as shown below, is convenient for prototyping in a notebook."
    ]
   },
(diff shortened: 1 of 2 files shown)


docs/notebooks/custom_dataset_collation.ipynb
```diff
@@ -163,6 +163,78 @@
...
+    "either field-level collate functions **OR** a dataset-level collate function.\n",
+    "Defining both will cause Hyrax to throw an error when training is attempted.\n",
+    "\n",
+    "Hyrax defines default behavior for fields which do not have a collate function defined.\n",
+    "The default behavior simply attempts to stack the given data into a single `numpy` array,\n",
+    "without adding padding or masking. While this works well for scalar data -- such as the\n",
+    "`object_id` and `label` fields, which do not have a custom `collate_*` function defined\n",
+    "in the example below for precisely this reason -- or data where a uniform shape is already\n",
+    "guaranteed, desired results may not be achieved in other scenarios; thus it is highly\n",
+    "recommended to define a collation function for nontrivial fields.\n",
```
drewoldag: This section feels a bit wordy. The 3rd sentence should probably be broken up or simplified.

I would also recommend that the last section "thus it is highly ... for nontrival fields" could be reworded as something like "Hyrax exposes the `collate_*` API  explicitly to handle complex or non-uniform data fields." (Or something like that)

docs/notebooks/custom_dataset_collation.ipynb
```diff
@@ -163,6 +163,78 @@
...
+    "\n",
+    "As with the dataset-level `collate` function, in production the field-level methods would live in the dataset class file."
```
drewoldag: Minor point: I would recommend using a term other than "live in". 
```suggestion
    "As with the dataset-level `collate` function, in production the field-level methods would included in the dataset class file."
```

docs/notebooks/custom_dataset_collation.ipynb
```diff
@@ -188,6 +261,28 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "delattr(HyraxRandomDataset, \"collate\")\n",
+    "dataset = h.prepare()\n",
+    "\n",
+    "# Access the \"train\" data group for clarity in the next steps\n",
+    "train_dataset = dataset[\"train\"]"
+   ]
+  },
```
drewoldag: Combine this cell into the one just before. i.e. call `delattr(...)` first. Explain in the note above the cell that you're removing `collate` because otherwise Hyrax would raise an error. 

This should simplify the notebook slightly and remove the issue with the expected error causing doc builds to break.

drewoldag (commented):
Left a couple notes, otherwise looking pretty good. Once these are resolved, I'm happy to approve. Thanks for putting this together!

SamSandwich07:
#921 includes the changes in this PR so once that one is approved I will close this PR.

drewoldag:
I see the changes that were made to this notebook in PR #921, and they look good to me.

PR #925, opened 2026-05-19
Setting torch to have a version floor of 2.9.0

mtauraso:
We need this so that hash_tensor is always available for trace mode


drewoldag (approved):
Approved. Looks like 2.9 was released back in Oct - that probably aligns with when they dropped support for Python 3.10. This pinning seems like a good idea for that reason as well.

PR #932, opened 2026-05-26
Implement reduce_dimensions

Graciaaa3:
## Change Description
Closes #859 

## Solution Description
This PR introduces the initial framework for a `reduce_dimensions` verb to support multiple dimensionality reduction algorithms with Hyrax.

### Current Progress
Added the `reduce_dimensions` verb.
Introduced the `ReductionAlgorithm` subclass structure to provide a generalized abstraction layer for dimensionality reduction methods. Migrated the existing UMAP implementation into the new `reduce_dimensions` framework while preserving current functionality and behavior.

### Next Steps
- [x] Implementing support for additional dimensionality reduction algorithms: t-SNE and PCA.
- [ ] Adding and updating documentation for the new API and configuration structure.
- [x] Adding and updating tests for `reduce_dimensions` and newly supported algorithms.

## Code Quality


src/hyrax/hyrax_default_config.toml
```diff
@@ -442,6 +442,55 @@ vector_size = 64
 inference_dir = false
 
 
+[reduce]
+# Save the fitted reducer model as a pickle file 
+save_fit_model = true
+
+# Name of the reduction algorithm to use
+algorithm = "umap"
```
drewoldag: Minor point - I would make this the first key under the `[reduce]` table here. My thought is that this is the most important thing that people would want to change, so it should be at the top.

src/hyrax/verbs/reduce_dimensions.py
drewoldag, line 1: How can we expose to the user the available reduction algorithms?

Documentation is good, but it requires that they leave the notebook or console for the browser. 
In setup_parser, we could read off the keys from the registry and add that to the `help` string.
We could also implement a simple function the Hyrax class to print them - this is what we do for verbs and dataset classes, and models. 

Doesn't need to be settled in this PR, but worth thinking about and perhaps creating a follow up issue for it.

src/hyrax/verbs/reduce_dimensions.py
```diff
@@ -0,0 +1,159 @@
...
+        input_dir: Union[Path, str] | None = None,
+        model_path: Union[Path, str] | None = None,
+    ):
+        """
+        Run dimensionality reduction on a dataset
+
+        This method loads the latent space representations from an inference run,
+        samples a subset of data points, flattens them if necessary,
+        and fits a dimensionality reduction model. If a model path is provided,
+        it loads the model instead of fitting a new one. Then the reducer will
```
drewoldag: Pick your own wording here, but it would be good to clarify that not all algorithms will support saving and loading trained models. 

```suggestion
        it loads the model instead of fitting a new one, given that the algorithm supports this. Then the reducer will
```

src/hyrax/verbs/reduce_dimensions.py
```diff
@@ -0,0 +1,159 @@
...
+        """See run()"""
+        from hyrax.config_utils import create_results_dir
+        from hyrax.datasets.result_factories import create_results_writer, load_results_dataset
+        from hyrax.verbs.reduction_algorithms.algorithm_registry import fetch_reducer_class
+
+        # Get reducer class
+        algorithm_name = algorithm or self.config["reduce"]["algorithm"]
+        reducer_cls = fetch_reducer_class(algorithm_name)
+
+        results_dir = create_results_dir(self.config, f"reduce_dimensions_{algorithm_name}")
```
drewoldag: I _think_ it would be ok to just use the algorithm_name here. instead of "reduce_dimensions_{algo_name}"

src/hyrax/verbs/reduce_dimensions.py
```diff
@@ -0,0 +1,159 @@
...
+
+        inference_results = load_results_dataset(self.config, results_dir=input_dir, verb="infer")
+        total_length = len(inference_results)
+
+        # Prepare data sample
+        config_sample_size = self.config["reduce"][algorithm_name].get("fit_sample_size", None)
+        sample_size = int(np.min([config_sample_size if config_sample_size else np.inf, total_length]))
+        rng = np.random.default_rng()
+        sample_indexes = rng.choice(np.arange(total_length), size=sample_size, replace=False)
+        data_sample = np.asarray(inference_results[sample_indexes]).reshape((sample_size, -1))
```
drewoldag: It's probably worth putting a comment above this section to clarify that we'll load the data sample here so that we can use it for validation when loading a pre-trained model. 

I think it will be easy to forget, and I'll wonder why this happens here instead of only happening after the `else` on line 130. i.e. only load the sample data for fitting the model.

src/hyrax/verbs/reduce_dimensions.py
```diff
@@ -0,0 +1,159 @@
...
+
+            if self.config["reduce"].get("save_fit_model", False):
+                logger.info(f"Saving fitted {algorithm_name} reducer to result directory")
+                algo_reducer.save_model(results_dir)
+
+        del data_sample
+        gc.collect()
+
+        # Transform dataset
+        batch_size = self.config["data_loader"]["batch_size"]
```
drewoldag: I would guess that we could add a batch_size config specifically for reduce_dimensions. i.e. `self.config['reduce_dimensions']['batch_size'] = 1024` (something much larger than what we use for data_loader, since that is usually for training and inference on the GPU.)

src/hyrax/verbs/reduction_algorithms/pca.py
```diff
@@ -0,0 +1,198 @@
...
+        object
+            The object loaded from the pickle file.
+        """
+        model_path = Path(model_path)
+        with open(model_path, "rb") as f:
+            object = pickle.load(f)
+
+            return object
+
+    def _transform_batch(self, batch_tuple: tuple):
```
drewoldag: I can't immediately tell, but it's possible that this method could be pushed up to the parent class if it's shared between PCA and UMAP

src/hyrax/verbs/reduction_algorithms/pca.py
```diff
@@ -0,0 +1,198 @@
...
+            second element is the results of running the PCA transform on the input as a numpy array.
+        """
+        batch_ids, batch = batch_tuple
+        with warnings.catch_warnings():
+            warnings.simplefilter(action="ignore", category=FutureWarning)
+            logger.debug("Transforming a batch ...")
+            return (batch_ids, self.reducer.transform(batch))
+
+    @staticmethod
+    def _log_memory_usage(message: str = ""):
```
drewoldag: Pretty sure this could be pushed up to the parent class too.

src/hyrax/verbs/reduction_algorithms/pca.py
```diff
@@ -0,0 +1,198 @@
...
+                f"The input dimension of the loaded PCA model ({reducer.n_features_in_})"
+                f" does not match the dimension of the inference data ({expected_input_dim})."
+            )
+
+        # Output dim check
+        if reducer.n_components_ != self.reducer.n_components:
+            raise ValueError(
+                f"The output dimension of the loaded PCA model ({reducer.n_components_})"
+                f" does not match the configured n_components ({self.reducer.n_components})."
+            )
```
drewoldag, line 97: Here and in UMAP, it might be nice to encapsulate the validation check in a private method. i.e. `_validate_pca_model_and_data` (find a better name, of course). 

This will clean up the load_model methods in both places.

src/hyrax/verbs/reduction_algorithms/algorithm_registry.py
```diff
@@ -0,0 +1,147 @@
...
+    Returns
+    -------
+    type[ReductionAlgorithm]
+        The class implementing the reducer algorithm.
+    """
+    reducer_cls = ALGORITHM_REGISTRY.get(cli_name.lower())
+    if not reducer_cls:
+        raise KeyError(
+            f"Reduction algorithm {cli_name} is not registered. "
+            f"Available algorithms: {sorted(ALGORITHM_REGISTRY)}"
```
drewoldag, line 208: Nice, I like this :)

drewoldag (commented):
This is looking like it's on the right track. Left a few comments, nothing too scary.

src/hyrax/verbs/reduction_algorithms/pca.py
Graciaaa3: Yes, I was also thinking about this the other day. I think so at least for this one. The only change would be with the `_trasform_batch` for t-SNE.

drewoldag (commented):
This batch of changes look reasonable.

Graciaaa3:
At this point all the code/docstring/tests are finished. I will address the documentation updates in a separate PR.

src/hyrax/config_migrations/migrations/004_move_umap_to_reduce.py
```diff
@@ -0,0 +1,77 @@
...
+
+    # under [reduce]
+    reduce_tbl["batch_size"] = 1024
+    move_key(cfg, "umap.save_fit_umap", "reduce.save_fit_model")
+    move_key(cfg, "umap.parallel", "reduce.parallel")
+    if "name" in umap_tbl and umap_tbl["name"] == "umap.UMAP":
+        reduce_tbl["algorithm"] = "umap"
+
+    # Move umap.UMAP kwargs to reduce.umap.kwargs
+    move_key(cfg, "umap.UMAP", "reduce.umap.kwargs")
```
drewoldag, line 54: I thought when we talked that this was going to be `rename_table`. Did that not end up working the way you would have expected it to?

src/hyrax/verbs/reduction_algorithms/tsne.py
```diff
@@ -0,0 +1,100 @@
...
+
+        if self.config["reduce"]["parallel"]:
+            import multiprocessing as mp
+
+            # Process pool loop
+            # Use 'spawn' context to safely create subprocesses after
+            # OpenMP threads are being opened by other processes in hyrax
+            # Not using spawn causes the issue linked below
+            # https://github.com/lincc-frameworks/hyrax/issues/291
+            # TODO: Find more elegant solution than just using spawn
```
drewoldag, line 58: Is this comment talking about something that is still a concern? It's not present in `pca.py`, but it's in `umap.py`.

src/hyrax/verbs/reduction_algorithms/umap.py
```diff
@@ -0,0 +1,158 @@
...
+        else:
+            # Sequential loop
+            for batch_ids, batch in tqdm(
+                args,
+                desc="Creating lower dimensional representation using UMAP:",
+                total=num_batches,
+            ):
+                transformed_batch = self.reducer.transform(batch)
+                self._log_memory_usage(f"During transformation of batch of shape {batch.shape}")
+                self.reduction_results.write_batch(batch_ids, transformed_batch)
```
drewoldag, line 156: Doesn't need to be done in this PR, but the code in this if/else for parallelism seems so similar between the concrete implementations - it almost feels like it could be moved up to the base class.

src/hyrax/verbs/reduce_dimensions.py
```diff
@@ -0,0 +1,167 @@
...
+            type=str,
+            required=False,
+            help="Directory containing the dataset to reduce dimensions for.",
+        )
+        parser.add_argument(
+            "-m",
+            "--model-path",
+            type=str,
+            required=False,
+            help="Path to a pre-existing reducer model.",
```
drewoldag:
```suggestion
            help="Path to a previously saved reducer model.",
```

src/hyrax/verbs/reduce_dimensions.py
```diff
@@ -0,0 +1,167 @@
...
+        ----------
+        algorithm : str, Optional
+            The dimensionality reduction algorithm to use.
+            If not specified, the method will look in the config for a default algorithm.
+
+        input_dir : str or Path, Optional
+            Directory containing the dataset to reduce dimensions for.
+
+        model_path : str or Path, Optional
+            Path to a pre-existing reducer model.
```
drewoldag:
```suggestion
            Path to a previously saved reducer model.
```

tests/hyrax/test_reduce_dimensions.py
```diff
@@ -0,0 +1,311 @@
...
+    tsne_result_ids = tsne_results.ids()
+    original_dataset_ids = dataset.ids()
+
+    data_shape = h.config["data_set"]["HyraxRandomDataset"]["shape"]
+
+    for idx, result_id in enumerate(tsne_result_ids):
+        dataset_idx = None
+        for i, orig_id in enumerate(original_dataset_ids):
+            if orig_id == result_id:
+                dataset_idx = i
```
drewoldag, line 306: Might want to just double check on this test, and the corresponding ones. My impression from the docstring is that this should check that `i == idx` when `orig_id == result_id`. i.e. that the output from the dimensional reduction algorithm has the same order as the input. 

It's entirely possible that my brain isn't working correctly, but I don't immediately see how this test checks that.

drewoldag (approved):
Approved with a couple of comments to look at. Half are just wording suggestions and can be ignored. The one about refactoring the parallel code can be deferred. The others might be worth looking at.

src/hyrax/config_migrations/migrations/004_move_umap_to_reduce.py
Graciaaa3: Yeah so it turns out `rename_table` only works for the very top level. It couldn't handle dotted keys like the ones we have here.

src/hyrax/verbs/reduction_algorithms/tsne.py
Graciaaa3: Sorry, I believe the `TODO` is still something that is a todo. It was from the old `umap` implementation that I transferred over. I have it for both `umap.py` and `tsne.py` and I will put these comments to `pca.py` too.

src/hyrax/verbs/reduction_algorithms/umap.py
Graciaaa3: Sure, I think the only immediate blocker is the difference for `tsne` implementation with `_fit_transform_batch`. I can open an issue for this or address it in a later pr.

tests/hyrax/test_reduce_dimensions.py
Graciaaa3: This was a bit confusing. I think the test only test for the result data to be matched with the correct original entry that has the same ID, but doesn't check for the order. I have added the assertion for `dataset_idx == idx` when it was found.

PR #934, opened 2026-05-27
Adds Field Level Collation Function Tracing to Trace Mode

SamSandwich07:
Adds a group tracking user-defined field-level collation functions to the trace output


src/hyrax/trace.py
```diff
@@ -378,6 +379,8 @@ def trace_call(self, trace_def: TraceDef, *args):
         logger.debug(f"Received Trace {trace_def.stage_name} {trace_def.func_name}")
 
         captured_params = {}
+        # print(trace_def.stage_name, trace_def.func_name, "ARGS:", args)
```
mtauraso: Probably pull these out

mtauraso (approved):
This looks good to me.

PR #940, opened 2026-06-05
Allow flexible returns from a forward function.

### LINKED ISSUE #858: Allow more flexible returns from `forward`
**Feature request**
Currently users can only return a Tensor of results from `forward` which will be saved as output. 

It seems reasonable that we should support more flexible outputs than this. Specifically, it would seem reasoanble to allow a user to return a `dictionary[str:value]` from `forward`. The `ResultDatasetWriter` would easily be able to create outputs with columns that match the key names, and likewise, the ResultDataset(Reader) class would be able to dynamically create the `getter` methods based on the schema from the files.

The backward compatable approach would simply write out the Tensor data to a `data` column. The dynamic creation of `getter` methods would work with that just fine.

maxwest-uw:
## Change Description
resolves #858 

`hyrax` users can now return a dict of `torch.Tensor`s from a model's `forward` function.

## Solution Description
- change is backwards compatible (i.e. user can return either a dict or a tensor from the forward function and everything should still work)
- in cases where the forward function just returns a tensor, the data will eventually be converted into a `dict` with the format `{ "data": <tensor> }`, which is how the data will eventually be saved to lance db anyway.
- `ResultDataset` should now always return a dict with the data for a give index or set of indices
- in cases throughout the data processing pipeline where we assume that a result is going to be a tensor, we have added a new function `ResultDataset.get_combined_tensor`, which for a given index or list of indices will return a concatenated and flattened tensor of all the available data keys. 

## Code Quality


src/hyrax/datasets/result_dataset.py
```diff
@@ -239,19 +295,33 @@ def __getitem__(self, idx: Union[int, np.ndarray]):
...
+                flat_data = result[key][i].as_py()
+                tensor = np.array(flat_data, dtype=self.tensor_dtype[key])
+                tensor = tensor.reshape(self.tensor_shape[key])
+                row_data[key] = tensor
+            tensors.append(row_data)
 
         # Return single tensor or array of tensors
         return tensors[0] if is_single else tensors
 
+    def get_combined_tensor(self, idx: int):
```
drewoldag, line 312: I'm unsure about this approach. What I had in mind was something more dynamic. Similar to what we do in teh CSVDataset. 

https://github.com/lincc-frameworks/hyrax/blob/main/src/hyrax/datasets/hyrax_csv_dataset.py#L51-L63

With a little bit of introspection at runtime, we can check the schema, and then dynamically build getter methods for each of the different columns that have been saved. 

And at that point, we would really have to put the responsibility on the user to do things like select the result dataset field that is going to be operated on. e.g. UMAP reduced, etc...

I think that this might have follow on implications for many other parts of this PR, so perhaps we can spend some time chatting about it tomorrow?

src/hyrax/verbs/umap.py
drewoldag, line 1: I'm going to need to work my way though this a bit more carefully. Ideally we would be a in a situation where backward compatibility would be maintained. I'm a little concerned about that here.

Did you by any chance have an opportunity to do some backward compatability testing?

src/hyrax/datasets/result_dataset.py
```diff
@@ -17,6 +17,7 @@
 import lancedb
 import numpy as np
 import pyarrow as pa
+import torch
```
drewoldag, line 20: I'm a little nervous about this import. It's been nice being able to keep pytorch out of our datasets. Is this just here for type checking?

drewoldag (commented):
This is looking like it's going in the right direction. I've got some comments. Might be easier to discuss these in person tomorrow if you've got the time.

src/hyrax/pytorch_ignite.py
```diff
@@ -976,12 +976,20 @@ def create_save_batch_callback(results_dir):
...
 
-    def _save_batch(batch: dict, batch_results: torch.Tensor):
+    def _save_batch(batch: dict, batch_results):
         """Receive and write batch results to results_dir immediately."""
         nonlocal data_writer
 
         # Ensure the batch results are on CPU and detached from the computation graph
-        batch_results = batch_results.detach().to("cpu")
+        if isinstance(batch_results, dict):
+            batch_results = {k: v.detach().to("cpu") for k, v in batch_results.items()}
```
drewoldag:
```suggestion
            batch_results = {k: v.detach().to("cpu").to_numpy() for k, v in batch_results.items()}
```

src/hyrax/pytorch_ignite.py
```diff
@@ -976,12 +976,20 @@ def create_save_batch_callback(results_dir):
...
+        elif isinstance(batch_results, torch.Tensor):
+            batch_results = batch_results.detach().to("cpu")
```
drewoldag:
```suggestion
            batch_results = batch_results.detach().to("cpu").to_numpy()
```

scripts/convert_results.py
```diff
@@ -210,7 +210,8 @@ def _batch_num(p: Path) -> int:
                     raise RuntimeError(
                         f"Row {lance_idx}: object_id {obj_id!r} found in Lance but not in {batch_path.name}."
                     )
-                lance_tensor = np.asarray(dataset[lance_idx])
+                print(dataset[lance_idx])
```
drewoldag:
```suggestion
```

src/hyrax/datasets/result_dataset.py
```diff
@@ -61,42 +61,54 @@ def write_batch(self, object_ids: np.ndarray, data: list[np.ndarray]):
         data : list[np.ndarray]
             List of numpy arrays (tensors) to write
         """
-        if len(object_ids) != len(data):
-            raise ValueError("Length of object_ids must match length of data")
+        # Normalize data to dict format for uniform handling
+        if isinstance(data, (np.ndarray, list)) or data.__class__.__name__ == "Tensor":
+            data_dict = {"data": np.array(data)}
+        else:
+            # data_dict = data
```
maxwest-uw: remove this

src/hyrax/datasets/result_dataset.py
```diff
@@ -61,42 +61,54 @@ def write_batch(self, object_ids: np.ndarray, data: list[np.ndarray]):
...
+            data_dict = {key: np.array(val) for key, val in data.items()}
 
-        if len(data) == 0:
+        if len(object_ids) == 0:
             return
 
         # Convert data to numpy array for uniform handling
-        data_array = np.array(data)
-        first_tensor = data_array[0]
+        # data_array = np.array(data)
```
maxwest-uw: REMOVE

src/hyrax/datasets/result_dataset.py
```diff
@@ -61,42 +61,54 @@ def write_batch(self, object_ids: np.ndarray, data: list[np.ndarray]):
...
+                        raise ValueError(
+                            f"Tensor at index {i} has shape {tensor.shape}, "
+                            f"but schema expects {tuple(self.tensor_shape[key])}"
+                        )
 
         # Flatten tensors for storage
-        flattened_data = [tensor.flatten() for tensor in data]
-
-        # Create PyArrow record batch
+        # flattened_data = [tensor.flatten() for tensor in data]
```
maxwest-uw: REMOVE

src/hyrax/datasets/result_dataset.py
```diff
@@ -61,42 +61,54 @@ def write_batch(self, object_ids: np.ndarray, data: list[np.ndarray]):
...
         batch_data = {
             "object_id": pa.array([str(oid) for oid in object_ids], type=pa.string()),
-            "data": pa.array(flattened_data, type=self.schema.field("data").type),
         }
+        for key, arr in data_dict.items():
+            flattened = [tensor.flatten() for tensor in arr]
+            batch_data[key] = pa.array(flattened, type=self.schema.field(key).type)
+
+        # Create PyArrow record batch
```
maxwest-uw: REMOVE

src/hyrax/datasets/result_dataset.py
```diff
@@ -113,39 +125,70 @@ def commit(self):
...
-        self.tensor_dtype = sample_tensor.dtype
-        self.tensor_shape = list(sample_tensor.shape)
-        flattened_size = int(np.prod(self.tensor_shape))
+        self.tensor_dtype = {}
+        self.tensor_shape = {}
+        fields = [pa.field("object_id", pa.string())]
 
         # Map numpy dtype to PyArrow type
-        pa_type = pa.from_numpy_dtype(self.tensor_dtype)
+        # pa_type = pa.from_numpy_dtype(self.tensor_dtype)
```
maxwest-uw: REMOVE

src/hyrax/datasets/result_dataset.py
```diff
@@ -113,39 +125,70 @@ def commit(self):
...
+            # Store metadata for this field
+            metadata_dict[key] = {
+                "tensor_shape": self.tensor_shape[key],
+                "tensor_dtype": str(self.tensor_dtype[key]),
+            }
+        # Create schema with metadata
+        metadata = {k: json.dumps(v).encode("utf-8") for k, v in metadata_dict.items()}
+        self.schema = pa.schema(fields, metadata=metadata)
+
+        # self.schema = pa.schema(
```
maxwest-uw: REMOVE

src/hyrax/datasets/result_dataset.py
maxwest-uw: I will remove this and change the visualize code (before it is deleted)

src/hyrax/datasets/result_dataset.py
```diff
@@ -281,7 +352,7 @@ def get_data(self, idx: int):
         np.ndarray
             Data tensor
         """
-        return self.__getitem__(idx)
+        return self.__getitem__(idx)["data"]
```
maxwest-uw: metaprogram getter functions for other fields (look at CSV dataset for guidance)

src/hyrax/verbs/save_to_database.py
```diff
@@ -162,7 +162,7 @@ def run(self, input_dir: Union[Path, str] | None = None, output_dir: Union[Path,
             vectors = []
             ids = []
             for idx in indices:
-                vector = np.asarray(inference_data_set[idx]).flatten()
+                vector = inference_data_set.get_combined_tensor(idx)
```
maxwest-uw: note to drew: think about this

tests/hyrax/test_forward_returns.py
```diff
@@ -0,0 +1,139 @@
...
+    "data": {
+        "dataset_class": "HyraxRandomDataset",
+        "data_location": "./data/test",
+        "primary_id_field": "object_id",
+    }
+}
+
+
+@hyrax_model
+class DummyModel(nn.Module):
```
maxwest-uw: get rid of `DummyModel` and use `HyraxLoopback` (if possible)

PR #941, opened 2026-06-05
Dataset splits and balancing

drewoldag:
Note that this technical spec was generated based on this design spec:
https://docs.google.com/document/d/11_Ma-hF5u7xV-MBWXvt68ibn-UqCesS9jrXeD8AzOlM/edit?tab=t.0

## Summary

Adds an implementation-facing **technical spec** for splits and dataset
balancing, derived from the *"Splits and dataset balancing in Hyrax"* design
doc as well as the correspoding code changes.

The spec grounds the design in the current codebase and specifies:

- New `[split]` and `[balance]` config tables (defaults reproduce current behavior).
- A new class-based `create_splits` verb (`h.create_splits()`) over a new
  `splitting_utils.py` module (split / stratify / weight / persist / equivalency).
- Config migration `003` moving `split_fraction` out of `[data_request]` into
  `[split]`, plus the corresponding Pydantic schema removals.
- `setup_dataset` reduced to provider construction only; split logic relocated.
- `Subset` + `WeightedRandomSampler` support in `dist_data_loader`.
- Persisted `<group>_split.npz` + `split_config.toml` and a config-equivalency
  reuse check.

## Resolved design-doc ambiguities

A decision table (§3) records where the spec diverges from the design doc and why.
Most notable:

- `validation` → `validate` to match existing group names.
- `WeightedRandomSampler` (`replacement=True`) is scoped to the *balancing* case
  only; the uniform default keeps `SubsetRandomSampler` (no replacement) so
  default training behavior is unchanged.
- Sampler indexes the `Subset` in local space (fixes a double-indexing bug in the
  design pseudocode).
- Results dir uses the existing `create_results_dir(config, "splits")` convention.

## Open questions

§17 lists decisions to settle before implementation (train results copy-vs-pointer,
infer persistence default, weight normalization, `.npz` "no weights" sentinel).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


specs/splits_and_balancing_spec.md
```diff
@@ -0,0 +1,676 @@
...
+4. Scales to O(10–100M) items: never allocate a full-length weight array when a
+   subset is selected (use `Subset` + per-subset weights).
+5. Backward compatible for train/validation and for inference subset selection.
+
+**Non-goals (this spec)**
+
+- Data augmentation (`augment_<field>` methods, TTA) — separate doc. We only
+  reserve the warning/error hooks described in §13.
+- A dataset-content hash for equivalency (future; we compare config only).
+- Removing the legacy `[data_set].{train,validate,test}_size` path or the
```
drewoldag: I kind of think that this might be a fine time to fully deprecate the legacy splitting code and all of the cruft that has developed around it.

specs/splits_and_balancing_spec.md
```diff
@@ -0,0 +1,676 @@
...
+  - Float domain: `0.0 < f <= 1.0`. Default `1.0`.
+  - For groups that share a `data_location` (e.g. `train`+`validate` over the
+    same source), `0.0 < Σ fractions <= 1.0`; the splits are **non-overlapping**
+    partitions of that shared source. Σ < 1.0 is allowed (use a subset).
+  - For groups with distinct `data_location`s, each fraction is an independent
+    subset of that group's own data.
+  - `infer` is always treated independently (never partitioned against
+    train/validate/test).
+- **`split.rng_seed`** — RNG seed for shuffling/partitioning. `""` ⇒ use
+  `config["data_set"]["seed"]` (which itself may be `false` ⇒ system entropy).
```
drewoldag: Make sure the fall back is the RNG that is currently used.

specs/splits_and_balancing_spec.md
```diff
@@ -0,0 +1,676 @@
...
+## 5. Weighting model (precise formulas)
+
+Weights are the per-sample weights handed to `WeightedRandomSampler` for a given
+split. Let a split contain indices `I`, with `class(i)` the label of item `i`,
+`count_c = |{i ∈ I : class(i) = c}|`, and `target_c` the target fraction for
+class `c`.
+
+- **No balancing** (group ∉ `balance.groups`, or `balance.field` falsy):
+  `weights = None`. (Internally `None` ⇒ uniform; see §11 for why we keep
+  `None` rather than an array of `1.0`.)
```
drewoldag: Check section to understand justification for None over weights =1.0.

specs/splits_and_balancing_spec.md
```diff
@@ -0,0 +1,676 @@
...
+
+`validate_split_config` enforces:
+- per-group float domain `(0,1]`;
+- all-floats-or-all-paths (else raise);
+- paths share a parent dir (else raise);
+- for groups sharing a `primary_data_location`, `Σ fractions <= 1.0` (else raise).
+
+`validate_balance_config` enforces the §4.2 rules: `field` getter exists on each
+relevant group's primary dataset; `groups ⊆ data_request` (warn on extras);
+`distribution` sums to 1.0; unknown labels raise; missing labels warn (weight 0).
```
drewoldag: The unknown labels and missing labels may be a separate check since those happen after the full dataset scan.

specs/splits_and_balancing_spec.md
```diff
@@ -0,0 +1,676 @@
...
+
+### 7.3 Core split computation
```
drewoldag, line 470: This doesn't cover the behavior when balance.distribution is non-empty

specs/splits_and_balancing_spec.md
```diff
@@ -0,0 +1,676 @@
...
+    (using `count_c` measured **within that split**);
+  - else: `weights = None`.
+
+Return `{group: {"indexes": ndarray, "weights": ndarray|None}}`.
+
+### 7.4 Persistence / loading
+
+```python
+def persist_splits(results_dir: Path, splits, config) -> None
+def load_split_files(paths: dict[str, Path]) -> dict[str, dict[str, np.ndarray]]
```
drewoldag, line 540: The logic in this function isn't defined below.

specs/splits_and_balancing_spec.md
```diff
@@ -0,0 +1,676 @@
...
+  same shape as `h.prepare()`.
+- This satisfies CLAUDE.md "only add a verb if specifically told to" — the design
+  explicitly mandates `create_splits`.
+
+---
+
+## 11. `dist_data_loader` changes (`pytorch_ignite.py`)
+
+Replace the `split=False` / `isinstance(split, bool)` branch to use `Subset` +
+weight-aware sampler. Keep returning `(loader, indexes)`.
```
drewoldag: Note for Drew - reconsider whether we need to return indexes anymore.

specs/splits_and_balancing_spec.md
```diff
@@ -0,0 +1,676 @@
...
+
+## 14. Backward compatibility & deprecations
+
+- Untouched `[split]`/`[balance]` ⇒ all fractions `1.0`, no balance ⇒ identical
+  to today (each group uses its full primary dataset; uniform no-replacement
+  shuffle for train).
+- Old configs with `data_request.*.split_fraction` are auto-migrated to `[split]`
+  (migration 003) with the standard deprecation log.
+- Legacy `[data_set].{train,validate,test}_size` + `pytorch_ignite.create_splits`
+  remain deprecated and functional; no behavior change.
```
drewoldag: Consider if this is the right time to deprecate and remove this older logic.

specs/splits_and_balancing_spec.md
```diff
@@ -0,0 +1,676 @@
...
+- `docs/dataset_splits.rst` — document `[split]`/`[balance]` and `create_splits`.
+
+---
+
+## 17. Open questions
+
+1. **Train results copy vs. write-in-place.** When `train` reuses an equivalent
+   pre-existing split dir, do we copy the `.npz`+TOML into the train results dir
+   (full self-contained record, per "take your data to go") or just record a
+   pointer? This spec assumes **copy**.
```
drewoldag: Copy is the correct choice here.

specs/splits_and_balancing_spec.md
```diff
@@ -0,0 +1,676 @@
...
+2. **Infer persistence.** Should `create_splits` persist for `infer`/`test` by
+   default, or only for `train`/explicit `h.create_splits()`? This spec defaults
+   `persist=False` for infer/test.
+3. **Weight normalization.** Normalize `split_weights` to mean 1.0 for
+   readability, or store raw `target_c/count_c`? No functional difference for WRS;
```
drewoldag: I think store the raw fraction.

specs/splits_and_balancing_spec.md
```diff
@@ -0,0 +1,676 @@
...
+   spec leaves it optional.
+4. **`np.savez` weight sentinel for `None`.** Confirm the on-disk convention for
+   "no weights" (omit the `weights` array vs. write empty) and mirror it in
```
drewoldag: I think omit the weights array. Save space, easy for someone to hack together by hand if needed.

specs/splits_and_balancing_spec.md
drewoldag: The legacy splitting code is being removed in PR #944

src/hyrax/splitting_utils.py
```diff
@@ -0,0 +1,582 @@
...
+        return splits
+
+    # Search for a reusable equivalent split
+    equivalent_paths = find_equivalent_split(config)
+    if equivalent_paths is not None:
+        logger.info("Reusing equivalent split from %s", next(iter(equivalent_paths.values())).parent)
+        splits = load_split_files(equivalent_paths)
+        assign_splits_to_providers(datasets, splits)
+        if persist and results_dir is not None:
+            persist_splits(results_dir, splits, config)
```
drewoldag, line 627: Doesn't need to happen in this PR, but if this is being run via `create_splits` _and_ we find identical splits in a previous directory, I don't think that we want to copy the splits into a new directory.

src/hyrax/splitting_utils.py
```diff
@@ -0,0 +1,582 @@
...
+
+    if not field:
+        return
+
+    for group_name, provider in datasets.items():
+        if group_name == "infer":
+            continue
+        primary_ds = _primary_instance(provider)
+        if not hasattr(primary_ds, f"get_{field}"):
+            raise RuntimeError(
```
drewoldag, line 183: Need much nicer looking error handling here. We don't need to emit the entire stack trace - perhaps this would be better accomplished with a Pydantic validation?

src/hyrax/splitting_utils.py
```diff
@@ -0,0 +1,582 @@
...
+        if g not in datasets:
+            logger.warning("balance.groups contains '%s' which is not in data_request; ignoring.", g)
+
+    if distribution:
+        for label, val in distribution.items():
+            try:
+                fval = float(val)
+            except (TypeError, ValueError) as err:
+                raise RuntimeError(
+                    f"balance.distribution['{label}'] = {val!r} is not a valid float."
```
drewoldag, line 199: Each of these RuntimeErrors should be presented to the user in a much nicer way. Pydantic stands a the obvious choice. Yes, we need to stop the run, but these errors deserve a simple error message, not an entire stack trace.

src/hyrax/splitting_utils.py
```diff
@@ -0,0 +1,582 @@
...
+        sum > 1.0, paths not in same directory).
+    """
+    split_cfg = config.get("split", {})
+    group_values = {name: split_cfg.get(name, 1.0) for name in datasets}
+
+    path_groups = [n for n, v in group_values.items() if _is_path_value(v)]
+    float_groups = [n for n in group_values if n not in path_groups]
+
+    if path_groups and float_groups:
+        raise RuntimeError(
```
drewoldag, line 115: Again, all these Runtime errors should be presented in a nice way - a stack trace doesn't help here.

src/hyrax/splitting_utils.py
```diff
@@ -0,0 +1,582 @@
...
+    Raises
+    ------
+    RuntimeError
+        If distribution contains a label absent from the dataset.
+    """
+    if not distribution:
+        return
+    for label in distribution:
+        if label not in observed_labels:
+            raise RuntimeError(
```
drewoldag, line 238: This runtime error should be nicely presented too. Basically any time that there is a misconfiguration, there shouldn't be a stack trace.

src/hyrax/splitting_utils.py
```diff
@@ -0,0 +1,582 @@
...
+                if name == last_name and total >= 1.0 - 1e-5:
+                    count = n_items - offset
+                result[name] = {
+                    "indexes": np.array(indices[offset : offset + count], dtype=np.int64),
+                    "weights": None,
+                }
+                offset += count
+
+        else:
+            # Stratified: build class index map, then distribute per-class
```
drewoldag, line 323: If we get to this part of `_compute_splits`, we should probably log a message explaining why it's taking so long. i.e. we're doing a full scan of the dataset to produce the requested _balanced_ subsets.

src/hyrax/trace.py
```diff
@@ -348,13 +346,7 @@ def new_len(obj):
...
                 return obj.split_indices[self.trace_batch_size - 1] + 1
 
-            split_fraction = 1.0 if obj.split_fraction is None else obj.split_fraction
-            max_len = int(np.ceil(self.trace_batch_size / split_fraction))
-
-            # Don't ever make the new length longer than the old length
-            # Can happen in some weird split situations on small datasets (like RandomDataset)
-            # in testing contexts
-            return min(max_len, raw_func(obj))
+            return raw_func(obj)
```
mtauraso: This isn't quite right, but #948 is a fix based on this branch.

src/hyrax/datasets/data_provider.py
```diff
@@ -426,8 +426,12 @@ def __repr__(self) -> str:
                 repr_str += f"  Dataset class: {data['dataset_class']}\n"
                 if "data_location" in data:
                     repr_str += f"  Data location: {data['data_location']}\n"
-                if "split_fraction" in data:
-                    repr_str += f"  Fraction of data to use: {data['split_fraction']}\n"
+                if self.primary_dataset == friendly_name and self.split_indices is not None:
+                    repr_str += f"  Selected items: {len(self.split_indices)}"
+                    if self.split_weights is not None:
+                        n_classes = len(set(self.split_weights))
+                        repr_str += f" (rebalanced, {n_classes} weight values)"
```
drewoldag: Follow up: I believe what is logged here could be made more useful

src/hyrax/hyrax_default_config.toml
```diff
@@ -500,3 +500,43 @@ num_detail_plots = 6
...
+groups = []
+
+[balance.distribution]
+# Maps class label → target fraction in (0.0, 1.0].  Empty table means equal
+# target across all observed classes (uniform).  Values must sum to exactly 1.0.
+# Example:
+# label_0 = 0.25
+# label_1 = 0.75
+
+# [label]
```
drewoldag: This should be an empty dict, not a commented out table.

src/hyrax/hyrax_default_config.toml
```diff
@@ -500,3 +500,43 @@ num_detail_plots = 6
 
 # The directory containing the ONNX model used for inference in production
 model_directory = false
+
+
+[split]
+# Per-group split configuration.  Keys correspond to groups in [data_request].
```
drewoldag, line 506: Review all the new comments and attempt to reduce them to 2-3 lines where possible.

src/hyrax/hyrax_default_config.toml
```diff
@@ -500,3 +500,43 @@ num_detail_plots = 6
...
+# Each value is a fraction of the group's primary dataset to use — float in
+# (0.0, 1.0] — or a path to a previously generated <group>_split.npz file.
+# Groups absent from this table default to 1.0 (use the full dataset).
+
+# RNG seed for shuffling/partitioning.  Empty string falls back to
+# config["data_set"]["seed"] (reproducing current behavior exactly).
+# A non-empty value uses a dedicated generator and does not touch global RNG state.
+rng_seed = ""
```
mtauraso: The global one is `seed=False` We should probably be consistent

src/hyrax/pytorch_ignite.py
```diff
@@ -58,30 +58,23 @@ def setup_dataset(
...
-        listed here are silently skipped.  When ``None`` (the default) every
-        group in the data_request is loaded — preserving backward compatibility.
+        When provided, only create DataProvider instances for the listed groups.
+        When ``None`` every group in the data_request is loaded.
     shuffle : bool, optional
-        Whether to shuffle indices when computing ``split_fraction``-based
-        partitions via :func:`create_splits_from_fractions`.  Defaults to
-        ``True``.  Set to ``False`` for inference / test verbs where
-        deterministic ordering is required.
+        Unused; kept for backward-compatibility with call sites that still pass
```
drewoldag, line 71: Follow up: attempt to remove the unused `shuffle` argument and simplify the input args.

src/hyrax/hyrax_default_config.toml
```diff
@@ -500,3 +500,43 @@ num_detail_plots = 6
...
+
+
+[balance]
+# Name of the field identifying each item's class.  Must resolve to a
+# get_<field> getter on the primary dataset.  Empty string disables stratification.
+field = ""
```
mtauraso: Less sure here whether default value should be `""` or `False->None` I think we've not been consistent about strings

src/hyrax/pytorch_ignite.py
```diff
@@ -224,129 +199,35 @@ def dist_data_loader(
...
-    # prevent silent out-of-range errors or incorrect data access.
-    provider_lengths = {name: len(provider) for name, provider in dataset_providers.items()}
-    unique_lengths = set(provider_lengths.values())
-    if len(unique_lengths) > 1:
-        raise RuntimeError(
-            f"All providers passed to create_splits_from_fractions must have the same length. "
-            f"Got lengths: {provider_lengths}"
-        )
+    elif shuffle:
+        sampler = SubsetRandomSampler(range(n), generator=torch_rng)
```
mtauraso, line 226: Follow up: Probably want all rngs to be cpu rngs since we expect the cpu to own indexes.

tests/hyrax/test_split_fraction_integration.py
drewoldag, line 1: Reinstate this file, and update configs used in here so that we can use this for regression testing

tests/hyrax/test_splitting_utils.py
```diff
@@ -0,0 +1,619 @@
...
+            label = index_to_label[idx]
+            counts[label] = counts.get(label, 0) + 1
+        total = sum(counts.values())
+        return {k: v / total for k, v in counts.items()}
+
+    train_fracs = class_fractions(result["train"]["indexes"].tolist())
+    val_fracs = class_fractions(result["validate"]["indexes"].tolist())
+
+    for lbl in labels:
+        assert abs(train_fracs.get(lbl, 0) - 1 / 3) < 0.10, f"train class {lbl} proportion off"
```
mtauraso: Not sure about this formula. Looks like the subtract, compare pattern for floating points, but I'd expect the 0.1 to be more like `1/300` since there are 300 data points?

Maybe this is right, not exactly sure how this test works.

tests/hyrax/test_split_fraction_integration.py
drewoldag, line 1: Revive this test and update the configs so that it can be used for regression testing

tests/hyrax/test_splitting_utils.py
mtauraso: Okay I think what's going on here is that each fraction by label should be close to 1/3, since HyraxRandomDataset will equally distribute between 3 classes.

The thing is with 300 data points, 20% in training (60 data points) and 1/3 to each class we should be getting 20 data points for each class in training. Variance of 1/10 in the fraction would be 6 data points

Maybe with variance 1/10th is what's needed for the test to pass. If so we should be able to bump the test to n=10,000 and the comparison fraction should go down to 0.02.

tests/hyrax/test_split_fraction_integration.py
```diff
@@ -1,358 +1,189 @@
-"""Tests for the split_fraction integration across DataProvider,
```
mtauraso, line 1: After reading `test_splitting_utils` I'm a little more convinced that cases are covered; however, the proof I was looking for in this file is something like:
- Tests that are no longer relevant (validation of old config, old dataprovider attributes) are removed.
- Tests that are are still looking at valid cases are modified in the *most minimal manner* to align with the new splits config.

What it looks like has happened here is more of a put it back in and encode all assumptions of this change, which still leaves us with a question about whether the old tests pass or not, and we need to read through and analyze these.

The diff should make a really clean distinction between which tests are not relevant and which tests are, and any test surviving the cull should have such small changes made that it is trivial to see they are still valid.

src/hyrax/splitting_utils.py
mtauraso: Agree.

src/hyrax/splitting_utils.py
```diff
@@ -0,0 +1,582 @@
...
+
+    # Search for a reusable equivalent split
+    equivalent_paths = find_equivalent_split(config)
+    if equivalent_paths is not None:
+        logger.info("Reusing equivalent split from %s", next(iter(equivalent_paths.values())).parent)
+        splits = load_split_files(equivalent_paths)
+        assign_splits_to_providers(datasets, splits)
+        if persist and results_dir is not None:
+            persist_splits(results_dir, splits, config)
+        return splits
```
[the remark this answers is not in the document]
mtauraso: Isn't this case covered in validation?

src/hyrax/splitting_utils.py
```diff
@@ -0,0 +1,582 @@
...
+
+    # Sum check per shared primary_data_location (infer always independent)
+    location_fracs: dict[str, list[float]] = {}
+    for name, provider in datasets.items():
+        if name == "infer":
+            continue
+        loc = provider.primary_data_location
+        if loc:
+            frac = float(split_cfg.get(name, 1.0))
+            location_fracs.setdefault(loc, []).append(frac)
```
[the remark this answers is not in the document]
mtauraso: Agree on explicit handling to expose this case to readers of code.

Not sure if we should reject `split=0.0` for a group. I see that the error says (0.0-1.0] is the range, but isn't "{train:0.8, validate:0.0, test:0.2}" valid?  (you just aren't doing validation)

src/hyrax/splitting_utils.py
```diff
@@ -0,0 +1,582 @@
...
+        If getter is missing, distribution is malformed, or distribution sum ≠ 1.0.
+    """
+    balance_cfg = config.get("balance", {})
+    field = balance_cfg.get("field") or None
+    balance_groups = list(balance_cfg.get("groups") or [])
+    distribution = dict(balance_cfg.get("distribution") or {})
+
+    if not field:
+        return
+
```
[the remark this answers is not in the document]
mtauraso: This does seem like an important error

src/hyrax/splitting_utils.py
mtauraso: Agree generally we'd rather the pydantic message than straight RuntimError

tests/hyrax/test_split_fraction_integration.py
mtauraso: This all being said... maybe the right choice was to delete all of it, be very sure that test_splitting_utils covers the cases we care about.

src/hyrax/splitting_utils.py
```diff
@@ -0,0 +1,618 @@
+"""Splitting and dataset balancing utilities for Hyrax datasets."""
```
mtauraso, line 1: I feel like there is more grouping of functions possible in this file, at least into classes, perhaps into separate files.

This is probably a follow-up action, just to keep the codebase organized. I'm not currently seeing the correct organization, but I'm pretty sure we can do better than pure functions.

src/hyrax/splitting_utils.py
```diff
@@ -0,0 +1,618 @@
...
+                            "will be excluded from all split groups.",
+                            raw_val,
+                            field,
+                            len(inds),
+                        )
+                    else:
+                        rekeyed[alias] = inds
+                class_inds = rekeyed
+
+            validate_distribution_labels(distribution, set(class_inds))
```
mtauraso, line 357: This feels out of place, like it should happen much earlier with the other validation code.

src/hyrax/splitting_utils.py
```diff
@@ -0,0 +1,618 @@
...
+                "Dataset class '%s' is absent from balance.distribution; "
+                "it will receive weight 0 (no samples drawn for this class).",
+                label,
+            )
+
+
+# ── Core split computation ─────────────────────────────────────────────────────
+
+
+def _compute_splits(config: dict, datasets: dict[str, DataProvider]) -> dict[str, dict]:
```
mtauraso, line 254: I think this function and its local vars are one of the best starting points for creating an object and member functions to impose some organizing principles on this file/subsystem

mtauraso: Its also a good thing to functionally decompose a bit. We get pretty nested with `if` and `for `in this function in ways that don't help readability.

mtauraso (approved):
On balance I think this should go in and most things should be addressed in follow-up. On the top of my wishlist of things to do on this PR and not as a follow up:

1) Crisp decision on the tests in `test_split_fraction_integration.py` / `test_splitting_utils.py` Right now, I think my feedback to drew of "put the tests back" has muddied the waters on "does this splitting work?" and I'd like to either put some back with really surgical changes, or just remove them and have a lot of scrutiny on `test_splitting_utils.py`

2) Raising N on the stratified sampling test and/or adding a few other permutations. A lot of our issues with splitting code have revolved around edge cases, so we should memorialize these in tests.

3) If there's a low-friction decision to take on the config default values for `seed_rng` we should take it and make all our rng configs consistent.

Putting "Approve" because I don't want further review bandwidth to be a blocker. To the extent my wishlist isn't satisfied, we will just have to deal with this stuff as high-priority follow-up items.

I do think we should make a list somewhere of the major follow up items, but that list I think is much more useful after this merges.

I'm also tacking on a commit to this branch which fixes CI, just so that's out of the way.

src/hyrax/hyrax_default_config.toml
drewoldag: Good call, I'll update to `false`.

src/hyrax/hyrax_default_config.toml
drewoldag: Same as above, will update to `false`.

tests/hyrax/test_splitting_utils.py
drewoldag: Bumped this up to n=10_000, and reduced the threshold to 2%. 1% seemed like it was consistenly fine when I ran manually, but 2% gives a nice buffer to reduce false alarms.

tests/hyrax/test_infer.py
```diff
@@ -77,31 +77,25 @@ def test_infer_order(loopback_hyrax, train_shuffle):
 
 
 def test_infer_split_fraction_preserves_underlying_dataset_order(loopback_hyrax):
-    """Inference with split_fraction should process and write the selected subset
-    in the same order as the underlying dataset.
+    """Inference with split config should restrict the dataset to a deterministic
+    subset whose indices match the first N indices of the underlying dataset.
     """
     from hyrax.pytorch_ignite import setup_dataset
```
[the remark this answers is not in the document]
drewoldag: This test does check that the returned indexes are what would be expected for the `infer` data group, and that the order is preserved. I think it's fine to leave this as is.

src/hyrax/splitting_utils.py
drewoldag: Captured this in #950 - resolving this comment

src/hyrax/splitting_utils.py
drewoldag: Captured this in #950 - resolving this comment

src/hyrax/splitting_utils.py
drewoldag: Captured this in #950 - resolving this comment

src/hyrax/splitting_utils.py
drewoldag: Captured this in #950 - resolving this comment

src/hyrax/splitting_utils.py
```diff
@@ -0,0 +1,622 @@
...
+    from hyrax.datasets.data_provider import DataProvider
+
+logger = logging.getLogger(__name__)
+
+
+# ── Private helpers ────────────────────────────────────────────────────────────
+
+
+def _is_path_value(val: Any) -> bool:
+    """Return True when val is a non-empty string (i.e. a file path, not a fraction)."""
```
drewoldag, line 22: Seems like this should actualy check if the value is a Path that can be resolved - not simply a string

src/hyrax/splitting_utils.py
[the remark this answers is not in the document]
drewoldag: yes, this seems to be pretty well covered in `find_equivalent_splits`, specifically: 
```
for group_name in cur_dr:
    if group_name not in prev_dr:
        diffs.append(f"group '{group_name}' absent from previous split config")
        continue
```

src/hyrax/splitting_utils.py
drewoldag: I'll double check on this, but I believe we have to wait until here because the full dataset scan must have completed before we can do this validation check. 

It could, however, happen prior to this `if label_cfg:...` block.

src/hyrax/splitting_utils.py
```diff
@@ -0,0 +1,632 @@
...
+
+    if distribution:
+        for label, val in distribution.items():
+            try:
+                fval = float(val)
+            except (TypeError, ValueError) as err:
+                raise RuntimeError(
+                    f"balance.distribution['{label}'] = {val!r} is not a valid float."
+                ) from err
+            if not (0.0 < fval <= 1.0):
```
drewoldag: I believe that it's ok to allow fval = 0.0. That would be the user explicitly stating "this label is not going to be part of the dataset". 
```suggestion
            if not (0.0 <= fval <= 1.0):
```

tests/hyrax/test_data_request_config.py
```diff
@@ -527,40 +527,6 @@ def test_as_dict_with_dict_configs():
...
-        dataset_class="HyraxRandomDataset",
-        data_location="/dev/null",
-        primary_id_field="id",
-        split_fraction=fraction,
-    )
-    assert cfg.split_fraction == fraction
-
-
-@pytest.mark.parametrize("fraction", [0.0, -0.1, -1.0, 1.01, 2.0, 100.0])
-def test_split_fraction_invalid_values(fraction):
```
drewoldag, line 553: We have retained the logic that ensures that split fractions are (0.0, 1.0], but we don't have a test for it.

tests/hyrax/test_data_request_config.py
drewoldag, line 1: I went through all the tests that were removed here and made sure that we have rewritten the check in test_splitting_utils.

tests/hyrax/test_split_fraction_integration.py
```diff
@@ -1,358 +1,189 @@
...
 
-    return _StubProvider(length, split_fraction)
+def _make_single_provider(config, data_location):
+    """Convenience wrapper returning a single DataProvider for the 'train' group."""
+    return _make_providers(config, data_location)["train"]
 
 
 # ===========================================================================
 # 1. DataProvider attribute tests
 # ===========================================================================
```
drewoldag, line 49: Looked over section 1 closely, this seems to have been ported over correctly.

tests/hyrax/test_split_fraction_integration.py
drewoldag: I tried to go through this file pretty thoroughly and reintroduce tests that were not recreated in test_splitting_utils.

drewoldag: I know this isn't the preferred clean method of regression testing, but I do think that we've got our coverage back given the api changes.

src/hyrax/splitting_utils.py
```diff
@@ -0,0 +1,582 @@
...
+    for group_name, provider in non_infer.items():
+        loc = provider.primary_data_location or group_name
+        location_groups.setdefault(loc, {})[group_name] = provider
+
+    for _loc, loc_datasets in location_groups.items():
+        first_provider = next(iter(loc_datasets.values()))
+        n_items = len(first_provider)
+        fractions = {name: float(split_cfg.get(name, 1.0)) for name in loc_datasets}
+        total = sum(fractions.values())
+        last_name = list(loc_datasets.keys())[-1]
```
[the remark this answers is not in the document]
drewoldag: I don't think that the current validation treats false as the default. If a data group from the data_request isn't present in the `split` table, the default value is 1.0. 

Additionally, the default value for the `split` table is an empty dict. And given that, the default values for each of the data_groups is then 1.0.

src/hyrax/pytorch_ignite.py
```diff
@@ -224,129 +199,35 @@ def dist_data_loader(
...
+        # context so the weights tensor lands on CPU, and use a matching CPU
+        # generator.  SubsetRandomSampler uses torch.randperm (int64, MPS-safe)
+        # so torch_rng can stay on idist.device().
+        cpu_rng = torch.Generator(device="cpu")
+        if seed is not None:
+            cpu_rng.manual_seed(seed)
+        with torch.device("cpu"):
+            sampler = WeightedRandomSampler(
+                weights=list(weights), num_samples=n, generator=cpu_rng, replacement=True
             )
```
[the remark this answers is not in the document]
drewoldag: I made this change locally, and started getting sporadic test failures. Specifically running `test_dist_data_loader_sampler_types` would fail only in the case of running the entire test suite. When I ran only that test, or only the tests in test_splitting_utils module, it would pass.

The error message is: 
```
self = <torch.utils._device.DeviceContext object at 0x358701450>
func = <built-in method as_tensor of type object at 0x113d8f948>, types = ()
args = (tensor([0.0227, 0.0227, 0.0227, 0.0227, 0.0227, 0.0227, 0.0227, 0.0227, 0.0227,
        0.0227, 0.0227, 0.0227, 0.022...0.0250, 0.0250, 0.0250, 0.0250, 0.0250,
        0.0250, 0.0250, 0.0250, 0.0250, 0.0250, 0.0250], dtype=torch.float64),)
kwargs = {'device': device(type='mps'), 'dtype': torch.float64}

    def __torch_function__(self, func, types, args=(), kwargs=None):
        kwargs = kwargs or {}
        if func in _device_constructors() and kwargs.get("device") is None:
            kwargs["device"] = self.device
>       return func(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^
E       TypeError: Cannot convert a MPS Tensor to float64 dtype as the MPS framework doesn't support float64. Please use float32 instead.
```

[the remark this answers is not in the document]
drewoldag: I asked Opus to go deep on this one, it came back with this:

The context manager is actually the robust path here, not the fragile one. WeightedRandomSampler.__init__ internally calls torch.as_tensor(weights, dtype=torch.double) with no device kwarg, so an explicit CPU tensor gets moved to the ambient default device regardless — I verified a CPU float64 tensor lands on the active default device after that call. The only ways to pin it to CPU are an active device context or resetting the global default; passing an explicit tensor doesn't help. The real fragility is create_engine leaking torch.set_default_device(mps) without restoring it, which I'm addressing with a test-isolation fixture.

mtauraso (approved):
Approved, I'm merging this

PR #942, opened 2026-06-09
Augmentation in Hyrax V0

mtauraso:
In [the spec](https://docs.google.com/document/d/1pNaIUNJ37zAK-rWzoBCZrS2PXWN5Jw6ar0Da0iQPtPA/edit?tab=t.0) this change covers:
- V1 of the config `augment: [true|false]`
- None of the caching changes.
- We defer multi-threaded reproducibility of augment `rng_seed` values, while retaining it when the dataloader is single threaded (this is the default case now)


src/hyrax/datasets/data_provider.py
```diff
@@ -362,6 +378,23 @@ def pull_up_primary_dataset_methods(self):
...
 
+    def _augment_rng_seed(self, idx: int) -> np.int64:
+        """Return a deterministic, epoch-varying seed for one (epoch, idx) pair.
+
+        A fresh RNG seeded with ``[self._epoch_seed, idx]`` gives independent seeds
+        per index within an epoch while remaining reproducible across runs.
+        """
+        return np.random.default_rng([self._epoch_seed, idx]).integers(
+            np.iinfo(np.int64).min, np.iinfo(np.int64).max, dtype=np.int64
+        )
```
mtauraso: This has an issue with repeated indexes during sampling. What we need to actually do is make the rng seeds given to augment functions reproducible given that: 
- The master rng seed is the same
- The same sequence of indexes and same number of epochs are in use

We also need to ensure that repeated indexes in the same epoch *don't* get the same random seed by design (though they might randomly).

This design fails to achieve the objective.

mtauraso: This is made more complicated by the fact that calls to `resolve_data` don't have a total order because they can be called on multiple threads. (e.g. if we pass `num_workers` to pytorch dataloader via hyrax config).

So even if the epoch has an order that includes repeated indexes (this would be the sampler output order from WeightedRandomSampler in the unimplemented sampling proposal), this order is not available to us here directly.

tests/hyrax/test_augmentation.py
```diff
@@ -0,0 +1,296 @@
...
+    config = _make_hyrax_config()
+    dp = _make_dp(config, "SeedTrackingDataset", tmp_path, augment=True, fields=["image"])
+    dataset_instance = dp.prepped_datasets["data"]
+
+    dp.resolve_data(0)
+    dp.resolve_data(1)
+    assert dataset_instance.seeds_by_idx[0]["image"] != dataset_instance.seeds_by_idx[1]["image"]
+
+
+def test_on_epoch_start_dispatches_to_all_datasets(tmp_path):
```
mtauraso, line 196: This makes a data provider with one dataset, but "all" in the name implies its verifying that when multiple datasets are in use, all of them get the dispatch.

We should test the multiple dispatch case, and I think the right way to do it is to modify this test to actually have more than one dataset in it.

src/hyrax/verbs/train.py
```diff
@@ -202,6 +203,12 @@ def run(self):
         # Create trainer, a pytorch-ignite `Engine` object
         trainer = create_trainer(model, config, results_dir)
 
+        # Dispatch on_epoch_start to all DataProviders at the start of each epoch.
```
mtauraso, line 127: Should we dispatch this only during training, or also during other verbs?

If the dataset does important bookkeeping/initialization in there, we probably should ensure its called once even in `infer`, `test` verbs (maybe even `engine`?)

TBD code change on answering these questions

src/hyrax/datasets/data_provider.py
```diff
@@ -808,51 +851,79 @@ def resolve_data(self, idx: int) -> dict[str, dict[str, Any] | str | None]:
...
-            if self.primary_dataset_id_field_name not in returned_data.get(self.primary_dataset, {}):
-                if object_id_str is not None:
-                    object_id = object_id_str
-                else:
-                    primary_getter = self.dataset_getters[self.primary_dataset]
-                    object_id = str(primary_getter[self.primary_dataset_id_field_name](idx))
-            else:
-                object_id = returned_data[self.primary_dataset][self.primary_dataset_id_field_name]
+            self.data_cache.insert_into_cache(idx, base_data)
+            tensorboardx_logger.log_duration_ts(f"{prefix}/cache_miss_s", start_time)
```
mtauraso, line 900: This is right for this PR, but there's interlocking issues here.

We probably want to log an analogous `augment_time_s` metric for augmentation
We probably also want in future (when its possible to cache augmented data) for the cache miss/hit numbers to include the time it takes to generate the augmented data on a miss.

mtauraso:
@copilot Can you look at the comments on this PR and do a two step process in a sub-PR:
1) Figure out which commentary you can address by direct code changes, vs which issues are open questions/future work.
2) Address the addressable comments with direct code changes in a sub-PR.

src/hyrax/datasets/data_provider.py
```diff
@@ -808,51 +846,81 @@ def resolve_data(self, idx: int) -> dict[str, dict[str, Any] | str | None]:
...
+            if fields_data is None:  # left outer join miss
+                augmented_data[friendly_name] = None
+                continue
+            new_fields: dict[str, Any] = {}
+            for field, value in fields_data.items():
+                augment_fn = self.augment_getters.get(friendly_name, {}).get(field)
+                new_fields[field] = augment_fn(value, idx, rng_seed) if augment_fn is not None else value
+            augmented_data[friendly_name] = new_fields
+        tensorboardx_logger.log_duration_ts(f"{prefix}/augmentation_s", augment_start)
+        return augmented_data
```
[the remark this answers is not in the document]
mtauraso: @drewoldag I think copilot is right here?

src/hyrax/verbs/train.py
drewoldag: My hunch is yes, we should probably dispatch this for other verbs. Perhaps we should also give an indication of which verb the is making the call.

i.e. `provider.on_epoch_start('train')` or something more intelligent. Since there may be distinct actions that the user would take depending on the verb being used.

src/hyrax/datasets/data_provider.py
```diff
@@ -312,6 +312,23 @@ def __init__(self, config: dict, request: dict):
...
+        self.augment_enabled = {}  # friendly_name -> bool
+        self._has_any_augmentation = False
+        # _augment_rng advances once per epoch (in on_epoch_start) to produce a fresh
+        # _epoch_rng for that epoch.  _epoch_rng is drawn from sequentially in
+        # resolve_data — one integer per call — so call order within an epoch determines
+        # the seed sequence.  Single-threaded access is reproducible; multi-threaded
+        # access is intentionally non-reproducible (no locks on this hot path).
+        # config["data_set"]["seed"] uses false as the Hyrax sentinel for "not set";
+        # treat it as None so numpy seeds from OS entropy rather than silently using 0.
+        _raw_seed = config.get("data_set", {}).get("seed", None)
```
drewoldag:
```suggestion
        _raw_seed = config["data_set"]["seed"]
```

src/hyrax/datasets/data_provider.py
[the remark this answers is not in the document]
drewoldag: The first part of the suggestion, is the correct thing to do. 

The second part seems a little odd. I think that on line 924, you can get away with just changing `idx` -> `dataset_idx`. I don't understand the need for `value = value.copy()` in the suggestion.

drewoldag (approved):
Overall looking very good. Left one comment about the use of config.get(...) and responded to the question about whether copilot was making the correct suggestion (it was half correct, I believe)

src/hyrax/datasets/data_provider.py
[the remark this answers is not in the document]
mtauraso: Its worried about `value` being passed to `augment_fn` and mutated, because the cache would store the mutated value.

I'm going to go see how numpy arrays handle `.copy` to see what we want to do here.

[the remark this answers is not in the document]
mtauraso: Ugly. What I'd eally want is Copy-on-write for whatever we pass into augment, so the dataset author just can't screw up the cache, and numpy basically doesn't support that. 

What I can do, is create a view (cheap), make the view read-only, and pass the view. They can always make it writable again but that will at least prevent inadvertent cache corruption.

src/hyrax/verbs/train.py
mtauraso: I like this plan.

mtauraso:
IMO this is ready to merge, but I want to read the tests once more to make sure there isn't something terrible lurking there.

PR #944, opened 2026-06-10
[auto-merge] Remove old split configs

mtauraso:


src/hyrax/pytorch_ignite.py
```diff
@@ -82,6 +84,26 @@ def setup_dataset(
...
+            "  dataset_class = 'YourDataset'\n"
+            "  data_location = '/path/to/data'\n"
+            "  primary_id_field = 'id'\n"
+            "  split_fraction = 0.6\n\n"
+            "  [data_request.validate.data]\n"
+            "  dataset_class = 'YourDataset'\n"
+            "  data_location = '/path/to/data'\n"
+            "  primary_id_field = 'id'\n"
+            "  split_fraction = 0.2\n\n"
+            "For more information, see: https://hyrax.readthedocs.io/"
```
mtauraso: This should point to something dataset specific

docs/core_concepts.rst
```diff
@@ -12,5 +12,6 @@ to apply these concepts in tutorial notebooks in the :doc:`Common workflows </co
 
    Required inputs <required_input>
    Configuration system <configuration_system>
+   Dataset splits <dataset_splits>
```
mtauraso, line 15: Note for reviewers: I steered this in, because I think the doc is really important for new users and much more clear now than it was before.

src/hyrax/verbs/train.py
```diff
@@ -108,93 +107,15 @@ def run(self):
```
mtauraso, line 107: @drewoldag This section is primarily where I want the most review.

tests/hyrax/test_train.py
```diff
@@ -295,6 +237,63 @@ def test_train_split_fraction(tmp_path):
...
+        },
+    }
+    h.config["data_set"]["HyraxRandomDataset"]["size"] = 30
+    h.config["data_set"]["HyraxRandomDataset"]["seed"] = 42
+    h.config["data_set"]["HyraxRandomDataset"]["shape"] = [2, 3]
+    h.config["data_set"]["seed"] = 42
+
+    h.train()
+
+    # Recreate datasets the same way train.py does, then ask dist_data_loader
```
mtauraso: I don't like this, but I'm not sure there's actually a better way.

docs/dataset_splits.rst
drewoldag, line 1: Would it be possible to slim this down a bunch, hit only the high points, and point people to a substantial example in "Common Workflows"? This doc isn't huge currently, but maybe the thrust of it is, "there is a lot of power in splitting, and we can't conver it all here, but if you look _over here_ you'll see a lot more."

src/hyrax/pytorch_ignite.py
```diff
@@ -218,189 +231,13 @@ def make_sampler(indexes: Sequence[int], sampler_shuffle: bool):
             return SubsetRandomSampler(indexes, generator=torch_rng)
         return SubsetSequentialSampler(indexes)
 
-    # Handle case where no split is needed.
```
drewoldag, line 221: Great! Cleaning this up feels so nice!

tests/hyrax/test_train.py
```diff
@@ -295,6 +237,63 @@ def test_train_split_fraction(tmp_path):
...
+    # for the actual indices it would sample — these are what the dataloaders
+    # consumed during training.
+    dataset = setup_dataset(h.config, splits=("train", "validate"))
+    _, train_indices = dist_data_loader(dataset["train"], h.config, shuffle=False)
```
drewoldag: I don't think we need to create the dist_data_loaders here, `dataset` is a dictionary with `train` and `validate` keys, each with a DataProvider instance. And each of those instances should have a `split_indices` attribute that could be used to ensure the sets are disjoint.

src/hyrax/pytorch_ignite.py
```diff
@@ -218,189 +231,13 @@ def make_sampler(indexes: Sequence[int], sampler_shuffle: bool):
...
+    if isinstance(dataset, DataProvider) and dataset.split_indices is not None:
+        indexes = dataset.split_indices
 
-    split_inds = {"train": train_idx, "test": test_idx}
-    if validate_size:
-        split_inds["validate"] = valid_idx
+    sampler = make_sampler(indexes, shuffle)
 
-    return split_inds
+    return idist.auto_dataloader(dataset, sampler=sampler, **data_loader_kwargs), indexes
```
drewoldag, line 240: I _think_ that we can have this just return the dataloader. It seems that the only place where the indexes are used is in tests. The actual code doesn't care about it. 

No need to make adjustments to this PR about this. But if the indexes are only returned in order to support tests, then when the split+balance PR happens, we should probably remove the return, since this function will no longer be driving split generation.

drewoldag (approved):
Overall this looks really nice. I'm glad that we're removing so much old code.

The comment about removing the returned `indexes` from `dist_data_loader` is more for me in the splits+balance PR.

tests/hyrax/test_train.py
mtauraso: Yeah, that works.

docs/dataset_splits.rst
mtauraso: Yeah the data_requests common workflow does really overlap with this. I'll see about slimming it down and linking there.

PR #945, opened 2026-06-11
Update lance-related deps: Use pyarrow based writing to preserve NaNs 

mtauraso:
### Motivation
- LanceDB's `table.add()` path in versions >= 0.30 rejects NaN values in fixed-size-list/vector columns, so writing through `pylance` is needed to preserve NaNs and maintain the existing fixed-size-list on-disk format.
- Upstream package compatibility changed around `lance-namespace` and `pylance`, so dependency caps were adjusted to track a compatible set of releases until broader support is available.

### Description
- Add `import lance` and switch `ResultDatasetWriter.write_batch` to write Arrow batches via `lance.write_dataset(...)` instead of creating an empty table and using `table.add()` through `lancedb`, preserving fixed-size-list schemas and NaNs.
- Track the current dataset handle via `self.lance_dataset` and `self.lance_uri` and append subsequent batches with mode switching (`overwrite` for the first write, `append` thereafter).
- Change `commit` to call `lancedb.connect(...).open_table(TABLE_NAME).optimize()` for optimization after writes. 
- Update `pyproject.toml` dependency pins to `lancedb < 0.34.0`, `lance-namespace >= 0.7.7, < 0.8.0`, and `pylance >= 7.0.0, < 8.0.0` to reflect the compatibility window used by the new write path. 
- Update tests in `tests/hyrax/test_result_dataset.py` to import `pyarrow as pa` and validate that the stored `data` field is a `fixed_size_list` with the expected list size.

### Testing
- Ran the updated unit tests for the result dataset file with `pytest tests/hyrax/test_result_dataset.py` and the tests completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a2b0ed194588331a97c3691b24b8b52)

### DIFF
--- a/src/hyrax/datasets/result_dataset.py
+++ b/src/hyrax/datasets/result_dataset.py
@@ -14,6 +14,7 @@ from typing import Union
 if "LANCE_LOG" not in os.environ:
     os.environ["LANCE_LOG"] = "error"
 
+import lance
 import lancedb
 import numpy as np
 import pyarrow as pa
@@ -29,8 +30,8 @@ LANCE_DB_DIR = "lance_db"
 class ResultDatasetWriter:
     """Writer for Lance-based inference results.
 
-    Writes inference results incrementally to Lance format using table.add()
-    for each batch, avoiding memory accumulation.
+    Writes inference results incrementally to Lance format through pylance,
+    avoiding memory accumulation while preserving the existing fixed-size-list schema.
     """
 
     def __init__(self, result_dir: Union[str, Path]):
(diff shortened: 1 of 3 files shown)


src/hyrax/datasets/result_dataset.py
```diff
@@ -47,6 +48,8 @@ def __init__(self, result_dir: Union[str, Path]):
         self.lance_dir = self.result_dir / LANCE_DB_DIR
         self.db = None
         self.table = None
+        self.lance_dataset = None
+        self.lance_uri = self.lance_dir / f"{TABLE_NAME}.lance"
```
mtauraso, line 52: I feel like these have replaced `self.db` and `self.table` and I want to track through all of ResultDataset and see if something weird is going on with the change. I think we might be leaving dead code around.

PR #946, opened 2026-06-11
Augmentation granular config

mtauraso:
## Code Quality


src/hyrax/config_schemas/data_request.py
```diff
@@ -43,9 +43,16 @@ class DataRequestConfig(BaseConfigModel):
         description="Dataset-specific configuration as a free-form dictionary.",
     )
 
-    augment: bool | None = Field(
+    augment: bool | dict[str, bool] | None = Field(
```
drewoldag: I should have thought about this in the design doc, but what is the reasoning for this being a dict of field:boolean? Would a list of field names suffice and help reduce the nesting? 

If a field name is present in the augment list, use the augment_<field> method, if not, use the get_<field> method. Use the user provided or default list of fields as the source of fields to check these against. 

Error when:
* a field is in the augment list, but not the fields list
* a field is in the augment list and doesn't have a augment_<field> method

tests/hyrax/test_data_request_config.py
drewoldag, line 1: I didn't see a test in here for behavior of `augment` when `fields` is not present in the data_request.

drewoldag (approved):
This looks good to me and I'll mark it as approved. I do think it would be worth thinking through whether we force the user to write a dictionary of {field_name:bool} for each field, or if it would suffice to only require a list of fields that should be augmented. 

Also, I think there's room for one more test for when `fields` isn't defined on the data_request.

src/hyrax/config_schemas/data_request.py
mtauraso: Yeah that makes a ton more sense honestly. Way simpler than what's in the spec.

tests/hyrax/test_data_request_config.py
mtauraso: Added now in test_augmentation.py note that the `make_dp` helper there does not automatically add `fields` to the data request.

src/hyrax/config_schemas/data_request.py
mtauraso: This is now implemented.

drewoldag (approved):
New changes look good to me.

PR #947, opened 2026-06-11
Update lance related deps: Handle NaN values with pyarrow based writing.

mtauraso:
### Motivation
- LanceDB >= 0.30 rejects NaN values in fixed-size-list vector columns, so writing must use pylance to preserve NaNs and keep the existing fixed-size-list schema. 
- Align `lance-namespace` and `pylance` version constraints with the pylance runtime that supports the required namespace behavior.

### Description
- Pin dependencies in `pyproject.toml`: set `lancedb < 0.34.0`, `lance-namespace >= 0.7.7, < 0.8.0`, and `pylance >= 7.0.0, < 8.0.0` to ensure compatibility with pylance-based Arrow writes. 
- Add `import lance` and change `ResultDatasetWriter` to convert batches to Arrow tables and write them with `lance.write_dataset(...)` instead of using `table.add(...)`, so the fixed-size-list `data` column and NaNs are preserved. 
- Remove creation of an empty Lance table via `lancedb.create_table` and instead consistently write Arrow tables with a schema created from the first sample tensor; validate tensor dtype and shape on subsequent batches. 
- In `commit()`, call `lancedb.connect(...).open_table(TABLE_NAME).optimize()` to optimize the final table. 
- Update tests in `tests/hyrax/test_result_dataset.py` to assert the created `data` column is a fixed-size-list with the expected `list_size` and to exercise multi-batch writes via `ResultDataset` loading.

### Testing
- Ran `pytest tests/hyrax/test_result_dataset.py` which exercises `test_writer_basic`, `test_writer_multiple_batches`, and multidimensional tensor writing, and all tests passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a2b0ed194588331a97c3691b24b8b52)


drewoldag (approved):
Looks ok to me.

PR #949, opened 2026-06-12
[Auto-merge] Augmentation: Per Dataset Caching

mtauraso:
## Code Quality


src/hyrax/config_migrations/migrations/005_remove_preload_config.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+@migration_step(from_version=4)
+def remove_preload_config(cfg: TOMLDocument) -> TOMLDocument:
+    """Remove deprecated preload_cache and preload_threads keys."""
+    data_set = cfg.get("data_set")
+    if isinstance(data_set, dict):
+        data_set.pop("preload_cache", None)
+        data_set.pop("preload_threads", None)
+    return cfg
```
mtauraso, line 45: I need to think about if we want to push values to various `data_loader` kwargs if they are not already set:
https://docs.pytorch.org/docs/2.12/data.html#torch.utils.data.DataLoader

src/hyrax/datasets/dataset_registry.py
```diff
@@ -131,6 +132,24 @@ def metadata_fields(self) -> list[str]:
         """
         return [] if self._metadata_table is None else list(self._metadata_table.colnames)
 
+    def row_cache_key(self, idx: int, rng_seed: np.int64 | None = None) -> np.int64 | None:
```
mtauraso: I don't like this because when augmentation is turned on, but the child dataset class hasn't overridden this... this will get double-called once with `rng_seed=<value>` to look for augmented data, and once for `rng_seed=None`

I think we are better off either:
- Not defining this function at base level, and only having the slowdown above be possible if someone implements the function
- Revisit the interface we're providing here somehow, so the caching system doesn't have a slow path for key generation.

tests/hyrax/mocks/lsst_butler_fixtures.py
```diff
@@ -72,7 +72,6 @@ def lsst_config(sample_catalog_saved):
             "transform": "tanh",
             "crop_to": [100, 100],
             "use_cache": True,
-            "preload_cache": True,
```
mtauraso, line 75: TODO: look at why we had this on in this function.

tests/hyrax/test_lsst_butler_mocks.py
```diff
@@ -75,7 +75,6 @@ def lsst_config():
             "transform": "tanh",
             "crop_to": [100, 100],
             "use_cache": True,
-            "preload_cache": True,
```
mtauraso, line 78: TODO look at why this was on previously!

src/hyrax/datasets/dataset_registry.py
mtauraso: I'm making the interface `augment_cache_key` only called when caching augmented data.

Normal data has to get by simply being cached by index. We don't really support cache locality in any way (nor are we likely to with the sampler-based approach for indexing) so the ability to define your own cache key for a normal getter I think does not have value to a dataset author.

src/hyrax/datasets/data_provider.py
```diff
@@ -869,95 +890,73 @@ def resolve_data(self, idx: int) -> dict[str, dict[str, Any] | str | None]:
...
+            else:
+                object_id = result[self.primary_dataset][self.primary_dataset_id_field_name]
+            result["object_id"] = str(object_id)
 
-            self.data_cache.insert_into_cache(idx, base_data)
+        # Timing metrics.
+        if had_any_miss:
             tensorboardx_logger.log_duration_ts(f"{prefix}/cache_miss_s", start_time)
+        else:
+            tensorboardx_logger.log_duration_ts(f"{prefix}/cache_hit_s", start_time)
```
mtauraso, line 965: TODO (for a later PR probably):
- Split out augment vs data cache
- Maybe: write a notebook demonstrating how to use these values

mtauraso:
I'm resolving the config migration conflict by putting all the config migration tests in their own files. This will increase the size of the diff, but I think its the right way to avoid this in the future.

tests/hyrax/test_augmentation.py
```diff
@@ -549,3 +577,129 @@ def test_on_epoch_start_called_during_test(tmp_path_factory):
...
+    assert already_aug is True
+
+    # Different rng_seed misses augmented but hits base
+    fetched, already_aug = cache.try_fetch("data", real_idx=0, rng_seed=np.int64(99))
+    assert fetched is base_data
+    assert already_aug is False
+
+
+def test_datacache_augment_cache_key_none_skips_augment_cache(tmp_path):
+    """augment_cache_key returning None means augmented data is not cached."""
```
drewoldag, line 644: This test seems reasonable, but the name and docstring don't make sense to me.

drewoldag (approved):
This looks really nice to me. Great work! Only one small comment about the name/docstring on one of the tests. But otherwise fine.

PR #951, opened 2026-06-15
1d light curve collation utility function

SamSandwich07:
Introduces `collate_utils.py` to put utility functions for collation. Adds one such utility function which can be used to pad and mask raw 1D light curves, along with a test for this function.


mtauraso:
Generally this seems alright to merge. I worry that it is not generic enough to be generally useful (but perhaps I simply haven't been part of the discussions here)

I'm lightly suggesting that we try to talk to folks who would use these collation utilities and try to generalize this, add documentation and examples so we get the full value here. I think a lot of this is human conversational/requirements gathering work, likely leading to a relatively small amount of new code. Not sure if we're on that path yet, but it seems like one we should consider.

mtauraso (approved):

src/hyrax/datasets/collate_utils.py
```diff
@@ -0,0 +1,40 @@
...
+        key passed in for the `field` argument
+    field
+        The field to collate
+
+    Returns
+    --------
+    dict
+        Contains three keys: `<field>`, `<field>_length`, and `<field>_mask`
+        `field` - float32 array (batch, max_len) containing the padded light curves
+        `<field>_length` - int64 array (batch) of true light curve lengths
```
drewoldag, line 20:
```suggestion
        `<field>_lengths` - int64 array (batch) of true light curve lengths
```

drewoldag (approved):
Seems fine to me. I can imagine that users might want to be able to specify a max length as well. To either truncate all light curves with length greater than X, or to ensure that all light curves have length X even if the max in the batch is <X. 

Regardless this is a nice jumping off point for people.

PR #953, opened 2026-06-19
Adding `.get(...)` to fallback in splitting_utils._resolve_seed.

drewoldag:
## Solution Description
The `_resolve_seed` function in `splitting_utils` assumes a fully resolved configuration file. But when the splitting machinery is checking previously generated splits for matches, it doesn't necessarily have the full config, instead it may only have access to the `split_config.toml` file. 

The easiest option was to simply provide a default value for `[data_set][seed]` when it isn't available. The other option would be to save [data_set][seed] in the split_config.toml file. That may still be the correct answer, but this felt like a fine stop gap for the moment. 

For reference an example split_config.toml file looks like this: 
```toml

[data_request.train.data]
dataset_class = "HyraxCifarDataset"
data_location = "/Users/drew/code/hyrax/docs/pre_executed/data"
fields = ["image", "label"]


[data_request.validate.data]
dataset_class = "HyraxCifarDataset"
data_location = "/Users/drew/code/hyrax/docs/pre_executed/data"
fields = ["image", "label"]


[data_request.test.data]
dataset_class = "HyraxCifarDataset"
data_location = "/Users/drew/code/hyrax/docs/pre_executed/data"
fields = ["image", "label"]

[split]
# Per-group split configuration.  Keys correspond to groups in [data_request].
# Values are a fraction of the group's primary dataset to use — (0.0, 1.0]
# or a path to a previously generated <group>_split.npz file.
# Groups absent from this table default to 1.0 (use the full dataset).

# RNG seed for shuffling/partitioning. `false` uses config["data_set"]["seed"].
# A non-empty value uses a dedicated generator and does affect global RNG state.
rng_seed = false
train = 0.7
validate = 0.2
test = 0.1


[balance]
field = "label"
groups = ["train"]

[balance.distribution]
cat = 0.5
dog = 0.3
boat = 0.2

[label]
# Optional. Maps human-readable string aliases to the raw values returned by
# get_<balance.field>.  Required when balance.distribution is used and 
# get_<balance.field> returns non-string values (e.g. integers),
# All balance.distribution keys must appear here.
# Example:
#   cat  = 0
#   dog  = 1
#   bird = 2
cat = 0
dog = 1
boat = 2
```


mtauraso (approved):
Approve as stop-gap. 

I do want us to think about the right solution here, so we can avoid the `.get()`

drewoldag:
My take at this point is that we should just ignore the data_set.seed value here, and only pay attention to the split.rng_seed. If it's set, use it, if not, just behave as though the user wants a true random value.

PR #956, opened 2026-06-22
Updating docs and notebooks with new split syntax.

drewoldag:
Small notebook updates and tweaks to the .rst pages to change `split_fraction` to `[split]` table.


aritraghsh09 (changes requested):
This looks good to me in general. 

In [getting_started_hello_world.rst](https://github.com/lincc-frameworks/hyrax/pull/956/changes#diff-2423add0686d88946778542370fc61197293401c74f9e2e6cc1522f3d845b64d), it got rid of the split command without replacing it (because the default is 1). But given that this is the "first" notebook people are supposed to interact with, we should probably put the new config in there and set it to 1, even though that is the default, just to make readers aware of how to do it.

drewoldag:
@aritraghsh09 I updated the getting started docs, let me know what you think.

mtauraso (approved):
LGTM

aritraghsh09 (approved):
Looks good now. Thanks!

PR #957, opened 2026-06-22
Clean up RNG ownership/WRS usage

mtauraso:
- Pytorch ignite code has one path that consistently uses WeightedRandomSampler
- Trace is simplified by using WRS for early exit rather than hooking
- Some tests altered to stop using split_fraction (there are many more though)
- Several tests that were implementation-dependent were deleted
- Removed places where we set torch default device that were causing test
  issues. We believe this setting was necessary for an early version of
  the onnx code, but is no longer
  
  Note base branch is set to `augmentation-v3-dataset-caching` (currently waiting for review) 
  
  I think this change could be cherry-picked down to main if we need it before the other branch is reviewed.


src/hyrax/pytorch_ignite.py
```diff
@@ -208,24 +210,26 @@ class distribution.
...
+            n = trace_limit
+        else:
+            # Limit via Subset
+            sub_dataset = Subset(sub_dataset, list(range(trace_limit)))
+
+    sampler = (
+        WeightedRandomSampler(weights=weights, num_samples=n, generator=torch_rng, replacement=True)
+        if shuffle
+        else None
+    )
```
drewoldag, line 232: Nice, very clean

drewoldag (approved):
This looks good. I like that dist_data_loader is getting smaller all the time.

PR #963, opened 2026-06-27
Mtauraso/alt dataset

mtauraso:
## Code Quality

### DIFF
--- a/src/hyrax/datasets/data_provider.py
+++ b/src/hyrax/datasets/data_provider.py
@@ -334,6 +334,9 @@ class DataProvider:
         self.split_indices = None
         self.split_weights = None
 
+        # Getitem mode: friendly names of datasets using __getitem__ instead of get_* methods.
+        self._getitem_datasets: set[str] = set()
+
         # Join support: populated by _build_join_indices after prepare_datasets.
         # Maps friendly_name → join_field name for datasets that use joining.
         self._join_fields: dict[str, str] = {}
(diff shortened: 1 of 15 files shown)


src/hyrax/datasets/data_provider.py
mtauraso, line 507: I think augment, rng, and caching can all maybe just be moved into this constructor if we pass appropriate config along.

However this would mean config other than the dataset-specific part of the data request would get passed in as "dataset config" which is already this weird amalgum.

I think we hit two birds with one stone here:
- HyraxDataset gets an initialize which receives various config elements and essentially does __init__ like things that are split in this diff between `HyraxDataset.__init__` and this preparation function.
- We remove the requirement on the dataset author to call `super.__init__(config)` Their __init__ just always runs before Hyrax's stuff, because we call our initializer here.

We're less brittle to Dataset implementation errors, and DataProvider is more focused on the parts of hyrax that truly cross dataset boundaries (joins, collation, etc) and less of a grab bag of everything related to data handling.

src/hyrax/datasets/dataset_registry.py
mtauraso, line 143: It is probably time for metadata to truly die and be completely removed from the codebase.

src/hyrax/datasets/data_provider.py
```diff
@@ -919,50 +890,27 @@ def resolve_data(self, idx: int) -> dict[str, dict[str, Any] | str | None]:
...
-                object_id = result[self.primary_dataset][self.primary_dataset_id_field_name]
+                object_id = self.get_object_id(idx)
             result["object_id"] = str(object_id)
 
-        # Timing metrics.
-        if had_any_miss:
-            tensorboardx_logger.log_duration_ts(f"{prefix}/cache_miss_s", start_time)
-        else:
-            tensorboardx_logger.log_duration_ts(f"{prefix}/cache_hit_s", start_time)
+        tensorboardx_logger.log_duration_ts(f"{prefix}/resolve_data_s", start_time)
```
mtauraso, line 913: We need to bring back cache_hit_s and cache_miss_s and their augment counterparts.

This probably means datasets will end up needing a tensorboardx_logger somehow.

PR #974, opened 2026-07-02
Single device multi gpu support

SamSandwich07:
We thought Hyrax already support distributed training across multiple GPUs on one node (computer)... apparently not, surprise! 

This PR aims to put that functionality back in by using `ignite.distributed`'s `DistributedDataParallel` to use as many GPUs as are available on a single device.


SamSandwich07:
All PR checks finally pass!!! Hooray!!! Requesting a review on the current state of this. Following things are what stand out to me as causes for concern:

1) To pass the GitHub checks, I had to make Hyrax use only 1 device if it was not on `cuda` which means distributed `cpu` and `mps` support is not available. I tried using `torch.multiprocessing.cpu_count()` but that didn't work on GitHub or locally because the backend has to be `gloo` and not `nccl`. I changed that and still ran into issues but I haven't really investigated this deeply.

2) More pressing than issue 1: there is no good system to copy over attributes from the user's model to the DDP model (see `pytorch_ignite.py`. The short version of this issue is that in the user's `train_batch` function, they may use several attributes that they have defined, including `optimizer`, `scheduler`, `grad_clip`, etc. We don't necessarily know what these will be named, but in order for the current implementation to work there has to be some way to copy over these attributes to the DDP model _or_ automatically pull them from the underlying model if not found on the DDP model.

    I tried to modify the `__getattr__` function to attempt this second method already but it didn't work (weird python semantics :/). For the first way, we could allow the user to simply specify the attributes somewhere or attempt to dig in and figure out how to extract used attributes from a function. I welcome feedback on these and/or alternative ideas.

SamSandwich07:
More info regarding issue 1 from previous comment: even if the backend is update to use `gloo` for `cpu`/`mps`, we run into pickling errors because of nested functions. This happens not only in `train.py` (which is an easy fix), but even in the model class (including something as simple as `hyrax_loopback`)

src/hyrax/models/model_registry.py
```diff
@@ -203,8 +203,10 @@ def hyrax_model(cls):
 
     def wrapped_init(self, config, *args, **kwargs):
         original_init(self, config, *args, **kwargs)
+        # self._hyrax_data_parallel = None
```
drewoldag:
```suggestion

```

src/hyrax/pytorch_ignite.py
```diff
@@ -664,9 +692,20 @@ def create_trainer(model: torch.nn.Module, config: dict, results_directory: Path
     pytorch-ignite.Engine
         Engine object that will be used to train the model.
     """
+
+    if idist.get_nnodes() > 1:
+        raise RuntimeError("Multi-node configurations not supported")
```
mtauraso, line 692: We might want to pull this if we're moving on to MPI/multi-node prototyping. Might be worth a comment.

src/hyrax/pytorch_ignite.py
```diff
@@ -664,9 +692,20 @@ def create_trainer(model: torch.nn.Module, config: dict, results_directory: Path
...
+
     device = idist.device()
     model.train()
     wrapped_model = idist.auto_model(model)
+
+    wrapped = type(wrapped_model) is DistributedDataParallel or type(wrapped_model) is DataParallel
```
mtauraso, line 700: Should it be an error to have a `DataParallel` model rather than `DistributedDataParallel` model?

src/hyrax/pytorch_ignite.py
```diff
@@ -31,6 +34,31 @@
 _LEGACY_SPLIT_KEYS = ("train_size", "validate_size", "test_size")
 
 
+# This is a near copy-paste of the __getattr__ method on torch.nn.module
```
mtauraso: This might be the best way, simply accepting the risk that `torch.nn.module.Module.__getattr__` changes and updating accordingly, but we might be able to mitigate

Consider:
```
_old__getattr__ = torch.nn.module.Module.__getattr__

def _new__getattr__(self, name:str) -> Union[Tensor, "Module"]:
    try:
        return _old__getattr__(self, name)
    except AttributeError:
        if hasattr(self.module, name):
            return getattr(self.module, name)
        raise

DistributedDataParallel.__getattr__ = _new__getattr__
```

If this works, it would reduce the area of code we have to duplicate to just the type annotation (which you could omit)

My hope is that the first line will prevent any infinite loop; however, we might need to do something like `copy` the original function somehow.

src/hyrax/verbs/train.py
```diff
@@ -86,88 +88,103 @@ def run(self):
...
+
+            # Save the trained model
+            model.save(results_dir / config["train"]["weights_filename"])
+            monitor.stop()
+
+            logger.info("Finished Training")
+            close_tensorboard_logger()
+
+        nproc_per_node = torch.cuda.device_count()
+        if nproc_per_node > 1:
```
mtauraso, line 97: Do we absolutely need to separate these cases?

src/hyrax/verbs/train.py
```diff
@@ -86,88 +88,103 @@ def run(self):
...
-        logger.info(f"{Style.BRIGHT}{Fore.BLACK}{Back.GREEN}Training dataset(s):{Style.RESET_ALL}\n{dataset}")
-
-        # If a pre-trained weights file is specified, load it before creating the trainer.
-        # This must happen before create_trainer() wraps the model with idist.auto_model
-        # (the distributed wrapper) to avoid parameter key name mismatches.
-        if config["train"]["model_weights_file"]:
-            from hyrax.models.model_utils import load_model_weights
 
-            load_model_weights(config, model, "train")
+        def training(rank):
```
mtauraso: Do we want to re-factor this to simply be a method on the train verb rather than an inline-defined function?

src/hyrax/pytorch_ignite.py
SamSandwich07: True, in train we should never have DataParallel. I can add that as a change. We might consider implementing our own version of `auto_model` so that DataParallel doesn't occur anywhere, including in other verbs.

src/hyrax/pytorch_ignite.py
SamSandwich07: This is something I had in mind as well and will definitely test. On the off-chance `torch.nn.module.__getattr__` is updated, we don't want to copy paste it _again_.

src/hyrax/verbs/train.py
SamSandwich07: Yes for two reasons
1) GitHub merge checks don't pass if they're not separated lol
2) the `idist.Parallel` case doesn't play nice with `mps` (how it plays with `cpu` is unknown). There are lot of pickling errors which can even come from user generated code. Separating these cases is how we can avoid using `idist.Parallel` except when we are actually on GPU.

src/hyrax/verbs/train.py
SamSandwich07: I tried that already when messing with something else. It works, so I can make it so if you and/or Drew feel strongly about it.

mtauraso (approved):
Overall this looks pretty good as a first pass.

Everything i'm commenting on could be addressed as follow up PRs

src/hyrax/verbs/train.py
mtauraso: I think it might be good both from a readability perspective, and also because you're going to end up with this same sort of pattern on other verbs (infer, test, etc) so the pattern of having a single-threaded startup and a maybe-multi-threaded worker function could actually end up living higher in the Verb class hierarchy at some future point, with each derived class just defining what work is done at each stage.

This might not be the Verb class per-se, like you might have Verb -> MultiThreadedVerb -> Train/Infer/test etc. and the parallelization logic lives in MultiThreadedVerb

src/hyrax/verbs/train.py
mtauraso: The second one is probably worth a comment.

src/hyrax/verbs/train.py
drewoldag: Agreed with this.

src/hyrax/pytorch_ignite.py
drewoldag: +1 on this

drewoldag (approved):
I like what I'm seeing. The comments that are already here deserve to be addressed. But this is definitely moving the right direction. I would like to know what happens when you run `train` with multiple gpus when there is a train/validate split in the dataset.

src/hyrax/verbs/train.py
SamSandwich07: Okay, for now I'll just move it as a static method on `Train`

src/hyrax/pytorch_ignite.py
SamSandwich07: Actually, looks like non-cuda environments will still trigger DataParallel wrapper on this since that's what the code used to do. I'll table this and address it in a future PR with a custom `auto_model` function probably.

PR #976, opened 2026-07-09
Refactor custom collation construction to use functools.partial instead of nested function

### LINKED ISSUE #975: Setting num_workers > 0 results in pickling error
**Bug report**
Trying to create a DataLoader with multiple workers fails because hyrax's `collate` function in `data_provider.py` is not picklable and cannot be passed to worker processes. There is a nested function structure inside `collate` function:
```python
def make_dataset_collate(field_collate_functions: dict, friendly_name: str):
    def dataset_collate(samples: list[dict]) -> dict:
        ....
    return dataset_collate
``` 
Tried pulling the functions out as static methods, but since the nested structure is necessary for the function to be generated at runtime, the issue still remains. 

**Environment Information**
Using Python 3.12 for the notebook with `hyrax.__version__` being `0.8.2.dev58+g1bd1eda3f.d20260526`. Observed locally on a macOS machine. 

<details>
<summary>Traceback</summary>

```python
AttributeError: Can't get local object 'DataProvider.collate.<locals>.make_dataset_collate.<locals>.dataset_collate'
```

A warning from PyTorch:
[/opt/anaconda3/envs/hyrax/lib/python3.12/site-packages/torch/utils/data/dataloader.py:433](https://file+.vscode-resource.vscode-cdn.net/opt/anaconda3/envs/hyrax/lib/python3.12/site-packages/torch/utils/data/dataloader.py:433) : UserWarning: Got pickle error when attempting to start a worker Process. This might be because the worker Process arguments are not picklable. Python 3.14+ changed the multiprocessing start method in non-Mac POSIX platforms to 'forkserver', which requires the worker Process arguments to be picklable. You can also try multiprocessing.set_start_method('fork').

</details>

**Before submitting**
Please check the following:

- [x] I have described the situation in which the bug arose, including what code was executed, and any applicable data others will need to reproduce the problem.
- [x] I have included information about my environment, including the version of this package (e.g. `hyrax.__version__`)
- [x] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a description of what I expected instead.
- [ ] If I have a solution in mind, I have provided an explanation and/or pseudocode and/or task list.

SamSandwich07:
Avoids using nested functions when constructing custom collation functions to resolve issue #975.

### DIFF
--- a/src/hyrax/datasets/data_provider.py
+++ b/src/hyrax/datasets/data_provider.py
@@ -1129,12 +1129,13 @@ class DataProvider:
     @staticmethod
     def default_field_collate(samples: list[dict], field: str, friendly_name: str) -> dict:
         """Default field-level collate function for a single field.
-        
+
         Parameters
         ----------
         samples : list of dict
-            A list of data samples, where each sample is a dictionary mapping some attribute to one of its values.
-        
+            A list of data samples, where each sample is a
+            dictionary mapping some attribute to one of its values.
+
         field : str
             The name of the field to collate.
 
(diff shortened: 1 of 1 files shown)


drewoldag (approved):
This looks good to me. I tried it out locally by running the getting started notebook without this change using num_workers = 2, which reproduced the problem. I pulled down this branch, and repeated the experiment, and saw that the error no long appears.

Graciaaa3:
Did the same thing of running it locally with this branch and it runs without error :)

PR #978, opened 2026-07-13
Don't force every field into float32 for streaming data

### LINKED ISSUE #977: `streaming_data_provider` converts everything to float32
In PR #930 copilot flagged that `streaming_data_provider` was trying to convert everything into a float32. I attempted a quick fix, but it caused lots of tests to fail. Need to revisit this because clearly, not _every_ value from a kafka stream will be a float. 

The copilot comment thread:
https://github.com/lincc-frameworks/hyrax/pull/930/changes#r3545563135

drewoldag:
Do not attempt to force every single field into a float32 for streaming data.

## Change Description
Closes #977 

## Solution Description
Taking the suggested change from copilot. Previously this caused many tests to fail. I assume that my environment was in an odd state because all tests pass now.

### DIFF
--- a/src/hyrax/datasets/streaming_data_provider.py
+++ b/src/hyrax/datasets/streaming_data_provider.py
@@ -121,7 +121,10 @@ class StreamingDataProvider(CollationMixin, torch.utils.data.IterableDataset):
         if not self.fields:
             self.fields = [key for key in sample if key != self.primary_id_field]
             self._register_field_collate_hooks()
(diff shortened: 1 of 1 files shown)


SamSandwich07 (approved):
This looks good to me, I'm assuming that whatever broke on your previous fix isn't breaking with this fix?

drewoldag:
Yep that's correct. My environment must have been in an odd state when I tried this fix the first time around.

PR #982, opened 2026-07-14
Kafka Streaming Credentials

maxwest-uw:
## Change Description
allows the user to pass along extra credentials/options to the `confluent_kafka` instance for `stream_infer`.

## Solution Description
same technique as the hsc download, allows user to specify a credentials .ini file with extra.

Also added a unit test.

## Code Quality

### DIFF
--- a/src/hyrax/datasets/kafka_stream_dataset.py
+++ b/src/hyrax/datasets/kafka_stream_dataset.py
@@ -73,15 +73,16 @@ class KafkaStreamDataset(HyraxDataset, torch.utils.data.IterableDataset):
         ds_config = config["data_set"]["KafkaStreamDataset"]
 
         # ``data_location``, when given, is a Kafka URI of the form
-        # ``kafka://<host>:<port>/<topic>`` supplied inline by the data_request. It takes
+        # ``kafka://<host>:<port>[/<topic>]`` supplied inline by the data_request. It takes
         # precedence over the [data_set.KafkaStreamDataset] config; anything the URI omits
         # falls back to that config block.
         host_port = ""
         topic = ""
         if data_location:
             parsed = urlparse(data_location)
-            host_port = parsed.netloc  # "broker.example.org:9092"
-            topic = parsed.path.lstrip("/")  # "lsst_topic"
+            if parsed.scheme == "kafka":
+                host_port = parsed.netloc  # "broker.example.org:9092"
+                topic = parsed.path.lstrip("/")  # "my-topic"
 
         self.bootstrap_servers = host_port or ds_config["bootstrap_servers"]
         self.topics = topic or ds_config["topics"]
@@ -147,10 +148,7 @@ class KafkaStreamDataset(HyraxDataset, torch.utils.data.IterableDataset):
         try:
             from confluent_kafka import Consumer
         except ImportError as err:
-            raise ImportError(
-                "KafkaStreamDataset requires the 'confluent-kafka' package. "
-                "Install it with the streaming extra: pip install 'hyrax[stream]'."
-            ) from err
+            raise ImportError("KafkaStreamDataset requires the 'confluent-kafka' package. ") from err
 
         consumer = Consumer(self.consumer_config)
 
--- a/src/hyrax/datasets/kafka_stream_dataset.py
+++ b/src/hyrax/datasets/kafka_stream_dataset.py
@@ -48,9 +48,11 @@ provider is what is passed to the DataLoader; configure it the normal way:
 
 import json
 import logging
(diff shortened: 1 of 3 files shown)


src/hyrax/datasets/kafka_stream_dataset.py
```diff
@@ -76,16 +73,15 @@ def __init__(self, config: dict, data_location=None):
...
         # falls back to that config block.
         host_port = ""
         topic = ""
         if data_location:
             parsed = urlparse(data_location)
-            if parsed.scheme == "kafka":
-                host_port = parsed.netloc  # "broker.example.org:9092"
-                topic = parsed.path.lstrip("/")  # "my-topic"
+            host_port = parsed.netloc  # "broker.example.org:9092"
+            topic = parsed.path.lstrip("/")  # "lsst_topic"
```
drewoldag: I think that we need to keep the check for `parsed.scheme == "kafka"` here.

src/hyrax/datasets/kafka_stream_dataset.py
```diff
@@ -138,16 +147,15 @@ def _make_consumer(self):
         try:
             from confluent_kafka import Consumer
         except ImportError as err:
-            raise ImportError("KafkaStreamDataset requires the 'confluent-kafka' package. ") from err
+            raise ImportError(
+                "KafkaStreamDataset requires the 'confluent-kafka' package. "
+                "Install it with the streaming extra: pip install 'hyrax[stream]'."
```
drewoldag: I don't think we need this second line in the error. I believe that `confluent-kafka` was added to the base list of dependencies.

src/hyrax/datasets/kafka_stream_dataset.py
```diff
@@ -76,16 +73,15 @@ def __init__(self, config: dict, data_location=None):
         ds_config = config["data_set"]["KafkaStreamDataset"]
 
         # ``data_location``, when given, is a Kafka URI of the form
-        # ``kafka://<host>:<port>[/<topic>]`` supplied inline by the data_request. It takes
+        # ``kafka://<host>:<port>/<topic>`` supplied inline by the data_request. It takes
```
drewoldag: `<topic>` is optional at this point, the user can define it in the URL string, or provide it as a list of topics in the data_request. I think that this change can be reverted.

drewoldag (approved):
This looks good. Left a couple of quick comments that should be easy to address.

src/hyrax/datasets/kafka_stream_dataset.py
```diff
@@ -114,6 +111,19 @@ def __init__(self, config: dict, data_location=None):
         self._consumer = None
         self._buffered: list[dict] = []
 
+        credentials_file = ds_config.get("credentials_file")
```
mtauraso, line 114: Seems reasonable to roll our own, and in no way should this block merge. Asking because I don't know at all:
Is there any standard way to do a credential file we could be following that is germane to Kafka clients?

src/hyrax/hyrax_default_config.toml
```diff
@@ -383,6 +383,14 @@ poll_timeout = 1.0
 # Maximum seconds to wait before flushing a partial (smaller than batch_size) batch.
 batch_flush_timeout = 5.0
 
+# Provided path credentials file for a `confluent_kafka` consumer instance
```
mtauraso, line 386: Similar question as above, same low priority. Is there a standard location on unix systems for this sort of thing?

tests/hyrax/test_kafka_stream_dataset.py
```diff
@@ -229,3 +229,28 @@ def test_peek_sample_buffers_and_replays(monkeypatch):
...
+    ds_config = h.config["data_set"]["KafkaStreamDataset"]
+    ds_config["topics"] = "test-topic"
+    extra_credentials = {
+        "security.protocol": "SASL_PLAINTEXT",
+        "sasl.mechanism": "SCRAM-SHA-512",
+        "sasl.username": "user",
+        "sasl.password": "pass",
+    }
+
+    with open(tmp_path / "credentials.ini", "w") as f:
```
mtauraso, line 246: .toml?

mtauraso (approved):
Only minor nit comments and questions. Overall looks good.

src/hyrax/datasets/kafka_stream_dataset.py
maxwest-uw: the `babamul` client wants you to pass credentials through environment variables (and seems to be the general standard) but it causes some issues with configurability imo

PR #984, opened 2026-07-15
Benchmark on Gondor with self-hosted GitHub runner

### LINKED ISSUE #965: Implement basic performance benchmark
We received a report that the time needed to run training with Hyrax was ~2-3x longer compared to training with pure PyTorch.

It would be good to benchmark our performance. The most obvious case to me is training the HyraxCNN using the HyraxCifar dataset.

PyTorch documentation provides a very straightforward implementation (in fact that is what HyraxCNN and HyraxCifar as based on) here: https://docs.pytorch.org/tutorials/beginner/blitz/cifar10_tutorial.html

A highlevel understanding of the difference in training time for Hyrax vs. PyTorch would be very helpful. Ultimately we would want to be able to get very fine grained so that we can improve Hyrax performance as needed, but anything is better than nothing. As a first pass we would want to know whether there is a significant disparity (on the order of 2-3x slower) or not.

Graciaaa3:
## Solution Description
Initial experiment with self-hosted GitHub runner. Basically the goal is to verify that jobs can be executed on the self-hosted runner successfully before moving to Rivendell's GPU.

Closes #965


.github/workflows/asv-nightly.yml
```diff
@@ -86,6 +86,14 @@ jobs:
       with:
         path: ${{env.WORKING_DIR}}
         key: nightly-results-${{steps.nightly-dates.outputs.today}}
+    - name: Debug custom data upload
```
mtauraso: This should probably get `git stash` ed or something to pull onto your dev branch.

.github/workflows/pipeline_time_bench.yml
```diff
@@ -0,0 +1,83 @@
...
+        run: |
+          uv pip install --python .venv/bin/python -e ".[dev]"
+          if [ -f requirements.txt ]; then uv pip install --python .venv/bin/python -r requirements.txt; fi
+
+      - name: Run benchmark
+        id: benchmark
+        run: |
+          source .venv/bin/activate
+
+          # path to downloaded cifar dataset
```
mtauraso, line 43: Might be good to comment these are gondor specific paths, but not a merge blocker.

.github/workflows/publish-benchmarks.yml
```diff
@@ -11,7 +11,7 @@ on:
 jobs:
   publish-benchmarks-nightly:
     runs-on: ubuntu-latest
-    if: github.event.workflow_run.conclusion == 'success'
+    # if: github.event.workflow_run.conclusion == 'success'
```
mtauraso: This should probably come out.

benchmarks_self_hosted/timebench_utils.py
```diff
@@ -0,0 +1,363 @@
...
+        Learning rate for the optimizer.
+
+    Returns
+    -------
+    dictionary
+        A dictionary containing the benchmarking results
+    """
+    # setup
+    h = Hyrax()
+    h.set_config("model.name", "HyraxCNN")
```
mtauraso, line 43: Do we want the Hyrax version and the pytorch version to use exactly the same model architecture?

This looks like HyraxCNN and Net (below) could get out of sync.

benchmarks_self_hosted/timebench_utils.py
```diff
@@ -0,0 +1,363 @@
...
+    device = next(net.parameters()).device
+    device = torch.device(
+        torch.accelerator.current_accelerator().type if torch.accelerator.is_available() else "cpu"
+    )
+    net.to(device)
+
+    print("Model running on:", next(net.parameters()).device)
+
+    criterion = nn.CrossEntropyLoss()
+    optimizer = optim.SGD(net.parameters(), lr=lr, momentum=0.9)
```
mtauraso, line 238: We may want to pull out momentum (if not as a arg to these functions that we change, to a constant)

mtauraso (approved):
Think this is fine to merge because we need to get data flowing through the system and most of my comments (outside of the workflow lines that should get pulled) are pretty minor.

We should return to most of this stuff in a follow-up PR

benchmarks_self_hosted/timebench_utils.py
Graciaaa3: Left a comment to come back later

benchmarks_self_hosted/timebench_utils.py
Graciaaa3: Also left comments to come back later :)

PR #989, opened 2026-07-17
Remove metadata interface

mtauraso:
Removes the Metadata interface from datasets, Includes a quick fix to the only remaining consumer (visualizev1) so it uses the `get_` interface for everything.


src/hyrax/datasets/fits_image_dataset.py
```diff
@@ -126,9 +126,7 @@ def __init__(self, config: dict, data_location=None):
 
         # Relies on self.filters_ref and self.filter_catalog_table which are both determined
         # inside _init_from_path()
-        logger.debug("Preparing Metadata")
-        metadata = self._prepare_metadata()
```
mtauraso, line 130: Check if we need this line

PR #990, opened 2026-07-17
[AUTOMERGE] Unpinning tomlkit <= 0.14.0 

### LINKED ISSUE #988: Unpin tomlkit version after >0.15.0 is released
This is a follow up issue to #981 

The original issue cropped up due to a new version of tomlkit (0.15.0). I created an issue here https://github.com/python-poetry/tomlkit/issues/562 that was resolved with the PR here https://github.com/python-poetry/tomlkit/pull/563.

It is not clear how long it will take for the version of tomlkit that includes this fix to be released. In the mean time we have pinned tomlkit to version 0.14.0 in Hyrax in this PR https://github.com/lincc-frameworks/hyrax/pull/986

When the version of tomlkit that includes the fix is released, we should drop the version pin in Hyrax - tracking that is the purpose of this issue.

FWIW, we can "watch" for release from the tomlkit repo. We should be notified shortly after a new release is created, and we can simply drop the pin (given the that fix they implemented works correctly)

drewoldag:
The `Float has no len()` bug has been resolved and deployed in tomlkit==0.15.1.

## Change Description
Closes #988 

## Solution Description
Unpin the dependency in pyproject.toml.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,11 +29,7 @@ dependencies = [
     "pytorch-ignite <= 0.5.4", # Used for distributed training, logging, etc.
     "more-itertools", # Used to work around the issue in pytorch-ignite above
     "toml", # Used to load configuration files as dictionaries
-    # NOTE: tomlkit 0.15.0+ has a bug that causes it to parse arrays in an odd way
-    # that prevents them from being immediately converted to torch tensors. See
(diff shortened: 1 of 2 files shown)


mtauraso (approved):
🚢 it

PR #992, opened 2026-07-20
Remove slack notify from asv-nightly action

Graciaaa3:
## Change Description
The current asv-nightly workflow is failing because of a broken slack notification setup. I'm removing the slack notification section to see if that resolves the benchmark workflow failures.


drewoldag (approved):
Looks good to me

PR #993, opened 2026-07-21
Flag for distributed training

SamSandwich07:
Adds flag for distributed training to default config. May remove or change to default `true` in the future; implemented currently for development purposes


mtauraso (approved):
Approve. 

I think its correct that we don't have a config migration here. Its unclear to me if we're going to need a config migration later. I think we can just set `distributed=true` in future once we have more confidence in the feature.

PR #995, opened 2026-07-21
Publish nightly benchmark fix

Graciaaa3:
## Change Description
Put `workflow_dispatch` trigger to the publish workflow on the main branch for independent testing. Also try fixing the conditions that determine if publish steps runs.


drewoldag (approved):
Looks good

PR #996, opened 2026-07-21
Introduce Hyrax's own auto_model function to remove DataParallel

SamSandwich07:
Introduce a helper function `_auto_model` to `pytorch_ignite.py` to replace `idist.auto_model` and remove `DataParallel`. In other words, models will never be wrapped in `DataParallel` anymore; they will either be unwrapped or wrapped in `DistributedDataParallel`.

### DIFF
--- a/src/hyrax/pytorch_ignite.py
+++ b/src/hyrax/pytorch_ignite.py
@@ -409,9 +409,8 @@ def create_engine(funcname: str, device: torch.device, model: torch.nn.Module, c
 
 
 def extract_model_method(model, method_name):
-    """Extract a method from a model, which may be wrapped in a DistributedDataParallel
-    or DataParallel object. For instance, method_name could be `train_batch` or
-    `infer_batch`.
+    """Extract a method from a model, which may be wrapped in DistributedDataParallel.
+    For instance, method_name could be `train_batch` or `infer_batch`.
 
     Parameters
     ----------
--- a/src/hyrax/pytorch_ignite.py
+++ b/src/hyrax/pytorch_ignite.py
@@ -409,13 +409,13 @@ def create_engine(funcname: str, device: torch.device, model: torch.nn.Module, c
 
 
 def extract_model_method(model, method_name):
-    """Extract a method from a model, which may be wrapped in a DistributedDataParallel
-    or DataParallel object. For instance, method_name could be `train_batch` or
+    """Extract a method from a model, which may be wrapped in DistributedDataParallel.
+     For instance, method_name could be `train_batch` or
     `infer_batch`.
 
     Parameters
     ----------
-    model : nn.Module, DistributedDataParallel, or DataParallel
+    model : nn.Module
         The model to extract the method from
     method_name : str
         Name of the method to extract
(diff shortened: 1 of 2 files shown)


src/hyrax/pytorch_ignite.py
```diff
@@ -53,6 +53,31 @@ def _new__getattr__(self, name: str):
...
+            distributed only. Default, False. Note, if using Nvidia/Apex, batchnorm conversion should be
+            applied before calling ``amp.initialize``.
+        kwargs: kwargs to model's wrapping class (`torch DistributedDataParallel`_ )if applicable.
+            Please, make sure to use acceptable kwargs for given backend.
+
+    Returns:
+        torch.nn.Module
+    """
+    wrapped_model = idist.auto_model(model, sync_bn=sync_bn, **kwargs)
+    if type(wrapped_model) is DataParallel:
```
mtauraso, line 62: We could simply make this a RuntimeError. Thoughts?

src/hyrax/pytorch_ignite.py
```diff
@@ -53,6 +53,31 @@ def _new__getattr__(self, name: str):
...
 
 
+def _auto_model(model: torch.nn.Module, sync_bn: bool = False, **kwargs: Any) -> torch.nn.Module:
+    """Helper method to wrap a model in DistributedDataParallel if needed.
+    Uses idist.auto_model to do the wrapping, then ignore the result if it is a DataParallel instance.
+    For more info, see https://docs.pytorch.org/ignite/generated/ignite.distributed.auto.auto_model.html
+
+    Args:
+        model: model to adapt.
+        sync_bn: if True, applies `torch convert_sync_batchnorm`_ to the model for native torch
```
mtauraso: Do we know we will need to change `sync_bn` from callers of this function?

mtauraso: I guess I'm unclear why we're promoting it out of **kwargs in this interface.

If we need it set to false consistently, then we can put that, refactor when usage forces it?

src/hyrax/pytorch_ignite.py
SamSandwich07: No, that wouldn't work. `idist.auto_model` is going to make it `DataParallel` whenever we run Hyrax in an environment with multiple GPUs (like Gondor) but doesn't use `idist.Parallel` (i.e. set distributed to false or run a verb other than train). If we raise an error here then Hyrax will essentially always raise an error on Gondor and the like.

src/hyrax/pytorch_ignite.py
SamSandwich07: We are the only callers of `idist.auto_model`, and in our code we never call it with any arguments other than the torch model, so the answer for now is "we never change the argument".

SamSandwich07: As for why we promote it out of `**kwargs`, it's because `idist.auto_model` keeps it outside of `**kwargs` too, so I just used the same input signature (below taken from ignite docs):
```
def auto_model(model: nn.Module, sync_bn: bool = False, **kwargs: Any) -> nn.Module:
```

mtauraso (approved):
LGTM, 🚢 it
 (after tests get passing etc)

PR #997, opened 2026-07-21
make peek sample use consistent data

maxwest-uw:
## Change Description
in `hyrax_alerts`, we define and allow the user to define `pre_filter` and `post_filter` functions that might affect the data, but `peek_sample` (used in `hyrax_alerts.process_alerts()` to do initial runtime setup) doesn't currently touch those functions, so we should check if those functions have been defined so that `peek_sample` can maintain data consistency.

## Solution Description
check for the functions and then apply if them if available.

Note: I added the sample to `_buffered` **before** the sample is pre processed since we should assume that when it gets read again through `process_alerts` as part of the main batch process that it'll be mixed in with data in the `__iter__` that hasn't been pre processed yet.

## Code Quality

### DIFF
--- a/src/hyrax/datasets/kafka_stream_dataset.py
+++ b/src/hyrax/datasets/kafka_stream_dataset.py
@@ -203,8 +203,18 @@ class KafkaStreamDataset(HyraxDataset, torch.utils.data.IterableDataset):
             msg = consumer.poll(self.poll_timeout)
             if msg is not None and msg.error() is None:
                 sample = self._decode(msg)
-                self._buffered.append(sample)
-                return sample
+
+                # check if pre_filter is defined and apply it to the sample
+                if hasattr(self, "pre_filter"):
+                    sample = self.pre_filter([sample])
+                if sample:
+                    sample = sample[0] if isinstance(sample, list) else sample
+                    self._buffered.append(sample)
+                    # check if pre_process is defined and apply it to the sample
+                    if hasattr(self, "pre_process"):
+                        sample = self.pre_process([sample])
+                    return sample[0] if isinstance(sample, list) else sample
+
         raise RuntimeError("KafkaStreamDataset.peek_sample() stopped before a message arrived.")
 
     def __iter__(self):
--- a/tests/hyrax/test_kafka_stream_dataset.py
+++ b/tests/hyrax/test_kafka_stream_dataset.py
@@ -221,6 +221,8 @@ def test_peek_sample_buffers_and_replays(monkeypatch):
     messages = [_make_message("first", [[1.0]]), _make_message("second", [[2.0]])]
     _patch_consumer(monkeypatch, dataset, messages, stop_when_exhausted=True)
 
+    print(hasattr(dataset, "pre_filter"))
+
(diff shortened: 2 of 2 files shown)


src/hyrax/datasets/kafka_stream_dataset.py
```diff
@@ -203,8 +203,18 @@ def peek_sample(self) -> dict:
...
+                # check if pre_filter is defined and apply it to the sample
+                if hasattr(self, "pre_filter"):
+                    sample = self.pre_filter([sample])
+                if sample:
+                    sample = sample[0] if isinstance(sample, list) else sample
+                    self._buffered.append(sample)
+                    # check if pre_process is defined and apply it to the sample
+                    if hasattr(self, "pre_process"):
+                        sample = self.pre_process([sample])
+                    return sample[0] if isinstance(sample, list) else sample
```
drewoldag, line 216: Before we merge this, let's think through if it makes sense to put this check here. I'm wondering it the same snippet of code might be better on the `HyraxKafkaConsumer` class in `hyrax_alerts`. i.e.

```python
class HyraxKafkaConsumer(HyraxAlertsBaseConsumer, KafkaStreamDataset):
    def __init__(self, config, data_location=None):
        HyraxAlertsBaseConsumer.__init__(self, config=config, data_location=data_location)
        KafkaStreamDataset.__init__(self, config=config, data_location=data_location)

    def peek_sample(self):
        sample = super.peek_sample()
        # process `sample` with the new logic from this PR
        return sample

File: tests/hyrax/test_kafka_stream_dataset.py [hunk 2]
```diff
@@ -231,6 +233,45 @@ def test_peek_sample_buffers_and_replays(monkeypatch):
     assert [s["object_id"] for s in batches[0]] == ["first", "second"]
 
 
+def test_peek_sample_pre_filter(monkeypatch):
```
drewoldag, line 236: Same question as above, the tests looks reasonable, but I wonder if this belongs in `hyrax_alerts`.

src/hyrax/datasets/kafka_stream_dataset.py
drewoldag: I wonder only because I think `KafkaStreamDataset` can be completely unaware of pre filtering and processing and therefore (hopefully) more reusable.

But if I'm missing something obvious, please let me know :)

drewoldag (commented):
I have an open question about this PR that I would like to bottom out before approving.

src/hyrax/datasets/kafka_stream_dataset.py
maxwest-uw: my only question would be if that would not work nice with the wait/yield loop in the `KafkaStreamDataset` (i.e. what's the way that `HyraxKafkaConsumer` will know to continue the loop and wait for an alert if the alert returned out of `super.peek_sample()` get's pre-filtered and we need to wait for a sample again ?). It might require us to duplicate the loop in the overwritten version of the function which could get messy and lead to divergent functionality. let me give it a shot though !

PR #998, opened 2026-07-21
Fix publish benchmarks again

Graciaaa3:
## Change Description
Fix issue for `publish-benchmarks` workflow with:
- Add repo checkout step
- Fix artifact download path for `html` and `results`

This time should work 🤞


drewoldag (approved):
approved

PR #999, opened 2026-07-23
update calls to idist.auto_model to be _auto_model instead

SamSandwich07:
Forgot to update calls to `idist.auto_model` to use `_auto_model` instead in #996. Quick PR to fix that.

### DIFF
--- a/src/hyrax/pytorch_ignite.py
+++ b/src/hyrax/pytorch_ignite.py
@@ -41,6 +41,10 @@ _LEGACY_SPLIT_KEYS = ("train_size", "validate_size", "test_size")
 _old__getattr__ = Module.__getattr__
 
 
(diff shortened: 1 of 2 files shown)


drewoldag (approved):
Looks reasonable, but check on the unit test failures before merging.

mtauraso (approved):

drewoldag (approved):
I accept this for now. I think there might be a more eligant way of doing this, but I'm not well versed enough in patching to say how to do it.

SamSandwich07:
Yeah I can't say I am a fan of it this way either. I originally just tried importing the `_new_getattr__` function but that led to an import error because apparently it can't be imported. I guess if an easy solution to this is found should be a relatively quick PR to update in the future.

PR #1001, opened 2026-07-24
Fixup LSSTDataset to use DownloadedLSSTDataset's butler and cutout handling

mtauraso:


aritraghsh09 (approved):
Looks good to me. 

Nit: `LSSTDataset._fetch_single_cutout` passes `self.config["data_set"]["skymap"]` to `_request_patch_cached`, while `DownloadedLSSTDataset._fetch_cutout_with_cache` passes `self._butler_config["skymap"]`. These resolve to the same value, but it might be worth making them consistent.

drewoldag (approved):
This seems reasonable enough to me. Nice refactor.

PR #1003, opened 2026-07-28
applying suggestion from #951

SamSandwich07:
I clicked "apply changes" on Drew's suggestion in #951 but for some reason it didn't go through and because I auto-merged I have to create a whole new PR for one character.


src/hyrax/datasets/collate_utils.py
```diff
@@ -17,7 +17,7 @@ def collate_as_1d_light_curve(samples: list[dict], field: str) -> dict:
     dict
         Contains three keys: `<field>`, `<field>_length`, and `<field>_mask`
```
Graciaaa3, line 18: I think it would be better to update this part to use `<field>_lengths` as well, so that the naming is consistent.

drewoldag (approved):
Looks good. Not much to review :)

PR #1005, opened 2026-07-28
applying suggestion from #1003

SamSandwich07:
I missed Gracia's suggestion in #1003 before auto-merge kicked in


Graciaaa3 (approved):
Thanks for the fix :)

PR #1004, opened 2026-07-28
ASV plot for custom benchmark results

Graciaaa3:
## Change Description

- Update the asv-nightly workflow to also fetch benchmark results generated from the self-hosted runner. Results generated on the self-hosted runner are currently stored under `gondor_results`.
- Update the existing `track_test` function to read data from a specific benchmark results file. The function is kept for testing.
- Add a benchmark tracking class that loads the results file once and reuses it across all `track_*` functions, avoiding repeated file reads.
- Improve the naming convention for self-hosted benchmark result files to make them more informative. Results are now named using the format `<short_commit>_<YYYYMMDD_HHMMSS>.json`.

PR #1006, opened 2026-07-29
Fix broken `setup_cache` for `track_*` functions

Graciaaa3:
## Change Description
This PR is to fix a bug from #1004.
The way ASV does `setup_cache` is to return a data structure that is saved to disk. I had it saved as a local variable previously so is not working.


mtauraso (approved):
Looks good to me!

PR #1008, opened 2026-07-30
Bind validate_batch to DDP and update autoencoderv2 to use self.forward()

SamSandwich07:
Fixes distributed desyncing issue on unsupervised training notebooks (use autoencoverv2 model) by

1) Updating hyrax_autoencoderv2 to use `self.forward()` rather than `self._eval_encode` so that DDP can actually have an effect
2) Binding `validate_batch` to DDP in `pytorch_ignite.py` so that validation is done correctly through DDP


drewoldag (approved):
Looks good to me

Issue #1009, opened 2026-07-30
Update models to use self.forward() in *_batch if they don't already


SamSandwich07:
In #1008 we update `hyrax_autoencoderv2` to use `self.forward()` in its `*_batch` functions so that it's compatible with DDP. We should go through all of Hyrax's models and update them to do this if they don't already.

SamSandwich07:
Closed with #1010

PR #1010, opened 2026-07-30
Update models for DDP compatibility

SamSandwich07:
Closes issue #1009. DDP compatibility requires calling `self()` or `self.forward()` in the `*_batch` methods, so I have updated the existing models to do that.

### DIFF
--- a/src/hyrax/models/hsc_dcae.py
+++ b/src/hyrax/models/hsc_dcae.py
@@ -95,7 +95,7 @@ class HSCDCAE(nn.Module):
         x1 = self.activation(self.encoder1(data))
         x2 = self.activation(self.encoder2(self.pool(x1)))
         x3 = self.activation(self.encoder3(self.pool(x2)))
-        x4 = self.activation(self.encoder4(self.pool(x3)))
+        x4 = self.forward(data)
 
         # Decoder with skip connections
         x = self.activation(self.decoder4(x4) + x3)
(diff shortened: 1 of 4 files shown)


src/hyrax/models/image_dcae.py
```diff
@@ -207,6 +207,7 @@ def train_batch(self, batch):
 
         # Encode to latent space
         latent, skip_connections, encoded_shape = self.encode(data)
```
drewoldag, line 209: Unfortunate that self.encode has to be called twice in all of these.

src/hyrax/models/hsc_dcae.py
```diff
@@ -95,7 +95,7 @@ def train_batch(self, batch):
         x1 = self.activation(self.encoder1(data))
         x2 = self.activation(self.encoder2(self.pool(x1)))
         x3 = self.activation(self.encoder3(self.pool(x2)))
-        x4 = self.activation(self.encoder4(self.pool(x3)))
+        x4 = self.forward(data)
```
drewoldag, line 98: Also unfortunate that we can't remove the calls for x1, x2, x3. 

In a near future, we'll have support for returning a dictionary from the forward function, so we would be able to extract all this info without having to make 2 calls.

drewoldag (approved):
This seems fine. In two of these cases, the work that @maxwest-uw has done to allow returning a dictionary from the `forward` function will help out a lot.

PR #1011, opened 2026-07-31
Fix interpreter hang on training failure by daemonizing GpuMonitor

maxwest-uw:
## Change Description
GpuMonitor was a non-daemon thread started on every train() call, and its stop() was only called after trainer.run()/model.save() succeeded, with no try/finally. Any exception during training (bad config, OOM, a real model bug) skipped stop() entirely, leaving an immortal background thread that blocks the interpreter forever at shutdown while it waits to join it.

Mark GpuMonitor as daemon=True so it can never block process exit, and wrap the training body in train.py in try/finally so stop() is always called promptly regardless of the daemon flag.

## Code Quality

### DIFF
--- a/src/hyrax/gpu_monitor.py
+++ b/src/hyrax/gpu_monitor.py
@@ -13,7 +13,9 @@ class GpuMonitor(Thread):
     """
 
     def __init__(self, interval_seconds=1):
-        super().__init__()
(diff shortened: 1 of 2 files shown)


mtauraso (approved):
LGTM

PR #1017, opened 2026-08-03
increasing model coverage in unit tests

maxwest-uw:
## Change Description
Since `HyraxCNN` and `HyraxAutoencoderV2` are both meant as general use models included with `hyrax`, it would be nice if we had some unit tests for them.

## Code Quality


drewoldag (approved):
This seems nice. It's true the `prepare_inputs` on the models are designed to be modified by the user, but what is currently implemented in code (and tested here) should be stable.

PR #1018, opened 2026-08-03
update cache key to contain date

Graciaaa3:
## Change Description
Follow-up to PR #1007.

This PR updates the cache key naming convention to `nightly-results-<date>-<commit>`.

Including the date ensures that a new cache entry is created each day, even if no new commits have been made. This allows the nightly benchmark workflow to persist new results for every scheduled run while still distinguishing caches by commit.


drewoldag (approved):
looks ok

PR #1020, opened 2026-08-04
[Auto merge] Relax torch dependency by adding hash_tensor fallback

mtauraso:
Remove the torch >= 2.9.0 pin by providing a zero-copy SHA-256 fallback for hash_tensor (introduced in torch 2.9). The fallback uses numpy's buffer protocol with hashlib to avoid copying tensor data.


drewoldag (approved):
Looks ok to me.

PR #1022, opened 2026-08-06
Integrate `reduce_dimensions` with data_request

### LINKED ISSUE #1000: Standardize Data Input For `reduce_dimensions`
## design
To ensure that we allow the user to be able to flexibly define data inputs analysis functions like `reduce_dimensions`, we should have the ability for them to define the data input through the `hyrax` standard `data_request`. Hyrax should have a default `data_request` for `reduce_dimensions` that looks like

```
{
    "reduce_dimensions": {
        "default": {
            "dataset_class": "ResultDataset",
            "data_location": None, # will resolve to most recent results dir
            "fields": ["data"],
        }
    }
}

```

or in toml config format
```
[data_request.reduce_dimensions.default]
dataset_class = "ResultDataset"
data_location = false
fields = ["data"]
```

Both of these options should able to be overwritten by a user who wants to provide their own data_request definition.

## requirements

- [ ] dynamically load the `ResultDataset` from the `data_request` if provided
- [ ] otherwise, load the default `data_request` and load the last result directory (the current default behavior of `reduce_dimensions`.
- [ ] enforce that the `dataset_class` is `ResultDataset` and raise an error if the user tries to specify something else
- [ ] dynamically load the `data_location` if user provided
- [ ] get the user provided field in `fields` and use that as the input to `reduce_dimensions` (otherwise, grab the "data" column\*)
- [ ] enforce that `fields` only has one field listed and raise an error if the user tries to specify something else

\* this might cause a slight conflict with the current behavior of `ResultDataset` as the `__get_item__` will just return the result tensor, if you run into this I can work with you to merge the behavior from #940 for you) 

if you have any questions or troubles let me know :)

Graciaaa3:
## Change Description
Closes #1000 

Updates `reduce_dimensions` to use the standard data request structure for inputs.

The updated workflow initializes a `DataProvider` from the configured data request and retrieves the dataset through `DataProvider`. The implementation includes:

- Verify that the requested dataset is either a `ResultDataset` or an `InferenceDataset`.
- Ensure that at most one field is specified in the data request (defaulting to "data" when no field is provided).
- Add a helper function to retrieve the requested field column for reduction.

If no data request is given, the code fall back to the default behavior.

## Code Quality


src/hyrax/verbs/reduce_dimensions.py
```diff
@@ -117,15 +147,79 @@ def _run(
...
+            dataset = setup_dataset(self.config)
+            reduce_datasets = dataset.get("reduce_dimensions")
+            if not isinstance(reduce_datasets, DataProvider):
+                raise RuntimeError(
+                    "Configured [data_request.reduce_dimensions] must resolve to a DataProvider."
+                )
+
+            # prepped_datasets: friendly name -> dataset instance
+            # TODO: how to get the friendly name? Is `default` a good name for friendly name?
+            inference_results = reduce_datasets.prepped_datasets.get("default")
```
maxwest-uw: I think to make this more general and allow for using friendly names, we should check the `prepped_datasets` for length, and if the length is 1 use that one and otherwise throw an error if there are more than one.

We should confirm with Drew and Michael that there isn't a case where a user would want to concat two datasets together for `reduce_dimensions`.

src/hyrax/verbs/reduce_dimensions.py
```diff
@@ -117,15 +147,79 @@ def _run(
...
+
+            if not isinstance(inference_results, (ResultDataset, InferenceDataset)):
+                raise RuntimeError(
+                    "Configured [data_request.reduce_dimensions] must be a "
+                    "ResultDataset or InferenceDataset for reduction."
+                )
+
+            reduce_request = data_request["reduce_dimensions"]
+            reduce_default = reduce_request.get("default", {})
```
maxwest-uw, line 200: along with above comment, this shouldn't be called `reduce_default`

maxwest-uw (commented):
everything looks good so far! the only outstanding issue imo is figuring out grabbing the friendly name :)

src/hyrax/verbs/reduce_dimensions.py
Graciaaa3: That make sense to me, Drew said he’ll discuss this with me on Monday, so we can go over it together then.

drewoldag: I think Max has the right idea, the length of `prepped_datasets` should be 1 in this case. The intent of the use would be ambiguous if >1. But yes, we can discuss more in person on Monday.

Also, given that the length is 1, then you can just get the friendly name directly from `prepped_datasets` and use that in the rest of the function.

Graciaaa3:
A few notes on the current change:
- I left the logic for retrieving the data column as is to keep the code structure later in the workflow consistent with the previous implementation. Since we also want to support the case where there's no data request, so in that case I can't use DataProvider to index into.
- I didn't use `requested_fields` from DataProvider to determine the requested field, since when no field is explicitly provided, DataProvider defaults it to all available fields. I kept the logic with directly checking the data_request to determine whether the user explicitly specified a field.
- Updated the config to rename `reduce` to `reduce_dimensions`.

maxwest-uw (approved):
lgtm!

src/hyrax/config_migrations/migrations/004_move_umap_to_reduce.py
drewoldag, line 1: I think that the correct way to do this would be to create a new config migration that changes the current latest toml config from `reduce` to `reduce_dimension`.

src/hyrax/config_migrations/migrations/004_move_umap_to_reduce.py
```diff
@@ -1,7 +1,7 @@
 """Config migration: version 4 → version 5.
 
-Move the legacy ``[umap]`` and ``[umap.UMAP]`` to be under ``[reduce]`` table
-with ``[reduce.umap]`` and ``[reduce.umap.kwargs]``.
+Move the legacy ``[umap]`` and ``[umap.UMAP]`` to be under ``[reduce_dimensions]`` table
+with ``[rereduce_dimensionsduce.umap]`` and ``[reduce_dimensions.umap.kwargs]``.
```
drewoldag:
```suggestion
with ``[reduce_dimensions.umap]`` and ``[reduce_dimensions.umap.kwargs]``.
```

tests/hyrax/test_config_migrations/test_migration_004.py
drewoldag, line 1: Same comment as on `004_move_umap...`. I believe that the correct approach here would be to create a new test_migration_005 module to ensure that going from `reduce` to `reduce_dimensions` worked as expected.

drewoldag (changes requested):
The logic of gathering the data seems fine, but it's probably best to write a new migratrion script instead of reusing the existing.

tests/hyrax/test_config_migrations/test_migration_004.py
Graciaaa3: Oh yes that's right! Thankfully I haven't merge it in yet. I will do that now.

drewoldag (approved):
Looks good, thanks for updating!

PR #1024, opened 2026-08-11
[AUTO-MERGE] Cleanly shut down kafka connections.

### LINKED ISSUE #1025: Make sure that restarts are rock solid
Not sure if it's just me, but while developing this package, I've noticed that frequently I have to take down the entire system and restart everything if there is an error. 

Typical work mode is:
* fix a bug / make a code change
* kill and recreate the kafka docker container (not just restart)
* resend alerts to the kafka stream
* restart the hyrax_alerts test to consume from the queue

It seems that if something fails in hyrax_alerts, that I shouldn't need to restart everything. Again, not sure if it's just me, but if it isn't, then we should determine what the best practices are for allowing a consumer from a kafka topic to shut down gracefully and cleanly so that it can be restarted without having to touch the producer or kafka servers.

drewoldag:
## Solution Description
It seems that the source of the problems that I was seeing while developing hyrax_alerts using a local kafka stream was due to lingering connections to the stream. As one manual test would fail, it would leave a zombie connection. When I started up a new test using the same group id, the partition would be split between the zombie and the new connection. Sometimes the new connection would get a message to instantiate the model, sometimes not. Sometime new messages sent to the queue would go to the zombie, sometimes to the new connection. 

The changes here attempt to ensure that whenever and however an instance of the streaming dataset is shut down, either via a ctrl-c, a raised exception, etc., that the connection is closed correctly. 

Additionally it eliminates the possibility of an infinite poll for a peek sample and allows the user to define a peek timeout. 

And, the changes here ensure that a batch of messages that were inflight when an exception was raised will not be lost. i.e. the commit location of the queue is only updated when the next batch of data is requested. This means that some messages may be delivered twice after a restart, but should not be lost.


src/hyrax/hyrax_default_config.toml
```diff
@@ -394,6 +405,17 @@ batch_flush_timeout = 5.0
...
 
+# Arbitrary librdkafka settings merged into the consumer configuration, applied after
+# `credentials_file` so they win over it. Use this to tune client behavior, e.g.
+# [data_set.KafkaStreamDataset.consumer_options]
+# "max.poll.interval.ms" = 600000
+# "session.timeout.ms" = 45000
+#
+# `enable.auto.commit` is forced to false and should be left alone: offsets are committed
+# by the stream only once a batch has been fully processed downstream, so that a failure
+# mid-pipeline replays those alerts instead of silently dropping them.
```
drewoldag, line 420: Might be worth noting that the user _could_ but **SHOULD NOT** provide credentials here.

src/hyrax/datasets/kafka_stream_dataset.py
```diff
@@ -118,18 +159,83 @@ def __init__(self, config: dict, data_location=None):
...
 
         if credentials_file_path and credentials_file_path.exists():
             credentials = toml.load(credentials_file_path)
             self.consumer_config.update(credentials)
 
+        # Free-form librdkafka settings (max.poll.interval.ms, session.timeout.ms, ...).
+        # Applied last so they win over both the defaults above and the credentials file.
+        consumer_options = ds_config.get("consumer_options")
+        if consumer_options:
+            self.consumer_config.update(consumer_options)
```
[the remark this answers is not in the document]
drewoldag: @copilot Instead of overriding the values or dropping them, let's emit a logger.warning saying that the user defined settings can cause messages to be dropped, and advise users to use the default values.

src/hyrax/datasets/kafka_stream_dataset.py
```diff
@@ -227,9 +452,27 @@ def __iter__(self):
...
+                    else:
+                        self._log_message_error(msg)
 
                 if batch:
                     yield batch
+                    # Resumed, so the caller is done with that batch and it is finally
+                    # safe to advance the committed offset. Asynchronous: a synchronous
+                    # round trip per batch adds latency and eats into max.poll.interval.ms.
+                    # Consumer.close() flushes whatever is still in flight.
+                    self._commit(consumer)
```
maxwest-uw: I think it's fine for us to keep the normal auto commit off for our consumer, but we should also give the user the option to not commit at all if they don't want to

maxwest-uw: unless I'm wrong and it already allows this ?

drewoldag: This PR doesn't contain a nice way to do that. I can add a boolean config. Would a name like `commit_after_batch = True` be reasonable? True would be the default, False would mean "do not commit" and presumably imply that the each time we reconnect we would start getting messages from the beginning of the queue.

maxwest-uw (approved):

PR #1031, opened 2026-08-13
give a better error if collation functions fails

maxwest-uw:
## Change Description
currently, if there's an issue with the default collation function, `hyrax` will send an error message saying there was a problem with the "custom" collation function, even if the user doesn't provide one. I changed the error message to specify the collation function name so that it's more readable/less confusing for the user.

## Code Quality


drewoldag (approved):
Looks ok to me.

PR #1037, opened 2026-08-17
Config change for hsc notebook

Graciaaa3:
## Change Description

Small follow-up config change for the HSC unsupervised notebook. Updated the `set_config` to use `reduce_dimensions` instead of the old `reduce`.


maxwest-uw (approved):

drewoldag (approved):
Looks good

PR #1039, opened 2026-08-25
Implements a `context` singleton.

drewoldag:
This PR implements some machinery to support a `context` TypedDict that can be init'ed and accessed with get_context() by class in Hyrax. 

The PR also includes some clean up of the residual context work that was there before to support `to_onnx` and the vector databases. Those all now use the new `context` TypedDict that they get with get_context() instead of being wired in throughout the code.

### DIFF
--- a/HYRAX_GUIDE.md
+++ b/HYRAX_GUIDE.md
@@ -163,6 +163,8 @@ Hyrax discovers components through three registries:
 - Models must inherit from `torch.nn.Module` and implement `__init__`, `forward`,
   `train_batch`, and `prepare_inputs`.
 - The decorator wires up save/load, optimizer, and criterion handling.
+- Models reach run-scoped state (results dir, verb, distributed rank) via
+  `from hyrax import get_context`. See "Run context" below.
 - Built-in: `HyraxAutoencoder`, `HyraxAutoencoderV2`, `HyraxCNN`, `SimCLR`, `ImageDCAE`, `HSCAutoencoder`, `HSCDCAE`, `HyraxLoopback`
 - **External plugins supported** — use a fully qualified import path in the config
   (e.g. `model.name = "my_pkg.my_module.MyModel"`).
@@ -252,6 +254,31 @@ High-level pipeline:
 7. **Vector DB** — store and query latent vectors (ChromaDB or Qdrant).
 
 Each verb that produces output creates its own timestamped results directory.
+
+## Run context
+
+`src/hyrax/context.py` holds one run context for the process — a plain dict
+describing the run currently underway. Verbs populate it with `init_context()`
+immediately after `create_results_dir()`; everything else reads it with
+`get_context()` (also exported as `hyrax.get_context`).
+
+Keys: `results_dir` (a `Path`), `verb`, `rank`, `world_size`, plus `ml_framework`
+for ONNX export. Consumers are models (writing artifacts that don't fit the
+TensorBoard/MLflow scalar-metric paradigm), vector database implementations, and
+the ONNX exporter.
+
+- **`get_context()` always returns the same dict object.** `init_context()` clears
+  and repopulates it *in place* rather than rebinding it. That's what lets a model
+  take a handle in `__init__` and still see values a verb adds afterward. Do not
+  "simplify" this by assigning a new dict to the module global.
+- **The context is per-run and cleared on every verb invocation.** A model held
+  across `h.train()` then `h.infer()` sees the infer run's directory during the
+  second call.
+- **`idist.Parallel` spawns fresh processes, so module globals reset in child
+  ranks.** `Train._training` repopulates the context per-rank; any new distributed
+  entry point must do the same.
+- **`results_dir` is always a `Path`**, never a `str`. `model_exporters.py` uses
+  the `/` operator on it.
 
 ## Testing Conventions
 
--- a/HYRAX_GUIDE.md
+++ b/HYRAX_GUIDE.md
@@ -265,7 +265,9 @@ immediately after `create_results_dir()`; everything else reads it with
 Keys: `results_dir` (a `Path`), `verb`, `rank`, `world_size`, plus `ml_framework`
 for ONNX export. Consumers are models (writing artifacts that don't fit the
 TensorBoard/MLflow scalar-metric paradigm), vector database implementations, and
-the ONNX exporter.
+the ONNX exporter. **Nothing takes a context parameter** — consumers call
+`get_context()` themselves, and whoever starts the work is responsible for
+pointing the context at the right directory first.
 
 - **`get_context()` always returns the same dict object.** `init_context()` clears
   and repopulates it *in place* rather than rebinding it. That's what lets a model
@@ -274,6 +276,13 @@ the ONNX exporter.
 - **The context is per-run and cleared on every verb invocation.** A model held
   across `h.train()` then `h.infer()` sees the infer run's directory during the
   second call.
+- **Objects that outlive their run must snapshot what they need, not hold the
+  live context.** `VectorDB.__init__` copies `results_dir` into `self.results_dir`
+  for exactly this reason: `database_connection` hands the database back to the
+  user for interactive querying, and ChromaDB re-reads that path at query time to
+  spawn worker processes. A live handle would make a held connection follow
+  whatever verb ran last. Regression test:
+  `test_context.py::test_vector_db_snapshots_its_directory`.
 - **`idist.Parallel` spawns fresh processes, so module globals reset in child
   ranks.** `Train._training` repopulates the context per-rank; any new distributed
   entry point must do the same.
--- a/HYRAX_GUIDE.md
+++ b/HYRAX_GUIDE.md
@@ -163,7 +163,7 @@ Hyrax discovers components through three registries:
 - Models must inherit from `torch.nn.Module` and implement `__init__`, `forward`,
   `train_batch`, and `prepare_inputs`.
 - The decorator wires up save/load, optimizer, and criterion handling.
-- Models reach run-scoped state (results dir, verb, distributed rank) via
+- Models reach run-scoped state (results dir, verb) via
   `from hyrax import get_context`. See "Run context" below.
 - Built-in: `HyraxAutoencoder`, `HyraxAutoencoderV2`, `HyraxCNN`, `SimCLR`, `ImageDCAE`, `HSCAutoencoder`, `HSCDCAE`, `HyraxLoopback`
 - **External plugins supported** — use a fully qualified import path in the config
@@ -262,20 +262,12 @@ currently underway. Each verb run gets its own context object, and everything
 else reads the current one with `get_context()` (also exported as
 `hyrax.get_context`).
 
-Keys: `results_dir` (a `Path`), `verb`, `rank`, `world_size`, plus `ml_framework`
-for ONNX export. Consumers are models (writing artifacts that don't fit the
-TensorBoard/MLflow scalar-metric paradigm), vector database implementations, and
+Keys: `results_dir` (a `Path`), `verb`, and `ml_framework`.
+Consumers are models, vector database implementations, and
 the ONNX exporter. **Nothing takes a context parameter** — consumers call
 `get_context()` themselves, and whoever starts the work is responsible for
 pointing the context at the right directory first.
 
-- **The `Verb` base class owns the lifecycle.** `Verb.__init_subclass__` wraps
-  every verb's `run()` so a fresh context is installed on entry and released in a
-  `finally` — including when the verb raises. Verbs do not call `run_context()`
-  themselves; they record their directory with `update_context(results_dir=...)`
-  right after `create_results_dir()`. A verb that never creates a results
-  directory simply has no `results_dir` key, rather than inheriting the previous
-  run's.
 - **Each run gets a new context object.** This is what lets an object hold a
   context instead of copying values out of it. `VectorDB.__init__` stores
   `get_context()` and exposes `results_dir` as a property: `database_connection`
@@ -288,14 +280,6 @@ pointing the context at the right directory first.
 - **Models should call `get_context()` at the point of use**, not stash it in
   `__init__`. A saved context is pickled into distributed child processes before
   their ranks are assigned, so it reports rank 0 everywhere.
-- **Verb runs do not nest.** A context is released to empty, not to whatever was
-  current before, so a verb that calls another verb is left without one. Only the
-  deprecated `umap` verb does this, and it returns immediately after delegating.
-- **`idist.Parallel` spawns fresh processes, so module globals reset in child
-  ranks.** `Train._training` repopulates the context per-rank; any new distributed
-  entry point must do the same.
-- **`results_dir` is always a `Path`**, never a `str`. `model_exporters.py` uses
-  the `/` operator on it.
 
 ## Testing Conventions
 
--- a/HYRAX_GUIDE.md
+++ b/HYRAX_GUIDE.md
@@ -257,10 +257,10 @@ Each verb that produces output creates its own timestamped results directory.
 
 ## Run context
 
-`src/hyrax/context.py` holds one run context for the process — a plain dict
-describing the run currently underway. Verbs populate it with `init_context()`
-immediately after `create_results_dir()`; everything else reads it with
-`get_context()` (also exported as `hyrax.get_context`).
+`src/hyrax/context.py` holds the run context — a plain dict describing the run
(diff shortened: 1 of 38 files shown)


drewoldag (commented):
I've reviewed the work here and it aligns with what I was going for.

src/hyrax/context.py
```diff
@@ -0,0 +1,177 @@
+"""Run-scoped context shared by verbs, models, vector databases, and exporters.
+
+A Hyrax verb run produces a results directory and a handful of other facts about
```
mtauraso, line 3: This should probably be updated to ONLY represent the current state (not talking about prior states that this change actually removes).

src/hyrax/context.py
```diff
@@ -0,0 +1,177 @@
...
+        kwargs["results_dir"] = Path(kwargs["results_dir"])
+
+    kwargs.setdefault("rank", idist.get_rank())
+    kwargs.setdefault("world_size", idist.get_world_size())
+
+    _context.update(kwargs)
+    return _context
+
+
+def clear_context() -> None:
```
mtauraso, line 193: I'm wondering if we actually want to call this at the end of each verb run...

src/hyrax/vector_dbs/vector_db_interface.py
```diff
@@ -1,27 +1,39 @@
...
         """
         self.config = config if config else {}
-        self.context = context if context else {}
+
+        # Snapshot the directory rather than holding the live run context. A
+        # database object outlives the run that created it - database_connection
+        # hands one back to the user for interactive querying - and subclasses
+        # read this path at query time, long after __init__. Holding the live
+        # context would make a held connection follow whatever verb ran last.
+        self.results_dir: Path = get_context()["results_dir"]
```
mtauraso: This is really subtle and makes me wonder if we should be doing something cleaner at the end of each verb run, e.g. force `clear_context()` to be called at every verb ending in a `finally` clause and then change `clear_context` to simply make a new context object, and leak the old one to the GC.

If the user ends up with something connected to the context, so be it, let the GC handle it, and we'd know these sort of subtle state bugs between verbs simply aren't possible to write.

tests/hyrax/test_context.py
```diff
@@ -0,0 +1,328 @@
...
+    h.train()
+    assert get_context()["verb"] == "train"
+    assert get_context()["results_dir"] == find_most_recent_results_dir(h.config, "train")
+
+    h.infer()
+    assert get_context()["verb"] == "infer"
+    assert get_context()["results_dir"] == find_most_recent_results_dir(h.config, "infer")
+
+
+def test_vector_db_snapshots_its_directory(loopback_inferred_hyrax):
```
mtauraso: This test is making me more sure that we want slightly cleaner state management.

HYRAX_GUIDE.md
```diff
@@ -253,6 +255,40 @@ High-level pipeline:
...
+pointing the context at the right directory first.
+
+- **`get_context()` always returns the same dict object.** `init_context()` clears
+  and repopulates it *in place* rather than rebinding it. That's what lets a model
+  take a handle in `__init__` and still see values a verb adds afterward. Do not
+  "simplify" this by assigning a new dict to the module global.
+- **The context is per-run and cleared on every verb invocation.** A model held
+  across `h.train()` then `h.infer()` sees the infer run's directory during the
+  second call.
+- **Objects that outlive their run must snapshot what they need, not hold the
```
mtauraso: This is that same subtle distinction coming up, and needing to be documented system-wide.

mtauraso (approved):
This seems reasonable. 

I want to get a little cleaner on the state management of the context dictionary. I think we should enforce the structural invariant that it is cleared after each verb run to completely avoid issues of access between verbs, and having to make subtle distinctions at access time about whether we need to copy a value from the context, vs simply hold a pointer to it.

I think this could be a relatively simple follow-up PR, and I don't think it necessarily blocks this going in as is.

src/hyrax/vector_dbs/vector_db_interface.py
drewoldag: Perhaps I'm misunderstanding your point, but I believe that the reason we have to do this kind of strange things here in the vector database is because the state get's wiped at the end of the verb. 

I think that it's actually the odd UX of `database_connection` that makes this stand out. The context is established in `database_connection`, the database instance is returned, and then `database_connection` finishes and the context is lost. 

So here we grab a copy of the location of the database off of the context before it's lost.

drewoldag: I do agree that this is odd, but I _think_ `context` is roughly doing what you're advocating for. But please correct me if I'm not understanding what you're getting at.

src/hyrax/context.py
```diff
@@ -0,0 +1,215 @@
...
+    """The results directory for the current run."""
+
+    verb: str
+    """The name of the verb that started the run, e.g. ``"train"``, ``"infer"``"""
+
+    rank: int
+    """The distributed rank of the current process. ``0`` when not running distributed."""
+
+    world_size: int
+    """The number of distributed processes. ``1`` when not running distributed."""
```
drewoldag: Let's go ahead and remove these for now, we can reintroduce later if needed.

src/hyrax/verbs/verb_registry.py
```diff
@@ -31,6 +34,35 @@ def __init__(self, config):
...
+        super().__init_subclass__(**kwargs)
+
+        run = cls.__dict__.get("run")
+        if run is None or getattr(run, "_hyrax_bracketed", False):
+            return
+
+        @functools.wraps(run)
+        def run_with_context(self, *args, **kwargs):
+            with run_context(type(self).cli_name) as context:
+                self.context = context
```
drewoldag: I think we should be able to remote this `self.context`.

src/hyrax/context.py
```diff
@@ -0,0 +1,215 @@
...
+    """
+    # Imported here rather than at module scope to keep this module cheap for
+    # user model code to import.
+    import ignite.distributed as idist
+
+    if "results_dir" in kwargs:
+        kwargs["results_dir"] = Path(kwargs["results_dir"])
+
+    kwargs.setdefault("rank", idist.get_rank())
+    kwargs.setdefault("world_size", idist.get_world_size())
```
drewoldag: To be removed.

src/hyrax/context.py
```diff
@@ -0,0 +1,215 @@
...
+        The newly installed context.
+    """
+    global _current
+
+    _current = RunContext(verb=verb)
+    try:
+        update_context(**extra)
+        yield _current
+    finally:
+        _current = RunContext()
```
drewoldag: Replace with self.clear_context()

src/hyrax/context.py
```diff
@@ -0,0 +1,215 @@
...
+    RunContext
+        The re-installed context.
+    """
+    global _current
+
+    _current = context
+    try:
+        yield _current
+    finally:
+        _current = RunContext()
```
drewoldag: Replace with `self.clear_context()`

src/hyrax/vector_dbs/vector_db_interface.py
```diff
@@ -1,27 +1,45 @@
...
         """
         self.config = config if config else {}
-        self.context = context if context else {}
+
+        # Hold the context of the run that created us. Run contexts are per-run
+        # objects that are released when their verb finishes, so this keeps
+        # pointing at our own database directory however many verbs run later.
+        # That matters because a database object outlives the run that created it
+        # - database_connection hands one back to the user for interactive
+        # querying - and subclasses read results_dir at query time.
```
drewoldag: Update this comment to be human readable, instead of LLM readable.

src/hyrax/verbs/create_splits.py
```diff
@@ -43,12 +43,14 @@ def run(self):
...
         """
         from hyrax.config_utils import create_results_dir, log_runtime_config
+        from hyrax.context import update_context
         from hyrax.pytorch_ignite import setup_dataset
         from hyrax.splitting_utils import create_splits
 
         config = self.config
 
         results_dir = create_results_dir(config, "splits")
+        update_context(results_dir=results_dir, verb="splits")
```
drewoldag: Move this update_context into `create_results_dir`.

src/hyrax/verbs/train.py
```diff
@@ -107,6 +109,19 @@ def run(self):
...
+        # is absent in the child ranks. Repopulate it here, where every rank passes.
+        # On the single-process path this writes into the context run() installed, so
+        # anything a model stashed there during setup_model survives. In a spawned rank
+        # it fills that process's empty context, which needs no release because the
+        # process exits when training finishes.
+        update_context(
+            results_dir=results_dir,
+            verb="train",
+            rank=idist.get_rank(),
+            world_size=idist.get_world_size(),
```
drewoldag: Remove these

HYRAX_GUIDE.md
drewoldag, line 1: Review the length of this update, so that we can minimize the context window that claude needs

drewoldag: Possibly just delete it all, and update if Claude starts making mistakes

HYRAX_GUIDE.md
```diff
@@ -253,6 +255,48 @@ High-level pipeline:
...
+  `get_context()` and exposes `results_dir` as a property: `database_connection`
+  hands the database back to the user for interactive querying, and ChromaDB
+  re-reads that path at query time to spawn worker processes. Regression test:
+  `test_context.py::test_vector_db_keeps_the_context_of_its_own_run`.
+- **Work that outlives its run re-installs the context it captured.** Use
+  `use_context()`, as `InferStreamSession.process()` does, so model code driven
+  after the verb returned sees the run it belongs to.
+- **Models should call `get_context()` at the point of use**, not stash it in
+  `__init__`. A saved context is pickled into distributed child processes before
+  their ranks are assigned, so it reports rank 0 everywhere.
```
drewoldag, line 282: Just keep these three bullets

tests/hyrax/test_context.py
```diff
@@ -0,0 +1,382 @@
...
+    assert first["verb"] == "train"
+
+
+def test_run_context_populates_expected_keys():
+    """run_context plus the verb's own update_context fill the documented key set."""
+    with run_context("infer", results_dir="/some/results/dir") as context:
+        assert context["results_dir"] == Path("/some/results/dir")
+        assert context["verb"] == "infer"
+        assert context["rank"] == 0
+        assert context["world_size"] == 1
```
drewoldag: Remove these

tests/hyrax/test_context.py
```diff
@@ -0,0 +1,382 @@
...
+def test_clear_context():
+    """clear_context releases a context populated outside a verb run."""
+    update_context(results_dir="/some/run", verb="train")
+    clear_context()
+    assert get_context() == {}
+
+
+def test_get_context_is_exported_from_package():
+    """Users are told to write `from hyrax import get_context`."""
+    assert hyrax.get_context is get_context
```
drewoldag: Remove this test

tests/hyrax/test_context.py
```diff
@@ -0,0 +1,382 @@
...
+        "optimizer": {"name": "torch.optim.SGD"},
+        "torch.optim.SGD": {"lr": 0.01, "momentum": 0.9},
+        "scheduler": {"name": None},
+    }
+
+    model = ContextlessModel(config=config, data_sample=None)
+
+    assert model.context == {}
+    with pytest.raises(KeyError, match="outside of a verb run"):
+        model.context["results_dir"]
```
drewoldag: Remove this one

tests/hyrax/test_context.py
```diff
@@ -0,0 +1,382 @@
...
+
+    def _write_notes(self):
+        """Write accumulated state into the results dir of the run underway.
+
+        Read the context here rather than stashing a handle in __init__: this is
+        the form that reports the right rank under distributed training, where a
+        handle taken in the parent is pickled into each child as a stale copy.
+        """
+        context = get_context()
+        rank = context["rank"]
```
drewoldag: Remove this line

tests/hyrax/test_context.py
```diff
@@ -0,0 +1,382 @@
...
+@hyrax_model
+class ContextWritingLoopback(nn.Module):
+    """A loopback model that persists something of its own to the results dir.
+
+    Defined at module scope so it survives the pickling that distributed training
+    and the model registry perform.
+    """
+
+    def __init__(self, config, data_sample=None):
+        from functools import partial
```
drewoldag: Would be wise to move this along side the other loopback fixutres. Perhaps inherete from them.

Think through where the following tests live, perhap in the test_train and test_test modules.

tests/hyrax/test_context.py
```diff
@@ -0,0 +1,382 @@
...
+    The in-run half is covered by test_model_writes_to_results_dir_during_train,
+    where the model's notes land in the train results directory.
+    """
+    h, _ = loopback_hyrax
+
+    h.train()
+    assert get_context() == {}
+
+    h.infer()
+    assert get_context() == {}
```
drewoldag: Remove this

tests/hyrax/test_context.py
```diff
@@ -0,0 +1,382 @@
...
+    from hyrax.verbs.model import Model
+
+    h, _ = loopback_hyrax
+    h.train()
+
+    model_verb = Model(h.config)
+    model_verb.run()
+
+    assert model_verb.context["verb"] == "model"
+    assert "results_dir" not in model_verb.context
```
drewoldag: Remove this test

tests/hyrax/test_context.py
```diff
@@ -0,0 +1,382 @@
...
+    assert db.results_dir == vdb_path
+
+    # Run another verb, which gets its own context and then releases it.
+    h.infer()
+    assert get_context() == {}
+
+    # The held connection must be unaffected and still able to query.
+    assert db.results_dir == vdb_path
+    an_id = list(inference_results.ids())[0]
+    assert an_id in db.get_by_id(an_id)
```
drewoldag, line 185: move this over with the other vector db tests to make sure that we aren't duplicating the tests.

src/hyrax/vector_dbs/vector_db_interface.py
drewoldag: Removed comment.

src/hyrax/verbs/create_splits.py
drewoldag: Done.

tests/hyrax/test_context.py
drewoldag: removed

tests/hyrax/test_context.py
drewoldag: removed

tests/hyrax/test_context.py
drewoldag: Done

tests/hyrax/test_context.py
drewoldag: done

tests/hyrax/test_context.py
drewoldag: done

tests/hyrax/test_context.py
drewoldag: Done

src/hyrax/verbs/verb_registry.py
drewoldag: done

HYRAX_GUIDE.md
drewoldag: Trimmed this down a bunch.

src/hyrax/context.py
drewoldag: Removed

tests/hyrax/test_context.py
drewoldag: Turns out there isn't an existing test suite for general vector database stuff. Going to leave this here because I believe we'll be consolidating out VDB stuff _eventually_.

tests/hyrax/conftest.py
```diff
@@ -244,3 +258,102 @@ def collate_image(batch):
...
+    context-writing loopback model on top of that."""
+    import json
+
+    import torch.nn as nn
+
+    from hyrax.context import get_context
+    from hyrax.models.model_registry import hyrax_model
+
+    @hyrax_model
+    class ContextWritingLoopback(nn.Module):
```
mtauraso, line 275: Intentionally not inherited from LoopbackModel?

mtauraso (approved):
LGTM. Lets ship this. Only a minor question and its in a test.

